PR #1, opened 2020-02-05
Add mobx demo skeleton

AlexBedley:
Haven't done tests yet but this is my first stab. Did I do it right??

```shell
git clone git@github.com:BrightspaceHypermediaComponents/consistent-evaluation.git
cd consistent-evaluation
git checkout abedley/setup/mobx-demo
npm i
npm start
```


components/consistent-evaluation-page.js
```diff
@@ -0,0 +1,90 @@
...
+
+	constructor() {
+		super();
+
+		this.overallScore = NaN;
+		this.scores = [];
+	}
+
+	handleChange(i) {
+		return (e) => {
```
AlexBedley: kinda gross here, maybe it would be better to add `i` as a property on `d2l-input-text`, but not really the purpose of this demo

components/consistent-evaluation.js
```diff
@@ -1,65 +1,26 @@
...
+import { MobxLitElement } from '@adobe/lit-mobx';
 
+export class ConsistentEvaluation extends MobxLitElement {
 	constructor() {
 		super();
+		this.store = createStore();
+	}
 
-		this.prop1 = 'consistent-evaluation';
+	handleScoreChanged(e) {
```
AlexBedley: this looks better inline but our eslintconfig expressly forbids it `lit/no-template-arrow`

components/stores/score.js
```diff
@@ -0,0 +1,24 @@
...
+
+function average(scores) {
+	let sum = 0;
+	scores.forEach(s => sum += s);
+
+	return sum / scores.length;
+}
+
+export const addScore = (store) => {
+	store.score = {};
```
AlexBedley: maybe by convention we can say each file gets their own top level property?

components/stores/store.js
```diff
@@ -0,0 +1,12 @@
+import { addScore } from './score';
+import { observable } from 'mobx';
+
+export const createStore = () => {
+	let store = observable({});
+
+	store = addScore(store);
```
AlexBedley: I really like this pattern

components/consistent-evaluation-page.js
```diff
@@ -0,0 +1,90 @@
+import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
+import '@brightspace-ui/core/components/inputs/input-text.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+
+export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) {
```
AlexBedley: I think each element that needs langterms has to extend `LocalizeMixin`. Feels a bit weird though, since we have 1 massive file for the whole app.

components/consistent-evaluation-page.js
```diff
@@ -0,0 +1,90 @@
...
+			:host([hidden]) {
+				display: none;
+			}
+			d2l-input-text {
+				margin-bottom: 1.5rem;
+			}
+		`;
+	}
+
+	static async getLocalizeResources(langs) {
```
AlexBedley: I'd be a huge fan of hiding this code somewhere, since it will be the same for every file, but I havne't dug too deeply into this yet. It's next on my plate, along with integrating with BSI

components/consistent-evaluation-page.js
dmackenz: We could just localize in the container component, then inject all of the translations via properties. This removes the mixin dependency on every component.

components/consistent-evaluation-page.js
dmackenz: I think we should treat translations as there own module just like we are with the store.

components/consistent-evaluation-page.js
```diff
@@ -0,0 +1,90 @@
+import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
```
dmackenz: This file is acting as our container component, right?

components/consistent-evaluation-page.js
```diff
@@ -0,0 +1,90 @@
...
+				}));
+		};
+	}
+
+	render() {
+		return html`
+			<d2l-template-primary-secondary>
+				<div slot="header"><h1>Hello, consistent-evaluation!</h1></div>
+				<div slot="primary">
+					${this.scores.map((s, i) => html`<d2l-input-text
```
dmackenz: In the real app, when we use slots in the template we will probably want to inject custom elements. That way we can do the intermediate event catching similar to in the demo I prepared?

components/consistent-evaluation.js
dmackenz: It's actually better to not have them inline. I just did that in the demo out of laziness haha.

components/stores/score.js
dmackenz: I am conflicted about this. If you refer back to the demo, you'll notice that I appended the instance data of the store inside of `store.data` (`store.data.count`). I then put the store methods directly in the root of the object, but it would probably have been better to put the methods inside of their own property i.e. `store.counter.increment()` because otherwise, it will get much messier as the app grows.

The reason why I didn't put the instance data (`store.data.count`) inside the same place as the counter methods was for testing purposes. if we hold all of the store's instance data in one place, we can schema validate the entire store data at once (almost like taking a snapshot of the store). **This is very important when using MobX as opposed to Redux** because MobX is mutable (this was the tradeoff with picking this library). Refer to my implementation using the `Joi` library.

dmackenz: A way that we can mitigate the complexity of this is to use getters and setters. Consider:

The data for the counter is held inside of a single object `store.data.counter` but we can access that data through a getter/setter in `store.counter.getCount()`. This may be a good compromise to reducing the complexity of the store's api and also being able to schema validate the entire store at once in each test.

components/consistent-evaluation-page.js
AlexBedley: Waaaay back when we did React we used this method. It seemed to work ok although it was annoying to pass through the langterms for everything (it will add some boilerplate to testing/demoing/vdiffing). Although this might be a positive since langterm changes will no longer fail vdiff and unit tests.

AlexBedley: I'd be interested to snoop other repos and see what they're doing. The template that I used to spawn this repo seemed to be for a single, reusable component, not a complex application so maybe the langterm approach isn't correct

components/consistent-evaluation-page.js
AlexBedley: Yep this should probably be like `d2l-score-list` or something but I figured that was overkill for the purposes of this demo

components/stores/score.js
AlexBedley: As per our offline discussion, I think your approach makes sense, we just need to figure out how to handle stores relying on other observable properties will work out. I will move `overallScore` to a different store and make it observe this one and see how it goes

AlexBedley:
@dmackenz i updated the stores based on our discussions today.

I started with the big `data` object but found it annoying that to access my own classes' things I need to go 3-4 levels deep. So I ended up just nesting the stores themselves and passing `RootStore` to each of them. I'm betting that its more common to modify things in your own store rather than in other stores, so we should design for that experience. Of course this makes it harder to use schema validation in testing so I'm working with `joi` now to see how that'll go.

components/consistent-evaluation-page.js
dmackenz: I wonder if there is something equivalent to React's Context Provider for webcomponents. This would solve that problem of having props passed everywhere.

apalaniuk: Peanut gallery here: I think our approach to state tunneling/avoid prop drilling is essentially https://github.com/BrightspaceUI/core/blob/master/mixins/provider-mixin.md.

dmackenz:
@AlexBedley 

I have updated the three PR's that are out right now to clean them up. This is what I think we should do in terms of the order that we merge:

1) Merge `outcomes-refactor` into `master`
2) Pull `master` into `abedley/setup/mobx-demo` to ensure that it is up to date
3) Merge `mobx-tests` into `abedley/setup/mobx-demo`
4) Merge `abedley/setup/mobx-demo` into `master`

dmackenz:
I think this is good enough to merge as it satisfies "setting up MobX". Without MobX it is difficult to work on some of the other things so it would be nice to get this merged asap.

components/consistent-evaluation-page.js
jeroach: Is this comment still valid? or on hold for now (for a 2nd PR)

dmackenz: I would suggest this should be addressed in a follow-up PR when we have more of a game plan

components/stores/overallScore.js
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+}
+
+decorate(OverallScoreStore, {
+	overallScore: computed,
+	overallScoreTwo: computed
+});
+
+export { OverallScoreStore as default };
```
AlexBedley: @dmackenz is this any different than in-lining `export default` above?

package.json
```diff
@@ -19,6 +19,7 @@
     "@open-wc/testing": "^2",
     "@open-wc/testing-karma": "^3",
     "babel-eslint": "^10",
+    "chai": "^4.2.0",
```
AlexBedley: @dmackenz unused import?

package.json
```diff
@@ -29,12 +30,15 @@
     "frau-ci": "^1",
     "karma-sauce-launcher": "^2",
     "lit-analyzer": "^1",
+    "mocha": "^7.0.1",
```
AlexBedley: @dmackenz unused import?

components/stores/overallScore.js
dmackenz: I wasn't sure about this one. I wasn't sure if you do `export default class OverallScoreStore` if `decorate` would run? I also didn't do `export default decorate()` just because seeing `decorate = () => void` seemed strange. This was my in-between solution 🤷‍♂️

test/store/schema.js
```diff
@@ -0,0 +1,20 @@
+import OverallScoreStore from '../../components/stores/overallScore';
+
+const enumerable = (obj, getter) => {
+	Object.defineProperty(obj, getter, { enumerable: true });
+};
+
+['overallScore', 'overallScoreTwo']
```
AlexBedley: @dmackenz I know this works but I think it will be a nightmare to maintain, and cause a massive learning curve for anyone working in this repo. and I worry that the next mobx feature we want to use will also cause headaches.

We're doing this in the name of catching unintended side-effects, but it seems the getters are causing huge problems. I think we're already way ahead of where we were with QE in terms of testability, I wonder if just relying on each other to write good unit tests will be enough to catch those kind of bugs

test/store/utils.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+const isGetter = (obj, prop) => {
+	const getter = Object.getOwnPropertyDescriptor(obj, prop);
+	return (getter && getter.get);
+};
+
+export const validate = (schema, store) => {
+	for (const field in store) {
+		if (field !== 'root') {
```
AlexBedley: @dmackenz I don't think we should have a magic field like this. I could easily see this causing a huge problem - you gotta know to call the root store `root` AND if you ever had another field called `root` it would be ignored. I think iterating through the schema (instead of the store) is the better option here

package.json
AlexBedley: looks like it is being used by karma under the hood

PR #3, opened 2020-02-10
mobx store test demo

dmackenz:
This PR sets up the structure of the initial test stubs for mobx

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,7 @@
+{
+    "presets": [
+        [
+            "@babel/preset-env"
+        ]
+    ]
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,0 @@
-{
-    "presets": [
-        [
-            "@babel/preset-env"
-        ]
-    ]
-}
--- a/components/stores/overallScore.js
+++ b/components/stores/overallScore.js
@@ -7,9 +7,6 @@ class OverallScoreStore {
 		this.root = root;
 	}
 
-	set overallScore(overallScore) {}
-	set overallScoreTwo(overallScoreTwo) {}
-
 	get overallScore() {
 		const scores = this.root.scoresStore.scores;
 
@@ -18,11 +15,13 @@ class OverallScoreStore {
 
 		return sum / scores.length;
 	}
-
+	set overallScore(overallScore) {}
 	// Can have a computed property on another computed property
 	get overallScoreTwo() {
 		return this.overallScore * 2;
 	}
+	set overallScoreTwo(overallScoreTwo) {}
+
 }
 
 export default decorate(OverallScoreStore, {
--- a/components/stores/overallScore.js
+++ b/components/stores/overallScore.js
@@ -25,6 +25,4 @@ decorate(OverallScoreStore, {
 	overallScoreTwo: computed
 });
 
-['overallScore', 'overallScoreTwo'].forEach(getter => Object.defineProperty(OverallScoreStore, getter, { enumerable: true }));
-
 export { OverallScoreStore as default };
--- a/components/stores/overallScore.js
+++ b/components/stores/overallScore.js
@@ -7,6 +7,9 @@ class OverallScoreStore {
 		this.root = root;
 	}
 
+	set overallScore(overallScore) {}
+	set overallScoreTwo(overallScoreTwo) {}
+
 	get overallScore() {
 		const scores = this.root.scoresStore.scores;
 
--- a/components/stores/overallScore.js
+++ b/components/stores/overallScore.js
@@ -1,7 +1,5 @@
 import { computed, decorate } from 'mobx';
 
-// This probably should be inside ScoreStore but let's do it this way just to
-// demonstrate inter-store dependencies
 class OverallScoreStore {
 	constructor(root) {
 		this.root = root;
@@ -15,16 +13,18 @@ class OverallScoreStore {
 
 		return sum / scores.length;
 	}
-	set overallScore(overallScore) {}
-	// Can have a computed property on another computed property
+
 	get overallScoreTwo() {
 		return this.overallScore * 2;
 	}
-	set overallScoreTwo(overallScoreTwo) {}
 
 }
 
-export default decorate(OverallScoreStore, {
+decorate(OverallScoreStore, {
 	overallScore: computed,
 	overallScoreTwo: computed
 });
+
+['overallScore', 'overallScoreTwo'].forEach(getter => Object.defineProperty(OverallScoreStore, getter, { enumerable: true }));
+
+export { OverallScoreStore as default };
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@brightspace-ui-labs/consistent-evaluation",
   "description": "A consistent evaluation page for all tools",
+  "version": "0.0.1",
   "repository": "https://github.com/BrightspaceHypermediaComponents/consistent-evaluation.git",
   "private": true,
   "scripts": {
@@ -42,6 +43,7 @@
     "@hapi/joi": "^17.1.0",
     "@webcomponents/webcomponentsjs": "^2",
     "lit-element": "^2",
-    "mobx": "^5.15.4"
+    "mobx": "^5.15.4",
+    "d2l-outcomes-level-of-achievements": "github:Brightspace/outcomes-level-of-achievement-ui#semver:2",
   }
 }
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "@open-wc/testing": "^2",
     "@open-wc/testing-karma": "^3",
     "babel-eslint": "^10",
+    "chai": "^4.2.0",
     "deepmerge": "^3",
     "es-dev-server": "^1",
     "eslint": "^6",
--- a/package.json
+++ b/package.json
@@ -16,10 +16,6 @@
   "author": "D2L Corporation",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@babel/cli": "^7.8.4",
-    "@babel/core": "^7.8.4",
-    "@babel/preset-env": "^7.8.4",
-    "@babel/register": "^7.8.3",
     "@open-wc/testing": "^2",
     "@open-wc/testing-karma": "^3",
     "babel-eslint": "^10",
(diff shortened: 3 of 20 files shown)


AlexBedley:
There's a bunch of cruft in here, can you clean the diff a bit? Or are you not planning on merging this

dmackenz:
> There's a bunch of cruft in here, can you clean the diff a bit? Or are you not planning on merging this

Refer to https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/1#issuecomment-584702934

.babelrc
```diff
@@ -0,0 +1,7 @@
+{
```
dmackenz: This is required to make mocha run on the command line

components/stores/overallScore.js
```diff
@@ -15,11 +15,13 @@ class OverallScoreStore {
 
 		return sum / scores.length;
 	}
-
+	set overallScore(overallScore) {}
```
dmackenz: These were added to make `Joi` work.

package.json
```diff
@@ -11,11 +11,16 @@
     "start": "es-dev-server --app-index demo/index.html --node-resolve --dedupe --open --watch",
     "test": "npm run lint && npm run test:headless",
     "test:headless": "karma start",
-    "test:sauce": "karma start karma.sauce.conf.js"
+    "test:sauce": "karma start karma.sauce.conf.js",
+    "test:store": "mocha test/store/index.test.js --require @babel/register"
```
AlexBedley: I'm not a fan of this, can we stick to a single test runner (aka karma with the wc stuff?) It runs pretty fast but if you really want to only run the store tests it takes a directory

AlexBedley: There's a browser version of Joi you can use for karma

components/stores/overallScore.js
AlexBedley: Hmm I don't like not obvious test stuff creeping into our code. Does this have ramifications for `mobx`?

dmackenz: I could not tell any difference in how MobX behaved

jeroach: Maybe at a minimum, we could group all such setters together in the file, and mark them as such?

either with a comment, or ideally by changing the method name

test/store/joi-wrapper.js
```diff
@@ -0,0 +1,5 @@
+import '@hapi/joi/dist/joi-browser.min.js';
+
+/* eslint-disable no-undef */
+export const Joi = joi;
```
jeroach: Forgive my ignorance here, but I thought we were using mocha/chai as per the demo today

dmackenz: So the reason why we decided to use Joi rather than chai is the fact that we can assert the entire state of the store at once (think of this like invoking an action on the store, then taking a snapshot of it to validate all data at once).

We went with this approach instead of using chai because keeping track of all of the observables in MobX and trying to assert them in chai would be very difficult. Consider the following scenario:

> You write an action that changes variable A in the store
> Variables B, C & D are observing A and change when A changes
> Variable E is observing C and changes when C changes

With chai, you would need to follow this chain of dependencies (or most likely forget these relationships and they go untested), whereas using Joi, you are schema validating the entire store at once. This means that if you write a test and only code to expect A to change, the test will fail because B, C, D & E also changed (meaning that it forces you to test the entire observable chain). 

This was our way of mitigating the major _mutability_ con of using MobX.

dmackenz: Please refer to `test/store/scores.test.js` for an example.

AlexBedley: So I agree with evaluating the root store (vs just the props we expect to change) but in terms of using `Joi` specifically - could we not just use chai's `deepEqual`? Will we ever leverage the other `Jo` methods or will be always be checking for equality? Setting up the expected schema seems much more verbose than initializing a simple object.

test/store/scores.test.js
```diff
@@ -0,0 +1,25 @@
...
+	it('set a score', (done) => {
+		const rootStore = new RootStore();
+		const schema = createSchema();
+
+		// invoke action
+		rootStore.scoresStore.setScore(0, 5);
+
+		// validate array change
+		schema.scoresStore.scores = Joi.array().length(4);
+		const { error } = schema.scoresStore.scores.validate([5, 0, 0, 0]);
```
AlexBedley: Shouldn't we set `schema.scoresStore.scores` to `[5, 0, 0, 0]` and allow the `validate` call below handle this check?

test/store/root.test.js
```diff
@@ -0,0 +1,29 @@
...
+	const { error } = Joi.object(schema).validate(store);
+	if (error) throw error;
+};
+
+describe('Root tests', () => {
+	it('test initial schema', (done) => {
+		const rootStore = new RootStore();
+		const schema = createSchema();
+		validate(schema, rootStore);
+		done();
```
AlexBedley: This test (and the other one as well) is synchronous so I don't think we need `done`

components/stores/overallScore.js
AlexBedley: I only found [this](https://mobx.js.org/refguide/computed-decorator.html#setters-for-computed-values) from the docs, wasn't very verbose however

test/store/joi-wrapper.js
dmackenz: @AlexBedley I can test this theory out today

dmackenz:
@AlexBedley @jeroach @cbusmand2l I updated this to now use chai instead of Joi for the tests. A bit has changed since the last review so I will leave comments in the code.

components/stores/overallScore.js
```diff
@@ -16,13 +14,17 @@ class OverallScoreStore {
...
+
 }
 
-export default decorate(OverallScoreStore, {
+decorate(OverallScoreStore, {
 	overallScore: computed,
 	overallScoreTwo: computed
 });
+
+['overallScore', 'overallScoreTwo'].forEach(getter => Object.defineProperty(OverallScoreStore, getter, { enumerable: true }));
```
dmackenz: This is required so that the getter methods can be iterated across in the tests. Chai is unable to evaluate getter methods, which is the reason why the `validate` function is not a simple `assert.deepEquals` because this does not work on getters.

test/store/scores.test.js
```diff
@@ -0,0 +1,18 @@
...
+describe('Score tests', () => {
+	it('set a score', () => {
+		const rootStore = new RootStore();
+		const schema = createSchema();
+
+		rootStore.scoresStore.setScore(0, 5);
+
+		schema.scoresStore.scores = [5, 0, 0, 0];
+		schema.overallScoreStore.overallScore = 1.25;
+		schema.overallScoreStore.overallScoreTwo = 2.5;
```
dmackenz: Now the tests are just simply setting the variables that have changed in the schema.

test/store/utils.js
```diff
@@ -0,0 +1,27 @@
...
+				assert.deepEqual(store[field], schema[field]);
+			}
+		}
+	}
+	for (const getter in Object.getPrototypeOf(store)) {
+		if (isGetter(store, getter)) {
+			assert.deepEqual(store[getter], schema[getter]);
+		}
+	}
+};
```
dmackenz: This function is a manual version of `assert.deepEqual(schema, store)`. There are a few reasons for this:

1. chai can't handle the root circular dependency.
2. chai cannot evaluate the variables that are produced from `@computed` getters. To fix this I had to define those getters as enumerable on the prototype of the class, and then they could be accessed via `Object.getPrototypeOf()`. Unfortunately, this was the major limitation of using chai, but from what I can tell, this solves that issue.

components/stores/overallScore.js
jeroach: So this logic is for tests only? if so, can we put a comment in the file to indicate as such?

test/store/utils.js
```diff
@@ -0,0 +1,27 @@
...
+
+const isGetter = (obj, prop) => {
+	const getter = Object.getOwnPropertyDescriptor(obj, prop);
+	return (getter && getter.get);
+};
+
+export const validate = (schema, store) => {
+	for (const field in store) {
+		if (field !== 'root') {
```
dmackenz: this requires the field for the `RootStore` to be named 'root'. Not sure of a better way to go about this as this seems brittle.

test/store/utils.js
```diff
@@ -0,0 +1,27 @@
+import { assert } from '@open-wc/testing';
```
jeroach: Could these utils benefit tests other than `store`?

I'm thinking if this file makes more sense at the `test` folder level

test/store/utils.js
jeroach: hmmm, yeah I don't love this....

could we pass in the name of the field? Though that feels weird too

test/store/utils.js
dmackenz: I think for now they are probably only useful for the store tests. They are really only useful for the store as far as I can tell.

components/stores/overallScore.js
dmackenz: I have now moved this logic to the test folder so that it is not included in the bundle and is a lot more obvious what it is for.

test/store/schema.js
```diff
@@ -0,0 +1,20 @@
+import OverallScoreStore from '../../components/stores/overallScore';
+
+const enumerable = (obj, getter) => {
+	Object.defineProperty(obj, getter, { enumerable: true });
+};
+
+['overallScore', 'overallScoreTwo']
+	.forEach(getter => enumerable(OverallScoreStore, getter));
```
dmackenz: Now when you are changing the schema, you just need to add the method name in the array so that it gets picked up.

test/store/utils.js
dmackenz: I was thinking one option is that the classes that require access to the root (e.g. `OverallScoreStore`) could extend a class that has `root` already set? This way it is consistent across all of the classes and this assumption is a little less brittle.

e.g.

```
class OverallScoreStore extends AbstractRootAccessStore {
    constructor(root) {
        super(root)
    }
}
```

Also not the best solution...

test/store/schema.js
jeroach: cool, I like this better - thanks!

test/store/utils.js
jeroach: meh - might as well leave as is, since there isn't a clear preferred solution

PR #4, opened 2020-02-10
outcomes refactor

dmackenz:
This PR is just to make the outcomes component more ready for actually going inside of the production page. The `consistent-evaluation-seconday-block` was my idea of wrapping all of the components that will be placed inside of the secondary panel, so I thought I'd include it now.


components/consistent-evaluation-page.js
```diff
@@ -0,0 +1,93 @@
+import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
```
AlexBedley: does this file belong in this PR?

dmackenz: Good point -- not sure how this got in here. Will remove

dmackenz: This is now removed.

dmackenz:
Refer to https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/1#issuecomment-584702934

demo/index.html
```diff
@@ -1,47 +1,17 @@
...
-				</d2l-template-primary-secondary>
-			</d2l-demo-snippet>
-		</d2l-demo-page>
+		<h2>Main Component</h2>
+		<ul>
+			<li>
+				<a href="./consistent-evaluation.html">consistent-evaluation</a>
+			</li>
+		</ul>
+		<h2>Secondary Panel</h2>
```
AlexBedley: I think something like `consistent evaluation outcomes` would be more appropriate

demo/consistent-evaluation.html
```diff
@@ -0,0 +1,28 @@
...
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script type="module">
+			import '@brightspace-ui/core/components/demo/demo-page.js';
+			import '@brightspace-ui/core/components/typography/styles.js';
+			import '@brightspace-ui/core/components/typography/typography.js';
+			import '../components/consistent-evaluation.js';
+			import 'd2l-outcomes-level-of-achievements/d2l-outcomes-level-of-achievements.js';
```
AlexBedley: unnecessary import?

package-lock.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "@brightspace-ui-labs/consistent-evaluation",
```
AlexBedley: this needs to be updated? Actually I have no idea how to keep package-lock in sync, since we don't npm i unless we install a new dep. But I upgraded this repo to 0.0.4 and changed the name. Maybe frau-ci will commit the new package-lock? 🤷‍♂

PR #6, opened 2020-02-21
Adding right panel content.

rowansilver:
image
Adding the initial hookups to get the components visible.

- Still need to add a localized title to the feedback textbox.
- I don't think we need the store for the hrefs and token?
- does the textbox toolbar need to be always visible? (I believe this would require setting `inline="0"` on the `d2l-html-editor`, however doing so in its current state completely breaks the editor, and I'm not sure yet what attributes/css needs to change to fix it).
- the textbox toolbar is set up to have the insert stuff button, however I don't think it will appear without a valid href from the richtexteditorconfig, and initial digging seems to indicate that it's more than a trivial effort to get that in a demo page. I'm not sure if the effort will be worth it (as long as the toolbar ends up correctly displaying once it gets into BSI/quad).


demo/consistent-evaluation.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en-us">
```
cbusmand2l: The text editor needs a valid language from https://github.com/Brightspace/d2l-html-editor/tree/master/langs.

components/consistent-evaluation-page.js
```diff
@@ -53,6 +73,9 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		this.overallScore = NaN;
 		this.overallScoreTwo = NaN;
 		this.scores = [];
+		this.richTextEditorDisabled = false;
+		this._richTextEditorConfig = {};
+		this._htmlEditorUniqueId = `htmleditor-${getUniqueId()}`;
```
dmackenz: Should all of this initialization be in the store classes and injected in `ConsistentEvaluation`?

components/consistent-evaluation-page.js
```diff
@@ -85,9 +108,39 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
+
+	_onRequestProvider(e) {
+		if (e.detail.key === 'd2l-provider-html-editor-enabled') {
+			e.detail.provider = true;
+			e.stopPropagation();
+		}
+	}
+
+	_saveInstructionsOnChange() {
+		console.log('save on change');
```
dmackenz: These private methods should bubble events up to the store eventually?

cbusmand2l: We can do that, or move the handlers up to where we need them. Since there's nothing to hook these up to yet, this is more just a proof of concept at this point.

components/consistent-evaluation-page.js
cbusmand2l: `richTextEditorDisabled` and `_richTextEditorConfig` are just being assigned their logical defaults here, and `_htmlEditorUniqueId` is a private computed value.
I'm not yet sure how `_richTextEditorConfig` is going to be coming in from the HM calls so we may need to change it to be a public property in the future so that we can pass it in from its parent, but I don't see any reason to mark it either private or public for the moment until we have more information.

dmackenz:
Not sure why the `tinymce` import keeps failing. Looked into it and there were quite a few notes in `node_modules\d2l-html-editor\d2l-html-editor.js` about `tinymce`. 

Also I just added some additional stuff to this PR that can be seen in https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/7.

cbusmand2l:
> Not sure why the `tinymce` import keeps failing. Looked into it and there were quite a few notes in `node_modules\d2l-html-editor\d2l-html-editor.js` about `tinymce`.

Yeah, that's what I've been trying to figure out. It's at least failing locally so I don't have to keep checking things in every time I want to try something, but I've yet to discover the magical configuration to get it to recognize tinymce.

jeroach:
Figure I should actually write something here

I have been following this PR, and I don't see anything that really needs to be changed.

Mostly waiting for the test stuff to be sorted, but you're already on that :)

Once everything is all working, perhaps the screenshot could be updated (if need be). But I think that's all I have to say.

AlexBedley:
I created https://github.com/Brightspace/d2l-html-editor/pull/191 to address our problem and remove the test flakiness. This doesn't solve the other problem - that the axe tests aren't really testing the html component since it finishes before it even imports tinymce

AlexBedley:
Hey so I merged the vdiff PR this morning. This PR will break master if merged since it changes the UI. Happy to just delete the vdiff test I wrote if it's too much of a problem (we probably don't want to maintain a vdiff test of the entire app from day 1). Maybe worth talking about in retro

package.json
```diff
@@ -39,9 +39,12 @@
...
     "@adobe/lit-mobx": "0.0.2",
     "@brightspace-ui/core": "^1",
+    "@polymer/iron-demo-helpers": "^3.1.0",
     "@webcomponents/webcomponentsjs": "^2",
     "d2l-outcomes-level-of-achievements": "github:Brightspace/outcomes-level-of-achievement-ui#semver:^2",
     "lit-element": "^2",
-    "mobx": "^5.15.4"
+    "mobx": "^5.15.4",
+    "d2l-html-editor": "github:brightspace/d2l-html-editor",
+    "d2l-rubric": "github:brightspace/d2l-rubric"
```
AlexBedley: I think we need to specify a semver version for these 2 (probably just major version)

package.json
```diff
@@ -39,9 +39,12 @@
   "dependencies": {
     "@adobe/lit-mobx": "0.0.2",
     "@brightspace-ui/core": "^1",
+    "@polymer/iron-demo-helpers": "^3.1.0",
```
AlexBedley: is this used?

cbusmand2l: Interesting. I think this got pulled in from a merge at some point, but I don't see that it's used anywhere other than the package files. I'll take it out.

components/consistent-evaluation-page.js
AlexBedley: I'm surprised this passes the linter

PR #8, opened 2020-02-26
Putting right panel components into separate files

dmackenz:


components/consistent-evaluation-html-editor.js
```diff
@@ -137,27 +137,30 @@ class ConsistentEvaluationHtmlEditor extends LitElement {
...
-					class="d2l-html-editor-container"
-					id="${this._htmlEditorUniqueId}"
-					aria-label="${this.ariaLabel}"
-					aria-describedby="toolbar-shortcut-${this._htmlEditorUniqueId}"
-					?disabled="${this.disabled}"
-					role="textbox"
-					prevent-submit>
-				</div>
-			</d2l-html-editor>`;
+			<d2l-consistent-evaluation-secondary-block title="Overall Feedback">
```
cbusmand2l: The title is shown on the page, yes? Should it not be localized?

PR #12, opened 2020-03-06
Adding feedback actions bar component.

cbusmand2l:
image

### DIFF
--- a/components/consistent-evaluation-feedback-actions.js
+++ b/components/consistent-evaluation-feedback-actions.js
@@ -1,0 +1,133 @@
+import '@brightspace-ui/core/components/button/button-icon.js';
+import '@brightspace-ui/core/components/button/button-subtle.js';
+import '@brightspace-ui/core/components/colors/colors';
+import 'd2l-tooltip/d2l-tooltip';
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
+
+class FeedbackActions extends LocalizeMixin(LitElement) {
+
+	static get properties() {
+		return {
+			canAddFile: { type: Boolean },
+			canAddLink: { type: Boolean },
+			canAddGoogleDriveLink: { type: Boolean },
+			canAddOneDriveLink: { type: Boolean },
+			canRecordAudio: { type: Boolean },
+			canRecordVideo: { type: Boolean },
+			showAll: { type: Boolean }
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: flex;
+				flex-direction: row;
+				justify-content: space-between;
+				align-items: center;
+				background: var(--d2l-color-regolith);
+				border-radius: 6px;
+				border: 1px solid var(--d2l-color-mica);
+				padding: 12px;
+			}
+
+			d2l-button-icon:not([hidden]),
+			d2l-button-subtle:not([hidden]) {
+				display: inline-block;
+			}
+		`;
+	}
+
+	static async getLocalizeResources(langs) {
+		for await (const lang of langs) {
+			let translations;
+			switch (lang) {
+				case 'en':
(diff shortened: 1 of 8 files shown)


components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+
+	static get properties() {
+		return {
+			canAddFile: { type: Boolean },
+			canAddLink: { type: Boolean },
+			canAddGoogleDriveLink: { type: Boolean },
+			canAddOneDriveLink: { type: Boolean },
+			canRecordAudio: { type: Boolean },
+			canRecordVideo: { type: Boolean },
+			showAll: { type: Boolean }
```
cbusmand2l: `showAll` will override all the other `canAdd...` etc. options, and just show all defined buttons.

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+	constructor() {
+		super();
+
+		this._tooltipBoundary = {
+			left: 20 + 12, // padding-left applied to d2l-consistent-evaluation-feedback-actions + padding-left of d2l-button-icon
+			right: 0
+		};
+	}
+
+	_dispatchButtonClick(name) {
```
cbusmand2l: This component is the dumb UI-only part. I don't think we have enough information at this time to determine where we're going to handle the actual adding of links/recordings/etc, so for now this is just dispatching the button click events for its eventual parent to handle.

cbusmand2l:
The perceptual diff background colour makes it look like the background fill on this component isn't working (but it's definitely there):
image

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		return null;
+	}
+
+	constructor() {
+		super();
+
+		this._tooltipBoundary = {
+			left: 20 + 12, // padding-left applied to d2l-consistent-evaluation-feedback-actions + padding-left of d2l-button-icon
```
jeroach: I haven't seen this pattern before - did you copy from somewhere or come up with on your own?

Just curious :)

components/consistent-evaluation-feedback-actions.js
jeroach: Sounds reasonable

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+					id="${uniqueId}"
+					icon="d2l-tier1:${icon}"
+					text="${text}"
+					@click="${handler}">
+				</d2l-button-icon>
+				<d2l-tooltip
+					for="${uniqueId}"
+					aria-hidden="true"
+					disable-focus-lock
+					.boundary="${this._tooltipBoundary}">${text}</d2l-tooltip>
```
jeroach: super super nit: should the closing tag be on the following line for consistency?

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+			`;
+		}
+	}
+	_renderSubtleButton(icon, textKey, eventKey, visible) {
+		if (visible || this.showAll) {
+			const uniqueId = getUniqueId();
+			const text = this.localize(textKey);
+			const handler = () => this._dispatchButtonClick(eventKey);
+			return html`
+				<d2l-button-subtle
```
jeroach: Do subtle buttons have tooltips?

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+					@click=${handler}>
+				</d2l-button-subtle>
+			`;
+		}
+	}
+
+	render() {
+		return html`
+			<div class="button-container-left">
+				${this._renderIconButton('upload', 'addFile', 'add-file', this.canAddFile)}
```
jeroach: should these strings be constants?

cbusmand2l:
We'll probably import this from activities.

PR #14, opened 2020-03-09
d2l-grade-result Initial Presentational Layer

dmackenz:
This is all of the templating for the `d2l-grade-result` component.

image

### DIFF
--- a/components/d2l-grade-result/d2l-grade-result-icon-button.js
+++ b/components/d2l-grade-result/d2l-grade-result-icon-button.js
@@ -1,6 +1,7 @@
 import '@brightspace-ui/core/components/button/button-icon.js';
 import 'd2l-tooltip/d2l-tooltip.js';
 import { html, LitElement } from 'lit-element';
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
 
 export class D2LGradeResultIconButton extends LitElement {
 	static get properties() {
@@ -13,7 +14,7 @@ export class D2LGradeResultIconButton extends LitElement {
 
 	constructor() {
 		super();
-		this._id = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
+		this._id = getUniqueId();
 	}
 
 	_onClick() {
--- a/components/d2l-grade-result/d2l-grade-result-icon-button.js
+++ b/components/d2l-grade-result/d2l-grade-result-icon-button.js
@@ -1,0 +1,47 @@
+import '@brightspace-ui/core/components/button/button-icon.js';
+import 'd2l-tooltip/d2l-tooltip.js';
+import { html, LitElement } from 'lit-element';
+
+export class D2LGradeResultIconButton extends LitElement {
+	static get properties() {
+		return {
+			tooltipText: { type: String },
+			icon: { type: String },
+			_id: { type: String }
+		};
+	}
+
+	constructor() {
+		super();
+		this._id = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
+	}
+
+	_onClick() {
+		this.dispatchEvent(new CustomEvent('d2l-grade-result-icon-button-click', {
+			bubbles: true,
+			composed: true,
+		}));
+	}
+
+	render() {
+		return html`
+			<div>
+				<d2l-button-icon
+					id="d2l-grade-result-icon-button-${this._id}"
+					icon=${this.icon}
+					@click=${this._onClick}
+				></d2l-button-icon>
+				
+				${this.tooltipText && this.tooltipText !== '' ? html`
+					<d2l-tooltip
+						for="d2l-grade-result-icon-button-${this._id}"
+						position="bottom"
+					>
+						${this.tooltipText}
+					</d2l-tooltip>
+				` : html`` }
+			</div>
+		`;
+	}
+}
+customElements.define('d2l-grade-result-icon-button', D2LGradeResultIconButton);
--- a/components/d2l-grade-result/d2l-grade-result-icon-button.js
+++ b/components/d2l-grade-result/d2l-grade-result-icon-button.js
@@ -31,7 +31,7 @@ export class D2LGradeResultIconButton extends LitElement {
 					icon=${this.icon}
 					@click=${this._onClick}
 				></d2l-button-icon>
-				
+
 				${this.tooltipText && this.tooltipText !== '' ? html`
 					<d2l-tooltip
 						for="d2l-grade-result-icon-button-${this._id}"
--- a/components/d2l-grade-result/d2l-grade-result-icon-button.js
+++ b/components/d2l-grade-result/d2l-grade-result-icon-button.js
@@ -33,7 +33,7 @@ export class D2LGradeResultIconButton extends LitElement {
 					@click=${this._onClick}
 				></d2l-button-icon>
 
-				${this.tooltipText && this.tooltipText !== '' ? html`
+				${this.tooltipText ? html`
 					<d2l-tooltip
 						for="d2l-grade-result-icon-button-${this._id}"
 						position="bottom"
--- a/components/d2l-grade-result/d2l-grade-result-letter-score.js
+++ b/components/d2l-grade-result/d2l-grade-result-letter-score.js
@@ -1,0 +1,30 @@
+import { css, html, LitElement } from 'lit-element';
+import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles';
+
+export class D2LGradeResultLetterScore extends LitElement {
+
+	static get styles() {
+		return [selectStyles, css`
+			#d2l-grade-result-letter-score-container {
+				width: 8rem;
+				margin-right: 0.5rem;
+			}
+			#d2l-grade-result-letter-score-select {
+				width: 100%;
+			}
+		`];
+	}
+
+	render() {
+		return html`
+			<div id="d2l-grade-result-letter-score-container">
+				<select id="d2l-grade-result-letter-score-select" class="d2l-input-select">
+					<option>A</option>
+					<option>B</option>
+					<option>C</option>
+				</select>
+			</div>
+		`;
+	}
+}
+customElements.define('d2l-grade-result-letter-score', D2LGradeResultLetterScore);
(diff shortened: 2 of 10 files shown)


dmackenz:
Will be adding more to this, just wanted to get some initial opinions.

components/d2l-grade-result/d2l-grade-result-presentational.js
```diff
@@ -0,0 +1,121 @@
...
+
+				<div class="d2l-grade-result-presentational-score-text">
+					${this.scoreDenominator ? html`
+						<div class="d2l-body-standard">/ ${this.scoreDenominator}</div>
+					` : html``}
+				</div>
+				
+				<div>
+					<d2l-button-icon
+						id="d2l-grade-result-presentational-grade-button"
```
cbusmand2l: Do we still need to worry about unique IDs on the page?

dmackenz: This has been changed since your comment, may need further review.

dmackenz:
I have now updated the component to include a lot more of the design specification. Here is a rendering of the demo page: 

image

components/d2l-grade-result/d2l-grade-result-icon-button.js
```diff
@@ -0,0 +1,47 @@
...
+		return {
+			tooltipText: { type: String },
+			icon: { type: String },
+			_id: { type: String }
+		};
+	}
+
+	constructor() {
+		super();
+		this._id = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
```
cbusmand2l: Why not use `getUniqueId` from `@brightspace-ui/core/helpers/uniqueId`?

dmackenz: Didn't know that that existed, will do this

components/d2l-grade-result/d2l-grade-result-letter-score.js
```diff
@@ -0,0 +1,79 @@
...
+			detail: {
+				option: e.target.value
+			}
+		}));
+	}
+
+	render() {
+		if (this.doesUserHavePermissionToEvaluation === true) {
+			return html`
+				<div id="d2l-grade-result-letter-score-container">
```
cbusmand2l: If these IDs are only used for css styles, could we make them classes instead?

components/d2l-grade-result/d2l-grade-result-numeric-score.js
```diff
@@ -0,0 +1,76 @@
...
+			}
+		}));
+	}
+
+	render() {
+		return html`
+			<div id="d2l-grade-result-numeric-score-container">
+
+				${this.doesUserHavePermissionToEvaluation ? html`
+					<div id="d2l-grade-result-numeric-score-score">
```
cbusmand2l: If these IDs are only used for css styles, could we make them classes instead?

components/d2l-grade-result/d2l-grade-result-letter-score.js
dmackenz: Yeah that's fine with me, is there an advantage to using classes over ids?

dmackenz: changed now

components/d2l-grade-result/d2l-grade-result-numeric-score.js
dmackenz: changed now

components/d2l-grade-result/d2l-grade-result-letter-score.js
cbusmand2l: One of the shady/shadow doms has trouble with ids that are not unique per page because it doesn't compartmentalize them within the component. I'm not sure if this is still a problem or not, but it can't hurt.

components/d2l-grade-result/d2l-grade-result-icon-button.js
```diff
@@ -0,0 +1,48 @@
...
+	render() {
+		return html`
+			<div>
+				<d2l-button-icon
+					id="d2l-grade-result-icon-button-${this._id}"
+					icon=${this.icon}
+					@click=${this._onClick}
+				></d2l-button-icon>
+
+				${this.tooltipText && this.tooltipText !== '' ? html`
```
AlexBedley: For `this.tooltipText && this.tooltipText !== ''`  I think the second part is redundant, since `''` evaluates to `false`

components/d2l-grade-result/d2l-grade-result-letter-score.js
```diff
@@ -0,0 +1,78 @@
...
+			composed: true,
+			bubbles: true,
+			detail: {
+				option: e.target.value
+			}
+		}));
+	}
+
+	render() {
+		if (this.doesUserHavePermissionToEvaluation === true) {
```
AlexBedley: nit: don't need the `=== true` part if `this.doesUserHavePermissionToEvaluation` is `Boolean`

Also in terms of the name, is that the only case where this component is readOnly? Would it be safer to just name the property `readOnly`?

components/d2l-grade-result/d2l-grade-result-icon-button.js
dmackenz: made this change

components/d2l-grade-result/d2l-grade-result-presentational.js
```diff
@@ -0,0 +1,132 @@
...
+		`;
+	}
+
+	getScoreComponent() {
+		if (this.gradeType === GradeType.Number) {
+			return this.getNumericScoreComponent();
+		} else if (this.gradeType === GradeType.Letter) {
+			return this.getLetterScoreComponent();
+		} else {
+			throw 'INVALID GRADE TYPE PROVIDED';
```
AlexBedley: nit: we should throw `Error` types

components/d2l-grade-result/d2l-grade-result.js
```diff
@@ -0,0 +1,118 @@
+import './d2l-grade-result-presentational.js';
+import { html, LitElement } from 'lit-element';
+// eslint-disable-next-line no-duplicate-imports
```
AlexBedley: I'm surprised you have to import it twice and the specific import below doesn't cover the basic import above

dmackenz: I had to do this because of how the linter works. The linter says "GradeType" is not defined if I do the normal import. This was my way of fixing that.

components/d2l-grade-result/d2l-grade-result-letter-score.js
dmackenz: This has all been addressed in the code now

components/d2l-grade-result/d2l-grade-result-presentational.js
dmackenz: Change made

components/d2l-grade-result/d2l-grade-result.js
AlexBedley: I meant the opposite - just import GradeType and not the normal import. Either way not a huge deal, just surprising

dmackenz: Oh ok yeah I understand now. Made this change and it works.

components/d2l-grade-result/d2l-grade-result.js
```diff
@@ -0,0 +1,116 @@
...
+			_isManualOverrideActive: { type: Boolean }
+		};
+	}
+
+	constructor() {
+		super();
+
+		this._gradeType = GradeType.Number;
+		this._labeltext = 'Overall Grade';
+		this._scorenumerator = 5;
```
AlexBedley: This component is interesting because it's basically a controller - it intakes all the events and keeps track of stuff like the numerator. I would expect it to emit events and not keep track of anything

components/d2l-grade-result/d2l-grade-result.js
```diff
@@ -0,0 +1,116 @@
...
+					reportsbuttontooltip=${this._reportsbuttontooltip}
+					?readOnly=${!this._doesUserHavePermissionToEvaluate}
+					@d2l-grade-result-grade-change=${this._handleGradeChange}
+					@d2l-grade-result-grade-button-click=${this._handleGradeButtonClick}
+					@d2l-grade-result-reports-button-click=${this._handleReportsButtonClick}
+					@d2l-grade-result-manual-override-button-manual-override-click=${this._handleManualOverrideClick}
+					@d2l-grade-result-manual-override-button-manual-override-clear-click=${this._handleManualOverrideClearClick}
+				></d2l-grade-result-presentational>
+			`;
+		} else if (this._gradeType === GradeType.Letter) {
```
AlexBedley: I find this conditional interesting - I feel like `d2l-grade-result-presentational` already contains all the "number or letter grade" logic. I wonder if we could just pass everything down in this component (or get rid of this component completely and just use `d2l-grade-result-presentational`)

components/d2l-grade-result/d2l-grade-result.js
AlexBedley: Is your intention to replace this component with the HM controller? or is the hm controller going to wrap around this?

demo/d2l-grade-result.html
```diff
@@ -0,0 +1,376 @@
...
+				background-color: #dee7ed;
+				border-radius: 25px;
+			}
+		</style>
+	</head>
+
+	<body class="d2l-typography">
+		<d2l-demo-page page-title="d2l-grade-result">
+
+			<div class="section">
```
AlexBedley: Love the coverage in this file

AlexBedley: This also feels like these cases could be the vdiff test cases

components/d2l-grade-result/d2l-grade-result.js
dmackenz: Yeah this component is where all of the HM logic will eventually be

components/d2l-grade-result/d2l-grade-result.js
dmackenz: Oh I only did this to show that the component can be rendered differently in the demo pages. This is not meant to be permanent code and I imagine everything in this file will change quite a bit once we start integrating with HM.

components/d2l-grade-result/d2l-grade-result.js
AlexBedley: ahh ok I think I was confused by the name (I would have expected the visual component to just be called `d2l-grade-result`)

components/d2l-grade-result/d2l-grade-result.js
AlexBedley: ah yes makes sense

PR #15, opened 2020-03-13
Adding more responsive demos for outcomes and rubrics

dmackenz:
These demos were made to demonstrate that the components are responding to parent element size rather than window size, which is what we were worried about.


dmackenz:
This PR is currently waiting on the visual difference bot to be applied to the repo.

PR #18, opened 2020-04-15
Adding right panel component to consistent evaluation

dmackenz:
This PR is related to US114710.

Acceptance Criteria:

- Only desktop usage is in scope for this story (we'll create a follow up story to handle the changes needed for mobile experience)
- Note: the components may or may not show up depending on how the Assignment is configured (e.g. there might not be a rubric added, or a grade item set up, etc). Decide during this story whether it's a useful intermediate step to just include all components all the time (to enable testing of what we have so far), or if we should just add the logic to show/hide right from the start.
- At the end of this story, all 4 components will exist in the right panel and will auto-save
- Decide whether it makes sense to do US114711 (Add right panel component to the app) at the same time as completing this story, or if it's easier to handle as a separate chunk of work


dmackenz:
Need to regenerate goldens once the bot is added.

components/consistent-evaluation-page.js
```diff
@@ -33,120 +26,44 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
-			this.dispatchEvent(
-				new CustomEvent('d2l-score-changed', {
-					detail: {
-						score: parseInt(e.target.value),
-						i: i
-					}
-				}));
-		};
-	}
+		console.log(
```
jeroach, line 34: likely an artifact of development - doubt we need to log this

components/consistent-evaluation.js
```diff
@@ -7,34 +8,46 @@ export class ConsistentEvaluation extends MobxLitElement {
...
 	}
 
-	handleScoreChanged(e) {
-		this.store.scoresStore.setScore(e.detail.i, e.detail.score);
+	async firstUpdated() {
+		super.firstUpdated();
+
+		const controller = new ConsistentEvaluationHrefController(this.href, this.token);
+		this._childHrefs = await controller.getHrefs();
+		console.log('this._childHrefs', this._childHrefs);
```
jeroach, line 35: more `console.log` - maybe just do a search :)

components/consistent-evaluation.js
```diff
@@ -7,34 +8,46 @@ export class ConsistentEvaluation extends MobxLitElement {
...
-			.overallScore=${this.store.overallScoreStore.overallScore}
-			.overallScoreTwo=${this.store.overallScoreStore.overallScoreTwo}
-			.rubricHref=${this.rubricHref}
-			.token=${this.token}
-			.rubricAssessmentHref=${this.rubricAssessmentHref}
-			.outcomesHref=${this.outcomesHref}
-			@d2l-score-changed=${this.handleScoreChanged}>
-		</d2l-consistent-evaluation-page>`;
+			<d2l-consistent-evaluation-page
+				.rubricHref=${this._childHrefs && this._childHrefs.rubricHref}
```
jeroach, line 42: Forgive my ignorance here - can you please explain the syntax?
e.g. what does `.` vs `?` mean before a variable name?

Also, do we want this much logic in the definition? Would it make more sense to have these as computed variable? I honestly don't know :) Just thinking out loud

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -0,0 +1,73 @@
...
+
+		let gradeHref = undefined;
+		let feedbackHref = undefined;
+		let evaluationHref = undefined;
+		let nextHref = undefined;
+		let previousHref = undefined;
+		let rubricAssessmentHref = undefined;
+		let rubricHref = undefined;
+
+		const gradesRel = 'https://activities.api.brightspace.com/rels/grade';
```
jeroach, line 28: Should these be defined at the top level `Rels` constants class?

components/consistent-evaluation.js
dmackenz: so the `.` and `?` are LitElement specific stuff (it has to do with binding variables). Honestly still trying to figure out what the `.` does but the `?` is for binding boolean viarables. From Lit website:

> Bind properties to templated elements
> You can insert JavaScript expressions as placeholders for HTML text content, attributes, Boolean attributes, properties, and event handlers.

> Text content: `<p>${...}</p>`
> Attribute: `<p id="${...}"></p>`
> Boolean attribute: ?disabled="${...}"
> Property: .value="${...}"
> Event handler: @event="${...}"

components/controllers/ConsistentEvaluationHrefController.js
dmackenz: I agree these should be in some form of a global namespace

components/consistent-evaluation.js
dmackenz: This is a syntax to evaluate whether or not the component appears on the page.

```
this._childRefs = {
    outcomesHref: "..."
}

this._childHrefs && this._childHrefs.rubricHref => undefined
this._childHrefs && this._childHrefs.outcomesHref => "..."
```

If an `href === undefined` then it will not be displayed in the right panel.

components/consistent-evaluation.js
dmackenz: removed

components/consistent-evaluation-page.js
dmackenz: removed

dmackenz:
right-panel

components/controllers/ConsistentEvaluationHrefController.js
jeroach: Are you thinking about this for later on? Or in this PR?

dmackenz: I should probably address that in this PR then do a similar change in the footer PR.

PR #19, opened 2020-04-24
dmackenzie/add footer content

dmackenz:
This PR is regarding the initial setup of the footer (publish, save, retract, next student). It is dependent on this PR https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/18 being merged first. I did not include the full integration with the API at the top-level component for this PR because that logic is in the other PR. Will add that once the other PR is merged.


dmackenz:
This PR also needs the visual diff bot reinstalled to be merged.

dmackenz:
footer

components/consistent-evaluation-page.js
```diff
@@ -134,6 +135,10 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 					></d2l-consistent-evaluation-outcomes>
 
 				</div>
+				<div slot="footer">
+					<!-- the hrefs need to be added here once main controller is merged -->
```
dmackenz, line 139: This is what is waiting for the other PR.

components/controllers/FooterController.js
```diff
@@ -0,0 +1,45 @@
+import 'd2l-polymer-siren-behaviors/store/entity-store.js';
+import { performSirenAction } from 'siren-sdk/src/es6/SirenAction.js';
+
+// these will eventually be exported to a global configuration namespace
```
dmackenz, line 4: This will also change once this is established in the other PR.

components/footer/consistent-evaluation-footer.js
```diff
@@ -0,0 +1,80 @@
...
+		this.token = undefined;
+
+		this._controller = undefined;
+		this._evaluationEntity = undefined;
+	}
+
+	async updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (changedProperties.has('evaluationHref')) {
```
dmackenz, line 31: I have this for now. This will likely need to change once we have more of the components on the page working.

components/footer/consistent-evaluation-footer.js
```diff
@@ -0,0 +1,80 @@
...
+			this._controller = new ConsistentEvaluationFooterController(this.evaluationHref, this.token);
+			this._evaluationEntity = await this._controller.getEvaluationEntity();
+		}
+	}
+
+	_onPublishClick() {
```
dmackenz, line 37: All of these will also be filled in once we have more of a full integration.

components/controllers/FooterController.js
```diff
@@ -0,0 +1,45 @@
...
+export class ConsistentEvaluationFooterController {
+	constructor(evaluationHref, token) {
+		this.evaluationHref = evaluationHref;
+		this.token = token;
+	}
+
+	async getEvaluationEntity(bypassCache = false) {
+		const evaluationResource = await window.D2L.Siren.EntityStore.fetch(this.evaluationHref, this.token, bypassCache);
+
+		if (!evaluationResource.entity) {
```
jeroach, line 17: should we check `evaluationResource` first? before checking the `entity`

components/controllers/FooterController.js
```diff
@@ -0,0 +1,45 @@
...
+			throw new Error('Error while fetching evaluation entity.');
+		}
+
+		return evaluationResource.entity;
+	}
+
+	isPublished(evaluationEntity) {
+		return evaluationEntity.hasActionByName(retractActionName);
```
jeroach, line 25: I take it the underlying logic is: "if retract action exists, then publish must have occurred"

components/footer/consistent-evaluation-footer-presentational.js
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	_emitPublishEvent()     { this._dispatchButtonClickEvent('on-publish'); }
+	_emitRetractEvent()     { this._dispatchButtonClickEvent('on-retract'); }
+	_emitUpdateEvent()      { this._dispatchButtonClickEvent('on-update'); }
+	_emitSaveDraftEvent()   { this._dispatchButtonClickEvent('on-save-draft'); }
+	_emitNextStudentEvent() { this._dispatchButtonClickEvent('on-next-student'); }
+
+	_getPublishOrRetractButton() {
+		const text = this.published ? 'Retract' : 'Publish';
```
jeroach, line 46: Lang terms coming later on I assume?

components/footer/consistent-evaluation-footer-presentational.js
```diff
@@ -0,0 +1,88 @@
...
+		const eventEmitter = this.published ? this._emitRetractEvent : this._emitPublishEvent;
+		const id = `consistent-evaluation-footer-${text.toLowerCase()}`;
+		return html`<d2l-button primary id=${id} @click=${eventEmitter} >${text}</d2l-button>`;
+	}
+
+	_getSaveDraftOrUpdateButton() {
+		const text = this.published ? 'Update' : 'Save Draft';
```
jeroach, line 53: Same comment re: lang terms

components/controllers/FooterController.js
dmackenz: Yeah, that was what I went with originally, there may be a better way though. I think time will tell on this one once we have more of the controller set up.

components/footer/consistent-evaluation-footer-presentational.js
dmackenz: Yeah that will have to be a follow-up to this

components/footer/consistent-evaluation-footer-presentational.js
dmackenz: same comment as above

components/controllers/FooterController.js
dmackenz: That's a good point. I assumed I wouldn't need to because I assumed EntityStore would throw an error, but I shouldn't trust that. Will add that.

components/controllers/FooterController.js
ChrisKraljevicD2L: Seems like the simplest way to track if an entity is published would be a boolean property that gets set when published, but I have very little context with this domain so I'm not sure if that's gonna fly. Just seems more explicit than checking for an action which could be changed/removed later?

components/controllers/FooterController.js
```diff
@@ -0,0 +1,42 @@
...
+
+	async retract(evaluationEntity) {
+		if (evaluationEntity.hasActionByName(retractActionName)) {
+			const retractAction = evaluationEntity.getActionByName(retractActionName);
+			await performSirenAction(this.token, retractAction, null, true);
+		} else {
+			throw new Error('Could not find the retract action from the evaluation entity.');
+		}
+	}
+}
```
ChrisKraljevicD2L, line 42: nit: you could make one `performAction` function here that both `publish` and `retract` call with their respective `actionName`s.

dmackenz: thanks for the suggestion, this change has now been made!

dmackenz:
> As an overall comment, these recent PRs that were posted in #web-component-review have been very large. I don't have enough context with these repos to say for certain if they could be broken down into smaller individual ones but going forward that would be greatly appreciated.

Yeah, I agree they are a bit large. Mostly this is due to the fact that a lot of it is the groundwork for getting the repos up and running, a lot of the future PRs will be much smaller I imagine.

components/controllers/FooterController.js
dmackenz: The `isPublished` change has also been implemented now too. Thanks!

components/controllers/FooterController.js
```diff
@@ -0,0 +1,46 @@
...
+import { publishActionName, retractActionName } from './constants.js';
+import { performSirenAction } from 'siren-sdk/src/es6/SirenAction.js';
+
+export class ConsistentEvaluationFooterController {
+	constructor(evaluationHref, token) {
+		this.evaluationHref = evaluationHref;
+		this.token = token;
+	}
+
+	async getEvaluationEntity(bypassCache = false) {
```
jeroach, line 11: Should this be `createEvaluationEntity`?

Seems like it would only be called on initialization (since we set the `published` status to false).

The `get` prefix seems a bit off, tis all

dmackenz: maybe a better name would've been `requestEvaluationEntity`

jeroach: Either of those works for me

PR #22, opened 2020-05-03
wiring up top-level component and created tests for href and footer controllers

dmackenz:
Creating this PR to make that footer interact with the right panel. This PR implements:
- the next student button now reloads the entire context and iterates forward
- the publish and retract buttons now work and will appear depending on the current context
- there are now unit tests for the href and footer controllers

### DIFF
--- a/components/consistent-evaluation.js
+++ b/components/consistent-evaluation.js
@@ -27,11 +27,17 @@ export class ConsistentEvaluation extends MobxLitElement {
 		this._childHrefs = undefined;
 	}
 
-	async firstUpdated() {
-		super.firstUpdated();
+	async updated(changedProperties) {
+		super.updated();
 
-		const controller = new ConsistentEvaluationHrefController(this.href, this.token);
-		this._childHrefs = await controller.getHrefs();
+		if (changedProperties.has('href')) {
+			const controller = new ConsistentEvaluationHrefController(this.href, this.token);
+			this._childHrefs = await controller.getHrefs();
+		}
+	}
+
+	onNextStudentClick() {
+		this.href = this._childHrefs.nextHref;
 	}
 
 	render() {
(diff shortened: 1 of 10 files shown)


components/consistent-evaluation.js
```diff
@@ -27,11 +27,17 @@ export class ConsistentEvaluation extends MobxLitElement {
 		this._childHrefs = undefined;
 	}
 
-	async firstUpdated() {
-		super.firstUpdated();
+	async updated(changedProperties) {
+		super.updated();
```
dmackenz, line 31: Now realizing this is pretty critical moving forward, this is how we can trigger the components to reload automatically when we change the hrefs

components/controllers/FooterController.js
```diff
@@ -2,45 +2,83 @@ import 'd2l-polymer-siren-behaviors/store/entity-store.js';
...
-		return evaluationEntity.published;
+		if (!evaluationEntity) {
+			return false;
+		}
+
+		if (!evaluationEntity.properties) {
+			return false;
+		}
+
+		return evaluationEntity.properties.state === 'Published';
```
dmackenz, line 62: `isPublished` now works by checking the state directly instead of keeping a boolean variable to track it.

components/controllers/FooterController.js
```diff
@@ -2,45 +2,83 @@ import 'd2l-polymer-siren-behaviors/store/entity-store.js';
...
+	}
+
+	async __performSirenAction(action) {
+		return await performSirenAction(this.token, action, null, true);
 	}
```
dmackenz, line 67: This was made this way so that it could be more easily stubbed in the tests.

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -43,10 +67,12 @@ export class ConsistentEvaluationHrefController {
...
 			rubricAssessmentHref = getHref(root, assessmentRel);
 
-			let assessmentEntity = await window.D2L.Siren.EntityStore.fetch(rubricAssessmentHref, this.token, bypassCache);
-			if (assessmentEntity.entity) {
-				assessmentEntity = assessmentEntity.entity;
-				rubricHref = getHref(assessmentEntity, rubricRel);
+			if (rubricAssessmentHref) {
+				let assessmentEntity = await this._getAssessmentEntity(rubricAssessmentHref, bypassCache);
+				if (assessmentEntity && assessmentEntity.entity) {
+					assessmentEntity = assessmentEntity.entity;
```
jeroach, line 73: Why do we reassign this immediately after getting it and ensuring it's valid?

components/controllers/FooterController.js
jeroach: The published state could be a  `const`

components/controllers/ConsistentEvaluationHrefController.js
dmackenz: This is just a way to avoid using another variable, if you look further up in this function you will see that I did the same thing for the `root` variable. It might be easier to understand if I separated it out into two variables, maybe `assessmentResponse` and `assessmentEntity`?

jeroach: Why not just `rubricHref = getHref(assessmentEntity.entity, rubricRel);`

jeroach: I get the usage for `root` since the new var is used in a lot of places,  but there is only one usage here.

dmackenz: that's a good point, will make that change today

PR #25, opened 2020-05-06
fixing the responsive demos

dmackenz:
There were some issues rendering some of the demos in smaller divs to show their responsiveness. This PR fixes that...


dmackenz:
before:
before
after:
after

PR #26, opened 2020-05-07
Cbusman/add feedback actions component

vix633:


components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
+import '@brightspace-ui/core/components/button/button-icon.js';
```
jeroach, line 1: nit: should this file go in the `right-panel` subfolder?

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+			canRecordAudio: { type: Boolean },
+			canRecordVideo: { type: Boolean },
+			showAll: { type: Boolean }
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: flex;
```
jeroach, line 26: mega nit: alpha align these

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		return null;
+	}
+
+	constructor() {
+		super();
+
+		this._tooltipBoundary = {
+			left: 20 + 12, // padding-left applied to d2l-consistent-evaluation-feedback-actions + padding-left of d2l-button-icon
```
jeroach, line 67: nit: should this just be 32 (with the math explained in the comment?

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		if (!token) {
+			throw new Error(FeedbackControllerErrors.INVALID_TOKEN);
+		}
+
+		if (typeof token !== 'string') {
+			throw new Error(FeedbackControllerErrors.INVALID_TYPE_TOKEN);
+		}
+
```
jeroach, line 34: Shouldn't this controller set `this.token` and `this.baseHref` here (like the HrefController does)?

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -0,0 +1,75 @@
...
+	}
+	async requestFeedback() {
+		const response = await window.D2L.Siren.EntityStore.fetch(this.baseHref, this.token);
+		if (!response) {
+			throw new Error(FeedbackControllerErrors.REQUEST_FAILED);
+		}
+		if (!response.entity) {
+			throw new Error(FeedbackControllerErrors.ENTITY_NOT_FOUND_REQUEST_FEEDBACK);
+		}
+		const entity = response.entity;
```
jeroach, line 44: Why not just return `response.entity`

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -0,0 +1,75 @@
...
+			throw new Error(FeedbackControllerErrors.NO_SAVE_FEEDBACK_ACTION);
+		}
+
+		const saveFeedbackAction = entity.getActionByName(actionName);
+
+		if (!saveFeedbackAction.hasFieldByName(fieldName)) {
+			throw new Error(FeedbackControllerErrors.FIELD_IN_ACTION_NOT_FOUND(saveFeedbackAction.name, fieldName));
+		}
+
+		const field = saveFeedbackAction.getFieldByName(fieldName);
```
jeroach, line 69: typo

vix633: Typo?

components/consistent-evaluation-feedback-actions.js
vix633: No idea, rowans code

components/consistent-evaluation-feedback-actions.js
vix633: Whats a mega nit? if a nit is a "do or dont" is a mega nit a "dont"?

components/consistent-evaluation-feedback-actions.js
vix633: Not sure, should it be? As you know the visual part of this are Rowan's changes.

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
+import '@brightspace-ui/core/components/button/button-icon.js';
+import '@brightspace-ui/core/components/button/button-subtle.js';
+import '@brightspace-ui/core/components/colors/colors';
+import 'd2l-tooltip/d2l-tooltip';
```
dmackenz, line 4: I believe this component is in `core` now, so it should be imported from there.

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+			}
+
+			d2l-button-icon:not([hidden]),
+			d2l-button-subtle:not([hidden]) {
+				display: inline-block;
+			}
+		`;
+	}
+
+	static async getLocalizeResources(langs) {
```
dmackenz, line 43: We should externalize this function so that it is not rewritten everywhere. Similar to this: https://github.com/BrightspaceUILabs/grade-result/blob/master/src/components/locale.js

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+					@click="${handler}">
+				</d2l-button-icon>
+				<d2l-tooltip
+					for="${uniqueId}"
+					aria-hidden="true"
+					disable-focus-lock
+					.boundary="${this._tooltipBoundary}">${text}</d2l-tooltip>
+			`;
+		}
+	}
```
dmackenz, line 99: Shouldn't this function return an empty `HTMLTemplate`? I didn't know you could do this without crashing Lit-element. i.e.

```
if (!(visible || this.showAll)) {
 return html``
}
// rest of function...
```

components/consistent-evaluation-feedback-actions.js
```diff
@@ -0,0 +1,133 @@
...
+	_renderSubtleButton(icon, textKey, eventKey, visible) {
+		if (visible || this.showAll) {
```
dmackenz, line 101: Same comment here

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if (!response.entity) {
+			throw new Error(FeedbackControllerErrors.ENTITY_NOT_FOUND_REQUEST_FEEDBACK);
+		}
+		const entity = response.entity;
+
+		return entity;
+	}
+	async updateFeedbackText(feedbackText, entity) {
+		if (!feedbackText) {
```
dmackenz, line 50: What happens if they just don't want to give any feedback anymore? e.g. clear the feedback

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -0,0 +1,76 @@
...
+			throw new Error(FeedbackControllerErrors.INVALID_FEEDBACK_TEXT);
+		}
+		if (!entity) {
+			throw new Error(FeedbackControllerErrors.FEEDBACK_MUST_HAVE_ENTITY);
+		}
+
+		const actionName = 'SaveFeedback';
+		const fieldName = 'feedback';
```
dmackenz, line 58: These should go in the global constants as they are subject to change https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/controllers/constants.js

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -0,0 +1,76 @@
...
+		const saveFeedbackAction = entity.getActionByName(actionName);
+
+		if (!saveFeedbackAction.hasFieldByName(fieldName)) {
+			throw new Error(FeedbackControllerErrors.FIELD_IN_ACTION_NOT_FOUND(saveFeedbackAction.name, fieldName));
+		}
+
+		const field = saveFeedbackAction.getFieldByName(fieldName);
+		field.value = feedbackText;
+
+		const newFeedbackEntity = await performSirenAction(this.token, saveFeedbackAction, [field], true);
```
dmackenz, line 73: nit: you can just say `return await performSirenAction(this.token, saveFeedbackAction, [field], true);`

components/consistent-evaluation-feedback-actions.js
jeroach: lol still dealer's choice - I'd fix it since I'm anal, but you're not me

components/consistent-evaluation-feedback-actions.js
jeroach: Yeah I think it should

karma.conf.js
```diff
@@ -18,6 +18,7 @@ module.exports = config => {
 				// if you are using 'bare module imports' you will need this option
 				nodeResolve: true,
 			},
+			hostname: '127.0.0.1'
```
dmackenz, line 21: What is this for? just wondering

package.json
```diff
@@ -43,10 +43,11 @@
...
     "@webcomponents/webcomponentsjs": "^2",
-    "d2l-html-editor": "github:Brightspace/d2l-html-editor#semver:^2",
+    "d2l-html-editor": "github:brightspace/d2l-html-editor",
     "d2l-outcomes-level-of-achievements": "github:Brightspace/outcomes-level-of-achievement-ui#semver:^2",
-    "d2l-rubric": "github:Brightspace/d2l-rubric#semver:^3",
+    "d2l-rubric": "github:brightspace/d2l-rubric",
     "lit-element": "^2",
-    "mobx": "^5.15.4"
+    "mobx": "^5.15.4",
+    "d2l-tooltip": "github:BrightspaceUI/tooltip"
```
dmackenz, line 51: I think this should be removed

components/controllers/ConsistentEvaluationFeedbackController.js
jeroach: lol sorry - I read this as `FileId` but with a typo - Ignore me

karma.conf.js
vix633: Safari doesnt resolve localhost to 127.0.0.1 properly for me for some reason

components/controllers/ConsistentEvaluationFeedbackController.js
dmackenz, line 50: It might be worth doing a type check on the feedback text too:
`typeof feedbackText === 'string'`

components/consistent-evaluation-feedback-actions.js
vix633: Not sure,maybe @svanherk knows? This is also Rowan's code.

components/controllers/ConsistentEvaluationFeedbackController.js
vix633: Good point, I checked and there's no delete methods on the backend for this route, so setting the feedback to empty string is actually valid

components/consistent-evaluation-feedback-actions.js
svanherk: Looks like this is just a helper function, not the actual `render` function.  The actual `render` will always return the `div`s, which may or may not be empty based on what this function returns

components/consistent-evaluation-feedback-actions.js
vix633: I meant to ask, should it be?

components/controllers/ConsistentEvaluationFeedbackController.js
vix633: These guys were easier to debug, seeing what its set to first. Can be simplified now.

components/consistent-evaluation-feedback-actions.js
vix633: Do we have an avenue for doing that? Some lang actions component or something of that sort?

components/consistent-evaluation-feedback-actions.js
jeroach: I think so - it's odd to assign a property to an equation (instead of a value) IMO

package.json
```diff
@@ -43,9 +43,9 @@
     "@brightspace-ui-labs/grade-result": "^1.0.5",
     "@brightspace-ui/core": "^1.36.0",
     "@webcomponents/webcomponentsjs": "^2",
-    "d2l-html-editor": "github:Brightspace/d2l-html-editor#semver:^2",
+    "d2l-html-editor": "github:brightspace/d2l-html-editor",
```
dmackenz, line 46: Is this correct? I thought we wanted to use semver? Never knew what the correct thing to do here though, it's not very clear at all.

vix633: Uhh, i wonder why this even changed. @svanherk, any insight on what this should be?

svanherk: This looks bad... I think you want to set it back to what it was before, this will definitely break things

PR #38, opened 2020-06-03
POC for submission list page

zhangium:
image

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -1,5 +1,6 @@
 import './footer/consistent-evaluation-footer.js';
 import './right-panel/consistent-evaluation-right-panel.js';
+import './submissions-page.js';
 import '@brightspace-ui/core/components/inputs/input-text.js';
 import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
@@ -18,6 +19,8 @@ export default class ConsistentEvaluationPage extends LitElement {
 			token: { type: String },
 			rubricReadOnly: { type: Boolean },
 			richTextEditorDisabled: { type: Boolean },
+			submissionList: { type: Array },
+			evaluationState: { type: String }
 		};
 	}
 
@@ -38,7 +41,10 @@ export default class ConsistentEvaluationPage extends LitElement {
 				<div slot="header"><h1>Hello, consistent-evaluation!</h1></div>
 				<div slot="primary">
 					<div>
-						<span>evidence</span>
+						<submissions-page
+						.submissionList=${this.submissionList}
+						evaluationState=${this.evaluationState}
+						.token=${this.token}></submissions-page>
 					</div>
 				</div>
 				<div slot="secondary">
--- a/components/controllers/ConsistentEvaluationHrefController.js
+++ b/components/controllers/ConsistentEvaluationHrefController.js
@@ -49,6 +49,8 @@ export class ConsistentEvaluationHrefController {
 		let previousHref = undefined;
 		let rubricAssessmentHref = undefined;
 		let rubricHref = undefined;
+		let submissionList = undefined;
+		let evaluationState = undefined;
 
 		if (root && root.entity) {
 			root = root.entity;
@@ -66,7 +68,8 @@ export class ConsistentEvaluationHrefController {
 			nextHref = getHref(root, nextRel);
 			previousHref = getHref(root, previousRel);
 			rubricAssessmentHref = getHref(root, assessmentRel);
-
+			submissionList = root.getSubEntityByClass('assignment-submission-list').links;
+			evaluationState = root.getSubEntityByClass('evaluation').properties.state;
 			if (rubricAssessmentHref) {
 				const assessmentEntity = await this._getAssessmentEntity(rubricAssessmentHref, bypassCache);
 				if (assessmentEntity && assessmentEntity.entity) {
@@ -85,7 +88,9 @@ export class ConsistentEvaluationHrefController {
 			nextHref,
 			previousHref,
 			rubricAssessmentHref,
-			rubricHref
+			rubricHref,
+			submissionList,
+			evaluationState
 		};
 	}
 }
--- a/components/submissions-page.js
+++ b/components/submissions-page.js
@@ -1,0 +1,158 @@
+import 'd2l-polymer-siren-behaviors/store/entity-store.js';
+import '@brightspace-ui/core/components/list/list.js';
+import '@brightspace-ui/core/components/list/list-item.js';
+import '@brightspace-ui/core/components/list/list-item-content.js';
+import '@brightspace-ui/core/components/icons/icon.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { heading4Styles, labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
+
+export class SubmissionsPage extends LitElement {
+	static get properties() {
+		return {
+			submissionList: { type: Array },
+			token: { type: String },
+			evaluationState: { type: String }
+		};
+	}
+
+	static get styles() {
+		return [heading4Styles, labelStyles, css`
+		.submission-info{
+			margin: 0;
+			padding: 0;
+			font-size: 0.7rem;
+			font-weight: 400;
+			letter-spacing: 0.2px;
+		}
+		.last-item {
+			padding-bottom: 1rem;
+		}
+		.first-item {
+			padding-top: 1rem;
+			margin: 0;
+		}
+		.item {
+			display: inline-block;
+			vertical-align: middle;
+			padding-left: 0.5rem;
+		}
+	`];
+	}
+
+	constructor() {
+		super();
+		this._submissionList = [];
+		this._token = undefined;
+		this._submissionEntities = [];
+	}
+
+	get submissionList() {
+		return this._submissionList;
+	}
+
+	set submissionList(val) {
+		const oldVal = this.submissionList;
+		if (oldVal !== val) {
+			this._submissionList = val;
+			this._initialize().then(() => this.requestUpdate());
+		}
+	}
+	get token() {
+		return this._token;
+	}
+
+	set token(val) {
+		const oldVal = this.token;
+		if (oldVal !== val) {
+			this._token = val;
+			this.requestUpdate();
+		}
+	}
+
+	async _initialize() {
+		if (this._submissionList !== undefined) {
+			for (const submissionLink of this._submissionList) {
+				if (submissionLink.href) {
+					const submission = await this._getSubmissionEntity(submissionLink.href);
+					this._submissionEntities.push(submission);
+				}
+			}
+		}
+	}
+
+	async _getSubmissionEntity(submissionHref) {
+		return await window.D2L.Siren.EntityStore.fetch(submissionHref, this._token, false);
+	}
+
+	_getIcon(filename) {
+		const ext = filename.split('.')[1];
+		if (ext === 'png' || ext === 'jpg') {
+			return 'file-image';
+		}
+		else {
+			return 'file-document';
+		}
+	}
+
+	_getReadableFileSizeString(fileSizeBytes) {
+		let i = -1;
+		const byteUnits = [' kB', ' MB', ' GB', ' TB', 'PB', 'EB', 'ZB', 'YB'];
+		do {
+			fileSizeBytes = fileSizeBytes / 1024;
+			i++;
+		} while (fileSizeBytes > 1024);
+		return Math.max(fileSizeBytes, 0.1).toFixed(1) + byteUnits[i];
+	}
+
+	_renderList() {
+		let builder = '';
+		for (let j = 0; j < this._submissionEntities.length; j++) {
+			if (this._submissionEntities[j].entity) {
+				const assignmentSubmission = this._submissionEntities[j].entity;
+				console.log(assignmentSubmission);
+				const displayNumber = this._submissionEntities.length - j;
+				const dateStr = assignmentSubmission.getSubEntityByClass('date').properties.date;
+				const date = new Date(dateStr);
+				const attachmentsListEntity = assignmentSubmission.getSubEntityByClass('attachment-list');
+				builder += `
+				<h3 class="first-item">Submission ${displayNumber}</h3>\n
+				<p class="submission-info">${date.toDateString()} ${date.toLocaleTimeString()}<d2l-icon icon="tier1:dot"></d2l-icon>${this.evaluationState}</p>`;
+				if (attachmentsListEntity && attachmentsListEntity.entities) {
+					const attachmentsList = attachmentsListEntity.entities;
+					for (let i = 0; i < attachmentsList.length; i++) {
+						const filename = attachmentsList[i].properties.name;
+						const fileSizeBytes = attachmentsList[i].properties.size;
+						builder += `<d2l-list-item>
+						<d2l-list-item-content>
+						<d2l-icon class="item" icon="tier2:${this._getIcon(filename)}"></d2l-icon>
+						<div class="item">
+						<a href="${attachmentsList[i].properties.href}"><span>${filename}</span></a>
+						<div slot="secondary">${this._getReadableFileSizeString(fileSizeBytes)}</div></div>
+						</d2l-list-item-content>
+						</d2l-list-item>\n`;
+					}
+				}
+				let comment = '';
+				if (assignmentSubmission.getSubEntityByClass('submission-comment')) {
+					comment = assignmentSubmission.getSubEntityByClass('submission-comment').properties.html;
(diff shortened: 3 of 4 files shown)


components/consistent-evaluation-page.js
```diff
@@ -39,6 +42,10 @@ export default class ConsistentEvaluationPage extends LitElement {
 				<div slot="primary">
 					<div>
 						<span>evidence</span>
```
jeroach, line 44: So in order to fix the visual diff test (and allow us to remove this `span`) we just need to regenerate the goldens after removing this.

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -1,91 +1,106 @@
...
+			feedbackHref,
+			evaluationHref,
+			nextHref,
+			previousHref,
+			rubricAssessmentHref,
+			rubricHref
+		};
+	}
+
+	async getSubmissionInfo() {
```
jeroach, line 92: Much cleaner 👍

components/submissions-page.js
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	render() {
+		return html`
+				${this._renderList()}
+		`;
+	}
+}
+
+customElements.define('submissions-page', SubmissionsPage);
```
jeroach, line 158: Some thoughts re: moving this from a POC to hardening a bit:
- let's prefix the tag name and class name accordingly
`consistent-evaluation-submissions-page` and `ConsistentEvaluationSumbissionsPage`

Could also move the file itself into a `left-panel` folder (since that's where it will end up eventually.

components/submissions-page.js
```diff
@@ -0,0 +1,158 @@
...
+						<a href="${attachmentsList[i].properties.href}"><span>${filename}</span></a>
+						<div slot="secondary">${this._getReadableFileSizeString(fileSizeBytes)}</div></div>
+						</d2l-list-item-content>
+						</d2l-list-item>\n`;
+					}
+				}
+				let comment = '';
+				if (assignmentSubmission.getSubEntityByClass('submission-comment')) {
+					comment = assignmentSubmission.getSubEntityByClass('submission-comment').properties.html;
+					builder += `<d2l-list-item><d2l-list-item-content>
```
jeroach, line 138: @AlexBedley Annie and I already talked about this, but for context:

We'll eventually sub componentize this bit - just as is for the POC

components/submissions-page.js
```diff
@@ -0,0 +1,158 @@
...
+		} while (fileSizeBytes > 1024);
+		return Math.max(fileSizeBytes, 0.1).toFixed(1) + byteUnits[i];
+	}
+
+	_renderList() {
+		let builder = '';
+		for (let j = 0; j < this._submissionEntities.length; j++) {
+			if (this._submissionEntities[j].entity) {
+				const assignmentSubmission = this._submissionEntities[j].entity;
+				console.log(assignmentSubmission);
```
jeroach, line 112: nit: Can remove the `log`

components/submissions-page.js
```diff
@@ -0,0 +1,158 @@
...
+		const ext = filename.split('.')[1];
+		if (ext === 'png' || ext === 'jpg') {
+			return 'file-image';
+		}
+		else {
+			return 'file-document';
+		}
+	}
+
+	_getReadableFileSizeString(fileSizeBytes) {
```
jeroach, line 97: @AlexBedley thoughts on these helpers moving to a common location? seems like the kind of logic that we might be able to reuse elsewhere.

Again, 100% fine for the POC - just a thought as we move forward

components/consistent-evaluation-page.js
zhangium: I'm not sure what the status of the visual diff bot is after Alex was looking into it so I just added back that span to revert it back to what it was originally for now.

jeroach: Oh yeah that's a good call to hold off for now.

zhangium: Actually should I be writing visual diff tests for the POC? I think having its own page would make sense like right panel has in that case.

AlexBedley: vdiff bot is good, I reverted my change

AlexBedley: don't write vdiff tests for this change, let's wait until the dust settles and we (us and design) converge on a UX

components/submissions-page.js
AlexBedley: I think in general having helper classes makes sense. For this particular piece of data, I wonder if file size strings are localizable

components/submissions-page.js
AlexBedley: agreed, it's a hassle to change component/file names after they've been created

components/submissions-page.js
AlexBedley: yeah wouldn't want to go to prod with this as-is but obviously that's not the point of this PR

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -0,0 +1,168 @@
...
+		let i = -1;
+		const byteUnits = [' kB', ' MB', ' GB'];
+		do {
+			fileSizeBytes = fileSizeBytes / 1024;
+			i++;
+		} while (fileSizeBytes > 1024);
+		return Math.max(fileSizeBytes, 0.1).toFixed(1) + byteUnits[i];
+	}
+
+	_getSubmissionDisplayHtml(displayNumber, date) {
```
jeroach, line 107: I love these helper methods for the HTML building - so much easier to read and understand (and split out in future) 👍

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -0,0 +1,168 @@
...
+					builder += this._getCommentDisplayHtml(comment);
+				}
+			}
+		}
+		const list = document.createElement('d2l-list');
+		list.innerHTML = builder;
+		return list;
+	}
+
+	render() {
```
dbatiste, line 161: Just a quick FYI: it looks like `eslint` and the `eslint-config-brightspace` are out of date in this repo.  

There was a bug in the class member sort config that was preventing sort order from being linted properly, which was fixed in a more recent version.  Assuming those dependencies are eventually updated for this repo, there will be some reordering required to address those linting errors.  

If you are wondering what that order is, it can be seen [here](https://github.com/Brightspace/eslint-config-brightspace/blob/master/lit-config.js#L4).

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -0,0 +1,168 @@
...
+				}
+				let comment = '';
+				if (assignmentSubmission.getSubEntityByClass('submission-comment')) {
+					comment = assignmentSubmission.getSubEntityByClass('submission-comment').properties.html;
+					builder += this._getCommentDisplayHtml(comment);
+				}
+			}
+		}
+		const list = document.createElement('d2l-list');
+		list.innerHTML = builder;
```
dbatiste, line 157: Is there a particular reason to build up the HTML as a string?

The recommended approach is to compose an HTML template result, and that can be done with a combination of nested `html` template results in a couple different ways, with support for looping [as described in the docs](https://lit-html.polymer-project.org/guide/template-reference#loops-with-array.map).

Simplified for brevity, but the templates can be composed however necessary.
```
render() {
    return html`<d2l-list>
        ${this._renderListItems()}
    </d2l-list>`;
}

_renderListItems() {
    return this._submissionEntities.map((entity) => html`<d2l-list-item>
        <d2l-list-item-content>
            <div class="d2l-label-text">Comments</div>
            <div slot="secondary">${entity.comment}</div>
        </d2l-list-item-content>
    </d2l-list-item>`);
}
```
Generally speaking, there are security concerns with using `innerHTML` to compose.  It is possible to insert unsafe HTML with `lit-html` templates as well, but one must intentionally use the unsafe directive to do so.  

Another benefit of composing with lit-html templates is that it keeps track of properties used in those template results, and when those properties change, only those parts will be rerendered to the DOM.  lit-element and lit-html take care of constructing the actual DOM nodes as part of it's life cycle, which we don't want to circumvent.

jeroach: Thanks @dbatiste there isn't a particular reason - Annie did ask me how to do this, and I could only remember the Polymer way (it's actually been a while since I've been in Lit code).

We figured we'd cross that bridge when we componentized the pieces, but as you've shown it doesn't look too hard.

Thanks! @zhangium want to give this approach a shot?

dbatiste: Sounds great Josh.  If you have any questions, feel free to reach out.  :)  The Lit ([lit-element](https://lit-element.polymer-project.org/guide/start) and [lit-html](https://lit-html.polymer-project.org/guide/getting-started)) docs are pretty good references as well.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	_renderListItems() {
+		const itemTemplate = [];
+		for (let i = 0; i < this._submissionEntities.length; i++) {
+			if (this._submissionEntities[i].entity) {
+				const assignmentSubmission = this._submissionEntities[i].entity;
+				itemTemplate.push(html`
+				<consistent-evaluation-submission-item
+				.assignmentSubmission=${assignmentSubmission}
```
jeroach, line 67: nit: indent the properties for readability

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	_renderTitle() {
+		return html`<h3 class="first-item">Submission ${this.displayNumber}</h3>\n
+		<p class="submission-info">${this._date.toDateString()} ${this._date.toLocaleTimeString()}<d2l-icon icon="tier1:dot"></d2l-icon>${this.evaluationState}</p>`;
+	}
+
+	_renderAttachments() {
+		return html`${this._attachments.map((attachment) => html`
+		<d2l-list-item>
```
jeroach, line 101: nit: indent the elements for readability

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+		<a href="${attachment.properties.href}"><span>${attachment.properties.name}</span></a>
+		<div slot="secondary">${this._getReadableFileSizeString(attachment.properties.size)}</div></div>
+		</d2l-list-item-content>
+		</d2l-list-item>\n`)}
+		`;
+	}
+
+	_renderComment() {
+		if (this._comment) {
+			return html`<d2l-list-item><d2l-list-item-content>
```
jeroach, line 114: nit: indent the elements for readability

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+			<div slot="secondary">${unsafeHTML(this._comment)}</div>
+			</d2l-list-item-content>
+			</d2l-list-item>`;
+		}
+		return html``;
+	}
+
+	render() {
+		return html`
+		${this._renderTitle()}
```
dbatiste, line 125: This doesn't look quite right... we would expect a `d2l-list` to contain `d2l-list-item` only.  In this case though, `_renderTitle()` is returning an `h3` and a `p`.  My initial assumption was that this component would return a `d2l-list-item` based on how it is named, and how it is being used in the page component below.  I'm not that familiar with the goal here though.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+		const byteUnits = [' kB', ' MB', ' GB'];
+		do {
+			fileSizeBytes = fileSizeBytes / 1024;
+			i++;
+		} while (fileSizeBytes > 1024);
+		return Math.max(fileSizeBytes, 0.1).toFixed(1) + byteUnits[i];
+	}
+
+	_renderTitle() {
+		return html`<h3 class="first-item">Submission ${this.displayNumber}</h3>\n
```
dbatiste, line 95: Should removed these `\n`s.  If we need to move down to a new line, we should make it a block element (which `p` already is).

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+
+	render() {
+		return html`
+		${this._renderTitle()}
+		${this._renderAttachments()}
+		${this._renderComment()}
+		`;
+	}
+}
+customElements.define('consistent-evaluation-submission-item', ConsistentEvaluationSubmissionItem);
```
dbatiste, line 131: This should be `d2l-consistent-evaluation-submission-item`, since all of our components are prefixed with `d2l-`.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+		</d2l-list-item-content>
+		</d2l-list-item>\n`)}
+		`;
+	}
+
+	_renderComment() {
+		if (this._comment) {
+			return html`<d2l-list-item><d2l-list-item-content>
+			<div class="d2l-label-text">Comments</div>
+			<div slot="secondary">${unsafeHTML(this._comment)}</div>
```
dbatiste, line 116: Is `unsafeHTML` required here?  The reason I ask (besides being unsafe) is that `d2l-list-item-content` is designed with a particular spacing and style in mind.  It imposes certain spacing and font-size for the primary and secondary message, so I wouldn't be expecting user authored HTML here... unless this is a new use case...?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+			evaluationState : { type: String }
+		};
+	}
+
+	static get styles() {
+		return [heading4Styles, labelStyles, css`
+		.submission-info{
+			margin: 0;
+			padding: 0;
+			font-size: 0.7rem;
```
dbatiste, line 23: It looks like the submission info element should use the `bodySmallStyles` from [typography](https://github.com/BrightspaceUI/core/blob/master/components/typography/styles.js#L25), rather than defining custom typography here.  There is one difference however, the standard typography uses the tungsten color for this size, which we would _probably_ want for consistency anyway.  It also applies standard responsive typography media query.

Other note: if a `div` is used instead of a `p`, then it will not be necessary to zero out the margin and padding.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+		return {
+			assignmentSubmission : { type: Object },
+			displayNumber : { type: Number },
+			dateStr : { type: String },
+			evaluationState : { type: String }
+		};
+	}
+
+	static get styles() {
+		return [heading4Styles, labelStyles, css`
```
dbatiste, line 19: Looks like these heading styles are being included, but not used.  Maybe the intent was to apply `d2l-heading-4` to the `h3`?  Normally we would always apply one of our D2L typography heading styles to that heading.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -0,0 +1,131 @@
...
+			letter-spacing: 0.2px;
+		}
+		.first-item {
+			padding-top: 1rem;
+			margin: 0;
+		}
+		.item {
+			display: inline-block;
+			vertical-align: middle;
+			padding-left: 0.5rem;
```
dbatiste, line 34: I think this would need to be flipped for RTL.  To do so:
```
import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';

class MyElement extends RtlMixin(LitElement) {

    static get styles() {
        return css`
            ...
            :host([dir="rtl"]) {
                padding-left: 0;
                padding-right: 0.5rem;
            }
        `;
    }

}
```

components/left-panel/consistent-evaluation-submission-item.js
zhangium: Basically I was running into the issue where the comment html would show up as the comment text instead of the properly formatted comment. Is there a better way of doing this?

dbatiste: Assuming the content is trusted, think may be "ok".  The content component is really just a helper component that tries to apply some consistent typography and spacing, for a primary and secondary message. 

If the content being inserted is user-authored, it may have all sorts of crazy formatting, which may defeat the purpose of using the content component.  If this component buys a little consistency that it's all good, but you also don't have to use it.

components/left-panel/consistent-evaluation-submission-item.js
zhangium: I didn't add this in since I couldn't find any other references in the repo for rtl. Maybe that would be its own story

components/left-panel/consistent-evaluation-submission-item.js
zhangium: I did mean to use the heading styles, thank you for catching this!

PR #39, opened 2020-06-08
Vy/prototype dont merge jun8

vix633:
Disregard branch name :|


components/consistent-evaluation-page.js
```diff
@@ -148,32 +148,38 @@ export default class ConsistentEvaluationPage extends LitElement {
...
 	async _transientSaveFeedback(e) {
+		const entity = await this._controller.fetchEvaluationEntity(false);
 		const newFeedbackVal = e.detail;
-		this.evaluationEntity = await this._controller.transientSaveFeedback(this.evaluationEntity, newFeedbackVal);
+		this.evaluationEntity = await this._controller.transientSaveFeedback(entity, newFeedbackVal);
 	}
 
-	async _transientSaveGrade(e) {
+	async _transientSaveGrade(e) { 
+		const entity = await this._controller.fetchEvaluationEntity(false);
```
AlexBedley, line 157: the smallest of nits: its hard to understand what `false` means in these calls. It might be more readable to have a constant like `DONT_BYPASS_CACHE` or something

components/right-panel/consistent-evaluation-feedback-actions.js
```diff
@@ -62,7 +62,7 @@ class FeedbackActions extends LocalizeMixin(LitElement) {
 
 	constructor() {
 		super();
-
+		this.showAll = true;
```
AlexBedley, line 65: I vaguely remember something about setting the defaults for booleans to `true` is dangerous but that might have been a polymer thing

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -1,12 +1,16 @@
 import 'd2l-activities/components/d2l-activity-editor/d2l-activity-text-editor.js';
 import 'd2l-activities/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-editor.js';
+import 'd2l-activities/components/d2l-activity-editor/d2l-activity-attachments/d2l-activity-attachments-list.js';
 import './consistent-evaluation-right-panel-block';
+import '@brightspace-ui/core/components/button/button.js';
+import '@brightspace-ui/core/components/button/button-icon.js';
```
AlexBedley, line 6: I _think_ these 2 imports are no longer needed

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -48,8 +52,11 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 			detail: feedback
 		}));
 	}
-
+	async saveAttachment(){
+		this.shadowRoot.querySelector('d2l-activity-attachments-editor').save();
```
AlexBedley, line 56: I wonder what happens if this fails. Does it return any information or does it just throw?

components/right-panel/consistent-evaluation-feedback-actions.js
AlexBedley: @Scy-D2L do you remember something around this topic?

dbatiste: There are certain instances where the Lit linter/analyzer will error when using a boolean that is `undefined`.  Initializing them in the constructor should be fine for Lit elements, as it is well before any attribute values are assigned or life cycle methods called.

zhangium: I know there's this: https://blog.d2l.dev/2020/05/11/LitElement-Best-Practices-Gotchas/#-do-default-boolean-properties-to-false
Best practices say that lit properties with attributes should default to false.

vix633: Cool, i've made the change. Afaik this file is not actually used.

PR #40, opened 2020-06-18
Frontend save/publish

zhangium:
- Created a "super-controller" for consistent evaluation (`ConsistentEvaluationController.js`) to house save/publish methods 
- Got rid of redundant methods from other controllers. 

This change also depends on a change in the @brightspace-ui-labs/grade-result repo: https://github.com/BrightspaceUILabs/grade-result/pull/25 so visual tests are failing

This PR corresponsds to Rally [US117190](https://rally1.rallydev.com/#/42960320374d/detail/userstory/395295885516)


components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -2,6 +2,7 @@ import './consistent-evaluation-feedback.js';
 import './consistent-evaluation-outcomes.js';
 import './consistent-evaluation-rubric.js';
 import './consistent-evaluation-grade-result.js';
+import '../footer/consistent-evaluation-footer.js';
```
zhangium, line 5: I didn't need this import, will remove

test/FooterController.test.js
```diff
@@ -1,222 +1,143 @@
...
-			const entity = await controller.requestEvaluationEntity(true);
-
-			assert.equal(entity.properties.state, 'Published');
-
-			controller._requestEvaluationEntity.restore();
-		});
-	});
-
-	describe('publish', () => {
-		it('can publish when action is present', async() => {
```
zhangium, line 146: these tests got "relocated" to the ConsistentEvaluationController tests

test/ConsistentEvaluationFeedbackController.test.js
```diff
@@ -112,131 +112,12 @@ describe('ConsistentEvaluationFeedbackController', () => {
...
 					assert.fail();
 				} catch (e) {
 					assert.equal(e.message, FeedbackControllerErrors.REQUEST_FAILED);
 				}
 			});
 		});
 	});
-
-	describe('updateFeedbackText', () => {
-		describe('proper feedback object should save properly', () => {
```
zhangium, line 125: these tests got "relocated" to the ConsistentEvaluationController tests

components/consistent-evaluation-page.js
```diff
@@ -50,25 +144,32 @@ export default class ConsistentEvaluationPage extends LitElement {
 				</div>
 				<div slot="secondary">
 					<consistent-evaluation-right-panel
+						.evaluationHref=${this.evaluationHref}
 						rubricHref=${this.rubricHref}
 						rubricAssessmentHref=${this.rubricAssessmentHref}
 						outcomesHref=${this.outcomesHref}
 						gradeHref=${this.gradeHref}
```
jeroach, line 151: Do we need to pass the `gradeHref` as a value as well?

components/consistent-evaluation.js
```diff
@@ -45,14 +45,14 @@ export class ConsistentEvaluation extends MobxLitElement {
 
 	render() {
 		return html`
-			<d2l-consistent-evaluation-page
+			<d2l-consistent-evaluation-page id='zero'
```
jeroach, line 48: Is this `id` needed? Can we use something more descriptive if so?

components/controllers/ConsistentEvaluationController.js
```diff
@@ -0,0 +1,136 @@
...
+		const evaluationResource = await this._fetchEvaluationEntity(bypassCache);
+		if (!evaluationResource || !evaluationResource.entity) {
+			throw new Error(ConsistentEvaluationControllerErrors.ERROR_FETCHING_ENTITY);
+		}
+		const evaluationEntity = evaluationResource.entity;
+
+		return evaluationEntity;
+	}
+
+	async __performSirenAction(action, field = null) {
```
jeroach, line 55: Looks like two underscores in the method name - likely only meant one

components/footer/consistent-evaluation-footer.js
```diff
@@ -1,107 +1,37 @@
-import './consistent-evaluation-footer-presentational.js';
-import { html, LitElement } from 'lit-element';
-import { ConsistentEvaluationFooterController } from '../controllers/FooterController.js';
```
jeroach, line 3: If `consistent-evaluation-footer` isn't using the FooterController, what is?

components/consistent-evaluation-page.js
```diff
@@ -35,6 +36,99 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+
+	get evaluationEntity() {
+		return this._evaluationEntity;
+	}
+
+	set evaluationEntity(entity) {
+		const oldVal = this.evaluationEntity;
+		if (oldVal !== entity) {
+			this._evaluationEntity = entity;
+			this.evaluationHref = entity.links[1].href;
```
AlexBedley, line 56: we should be fetching links (or hrefs) by rel (or class but I think rel makes more sense here) and not using static index lookup. See https://github.com/Brightspace/node-siren-parser#entitygetxbyystringregexp-key

components/footer/consistent-evaluation-footer.js
zhangium: I think we can actually remove the FooterController and tests. It was essentially replaced with the "supercontroller" ConsistentEvaluationController. The only thing it could be used for was fetching the evaluation state to change the text in the presentation footer buttons, and I pass in an evaluation entity with that property instead now.

components/consistent-evaluation-page.js
```diff
@@ -35,6 +37,107 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+			if (this._evaluationHref && this._token) {
+				this._initializeController().then(() => this.requestUpdate());
+			}
+		}
+	}
+
+	async _initializeController() {
+		this._controller = new ConsistentEvaluationController(this._evaluationHref, this._token);
+		this.evaluationEntity = await this._controller.fetchEvaluationEntity();
+		console.log(this.evaluationEntity.getSubEntityByRel('grade'));
```
jeroach, line 92: nit: can drop the `console.log`

components/consistent-evaluation-page.js
jeroach: Is this still outstanding or do we not need it?

components/consistent-evaluation-page.js
zhangium: oops, thought I got rid of them all!

components/consistent-evaluation-page.js
zhangium: I think this comment exists on an outdated version of the file. I removed both grade and feedback hrefs from consistent-evaluation-page

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -38,14 +35,37 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
...
+		this._token = undefined;
 
 		this.hideRubric = false;
 		this.hideGrade = false;
 		this.hideFeedback = false;
 		this.hideOutcomes = false;
 	}
 
+	async _transientSaveFeedback(e) {
+		this._emitTransientSaveEvent('on-transient-save-feedback', e.detail);
```
AlexBedley, line 48: we should prefix event names with our component names: https://blog.d2l.dev/2020/05/11/LitElement-Best-Practices-Gotchas/#-do-qualify-events-with-the-component-name

components/right-panel/consistent-evaluation-feedback.js
```diff
@@ -55,12 +55,8 @@ class ConsistentEvaluationFeedback extends LitElement {
...
 		this._controller = new ConsistentEvaluationFeedbackController(this._href, this._token);
 
-		this._feedbackEntity = await this._controller.requestFeedback();
-		this._feedbackText = this._feedbackEntity.entities[0].entities[0].properties.text;
-	}
-
-	_saveFeedback(feedback, entity) {
-		this._controller.updateFeedbackText(feedback, entity);
+		this._feedbackEntity = await this._controller.requestFeedbackEntity();
+		this._feedbackText = this._feedbackEntity.properties.text || this._feedbackText;
```
AlexBedley, line 59: How come we fallback to the old `_feedbackText`? What if the new text is just empty?

components/footer/consistent-evaluation-footer.js
```diff
@@ -1,107 +1,37 @@
...
-customElements.define('d2l-consistent-evaluation-footer', ConsistentEvaluationFooter);
+import './consistent-evaluation-footer-presentational.js';
+import { html, LitElement } from 'lit-element';
+import { publishedState } from '../controllers/constants.js';
+
+export class ConsistentEvaluationFooter extends LitElement {
+
+	static get properties() {
+		return {
+			evaluationEntity: { type: Object },
```
AlexBedley, line 9: In my mind this makes this component "know too much". It really only cares about IsEntityPublished. I would just determine the published state in a parent's controller and pass that boolean down to this component instead of having this component know what "published" or "entity" means.

If we do go with that change then I think we can just delete this component and use `d2l-consistent-evaluation-footer-presentational` directly.

components/consistent-evaluation-page.js
```diff
@@ -35,6 +37,107 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+		this.evaluationState = this.evaluationEntity.properties.state;
+	}
+
+	_hasFeedbackComponent() {
+		return this.evaluationEntity && this.evaluationEntity.getSubEntityByRel('feedback') === undefined;
```
AlexBedley, line 97: I could be misunderstanding this, but if the feedback subentity is undefined, wouldn't that mean we don't have a feedback component? Seeing where it's used (as a hidden value) I think maybe just the name needs to change.

components/consistent-evaluation-page.js
```diff
@@ -50,25 +153,30 @@ export default class ConsistentEvaluationPage extends LitElement {
 				</div>
 				<div slot="secondary">
 					<consistent-evaluation-right-panel
+						.evaluationHref=${this.evaluationHref}
```
AlexBedley, line 156: This is the design problem I was talking about during our 1-1s around composable controllers.

I think passing the href down below the controller level (which is this class right now) is not the design we want to go with. My expectations would be that below the controller level, we're just dealing with normal js objects and those children components (right panel, but specifically feedback and grades) don't know anything about hypermedia, or entities etc. (see my comment about the feedback component). So ideally this class, being the controller class, only contains presentational children components.

I realize this might cause a bit of problems with the grade component, since if we just use https://github.com/BrightspaceUILabs/grade-result/blob/master/src/components/d2l-grade-result-presentational.js then we'll have to re-implement the controller behavior here. 

Ideally we can just import the grades controller https://github.com/BrightspaceUILabs/grade-result/blob/master/src/controller/GradesController.js and merge it's functionality into our "consistent evaluation controller".

Our other option is we just say "nobody will ever use grades/feedback outside of the concept of an evaluation". This makes our lives much easier, as we can just lump all the grades/feedback stuff into our controller and delete the individual grades/feedback controllers (since they will never be used on their own).

This is definitely the most important decision we will make in this PR and we'll be setting a trend for how nimbus, and ideally d2l, deals with evaluation in the front end. Now that we have a POC that we can point at, it might make our discussions around composable controllers easier to understand. I'd be happy to jump on a call to discuss this decision.

jeroach: Is this on hold then until next week when the API guys are back?

zhangium: yes I think we should discuss this in more detail next week.

components/consistent-evaluation-page.js
```diff
@@ -20,7 +22,11 @@ export default class ConsistentEvaluationPage extends LitElement {
 			rubricReadOnly: { type: Boolean },
 			richTextEditorDisabled: { type: Boolean },
 			submissionList: { type: Array },
-			evaluationState: { type: String }
+			evaluationState: { type: String },
+			feedbackText: { attribute: false },
+			grade: { attribute: false },
+			_feedbackEntity: { attribute: false },
+			_gradeEntity: { attribute: false }
```
zhangium, line 29: Internal properties for feedback and grade

components/controllers/ConsistentEvaluationController.js
```diff
@@ -0,0 +1,153 @@
...
+				}
+			} else {
+				return await this._performSirenAction(action);
+			}
+		} else {
+			throw new Error(ConsistentEvaluationControllerErrors.ACTION_NOT_FOUND(actionName));
+		}
+	}
+
+	parseGrade(entity) {
```
zhangium, line 80: This is from the grade-result repo, basically just needed this one method from the grades controller copied over

components/controllers/ConsistentEvaluationFeedbackController.js
```diff
@@ -1,6 +1,4 @@
 import 'd2l-polymer-siren-behaviors/store/entity-store.js';
```
zhangium, line 1: Should maybe remove unused feedback stuff in a separate PR

components/footer/consistent-evaluation-footer.js
```diff
@@ -1,107 +1,37 @@
...
-				@on-retract=${this._onRetractClick}
-				@on-update=${this._onUpdateClick}
-				@on-next-student=${this._onNextStudentClick}
-			></d2l-consistent-evaluation-footer-presentational>
-		`;
-	}
-}
-
-customElements.define('d2l-consistent-evaluation-footer', ConsistentEvaluationFooter);
+import './consistent-evaluation-footer-presentational.js';
```
zhangium, line 1: Should maybe remove unused footer stuff in a separate PR

components/right-panel/consistent-evaluation-feedback.js
zhangium, line 59: Should maybe remove unused feedback stuff like this file in a separate PR

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -1,56 +1,101 @@
...
+
+		// handle when grade is not yet initialized on the server
+		if (gradeType === GradeType.Number && score === null) {
+			score = 0;
+		} else if (gradeType === GradeType.Letter && score === null) {
+			score = '';
+		}
+		return html`
+			<d2l-consistent-evaluation-right-panel-block title="Overall Grade">
+			<d2l-labs-d2l-grade-result-presentational
```
zhangium, line 71: Using presentational instead of the regular grade-result to get rid of need for href and token

components/consistent-evaluation-page.js
jeroach: I don't think we need to specify internal properties here (e.g. `_feedbackEntity` or `_gradeEntity`) since they will never be passed in from the parent.

Maybe I've got the JS "logic" wrong but I was under the impression that the only thing needed to be defined in `properties` is stuff that will be either set from the parent, or used outside the class itself. @AlexBedley maybe you can confirm/correct my understanding? Thanks!

components/consistent-evaluation-page.js
```diff
@@ -35,6 +41,141 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+
+	_isEvaluationPublished() {
+		if (!this.evaluationEntity) {
+			return false;
+		}
+		return this.evaluationEntity.properties.state === publishedState;
+	}
+
+	_onNextStudentClick() {
+		this.dispatchEvent(new CustomEvent('next-student-click', {
```
jeroach, line 143: Should this event name have the `d2l-consistent-evaluation` prefix?

components/controllers/ConsistentEvaluationController.js
jeroach: Does this need to be `async` as well? All the other methods in the controller are.

jeroach: err- I guess not - it's not calling out to anything else

components/right-panel/consistent-evaluation-feedback.js
jeroach: This comment is still outstanding - thoughts?

components/right-panel/consistent-evaluation-feedback.js
```diff
@@ -69,10 +65,18 @@ class ConsistentEvaluationFeedback extends LitElement {
 		this._debounceJobs.feedback = Debouncer.debounce(
 			this._debounceJobs.feedback,
 			timeOut.after(500),
-			() => this._saveFeedback(feedback, this._feedbackEntity)
+			() => this._emitFeedbackEditEvent(feedback)
 		);
 	}
 
+	_emitFeedbackEditEvent(feedback) {
+		this.dispatchEvent(new CustomEvent('on-feedback-edit', {
```
jeroach, line 73: Event prefix (same as other comment)

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -1,56 +1,101 @@
...
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+
+export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
+
+	static get properties() {
+		return {
+			grade: { type: Object },
+			customManualOverrideText: { type: String },
+			customManualOverrideClearText: { type: String },
+			_labelText: { type: String },
```
jeroach, line 14: Similar to my other comment, do we need to define private variables in `properties`?

components/consistent-evaluation-page.js
zhangium: That makes sense. Would it be acceptable declaring initial values in the constructor?

components/right-panel/consistent-evaluation-feedback.js
zhangium: I decided to no longer use feedback (just using feedback-presentational component now) so I didn't bother changing it here

components/consistent-evaluation-page.js
AlexBedley: I'm not 100% sure when you can avoid listing members as `properties`. For me the main consideration is if we need to do "stuff" (like render) based on that member changing. I suspect in this case we want to re-render if `_gradeEntity` changes (it looks like `grade` depends on `_gradeEntity`). I'm fine erring on the side of caution and just listing them (it might incur a perf penalty 🤷 )

components/consistent-evaluation-page.js
```diff
@@ -86,21 +90,55 @@ export default class ConsistentEvaluationPage extends LitElement {
 		}
 	}
 
+	get feedbackText() {
+		if (this._feedbackEntity) {
```
AlexBedley, line 94: I think we can be safer and also check if `properties` exist

components/consistent-evaluation-page.js
jeroach: Yep, that's what I had in mind :)

components/right-panel/consistent-evaluation-feedback.js
AlexBedley: I would just delete this file then

components/consistent-evaluation-page.js
dbatiste: Alex is spot on here.   If the `render` output depends on some component state, then it should typically be represented in the in the static `properties` block so that the render life cycle will be triggered when that value changes.  Otherwise, one would have to manually request an update via `requestUpdate` (see lit docs).

jeroach: Thank you both for clearing up my misunderstanding. Looks like we need them after all.

components/right-panel/consistent-evaluation-grade-result.js
jeroach: As per other discussion, ignore this.

test/perceptual/consistent-evaluation-right-panel.visual-diff.html
```diff
@@ -1,116 +1,125 @@
...
+			import '@brightspace-ui/core/components/typography/styles.js';
+			import '@brightspace-ui/core/components/typography/typography.js';
+			import '../../components/right-panel/consistent-evaluation-right-panel.js';
+			import { Grade, GradeType } from '@brightspace-ui-labs/grade-result/src/controller/Grade';
+			const data = require('data/grade/grade.json');
+			const entity = JSON.parse(data);
+			const grade = new Grade(GradeType.Number, 47, 100, null, null, entity);
+			const panels = document.getElementsByTagName('consistent-evaluation-right-panel');
+			for (let i = 0; i < panels.length; i++) {
+				panels[i].setAttribute('grade', grade);
```
zhangium, line 18: @AlexBedley @jeroach I'm trying to attach a grade item to the right panel component grade attribute but this doesn't seem to be working, visual diff tests are still failing on travis CI. Is there a way to do this here so that we can fix the visual diff tests?

AlexBedley: I wonder if it's your test data file (grade.json). It looks like the result of a siren parse (instead of a clean entity). I don't think the sirenparser recognizes it's own output

PR #41, opened 2020-06-25
Submissions list hardening

zhangium:
- adding some UI changes to submissions POC
- add handling for displaying text submissions
- add submissions list demo page
- some controller testing in ConsistentEvaluationHrefController.test.js

### **File submissions list:**
image

### **Text submissions list:**
image


zhangium:
One note is that the "Unevaluated" status indicator should be the only one showing up, according to Daniel's mockups. So "Draft" and "Published" states shouldn't be showing up, like it is in this PR. But that requires comparing a submission's date vs an evaluation published date, which is not readily available in the current HM API for evaluation so I decided to not go through that trouble only to have to change it in the near future.

components/controllers/constants.js
```diff
@@ -1,13 +1,16 @@
...
+export const publishActionName = 'publish';
+export const retractActionName = 'retract';
+export const gradesRel = 'https://activities.api.brightspace.com/rels/grade';
+export const feedbackRel = 'https://activities.api.brightspace.com/rels/feedback';
+export const evaluationRel = 'https://activities.api.brightspace.com/rels/evaluation';
+export const evaluationSubEntityRel = 'https://api.brightspace.com/rels/evaluation';
+export const assessmentRel = 'https://api.brightspace.com/rels/assessment';
+export const rubricRel = 'https://api.brightspace.com/rels/rubric';
+export const assignmentRel = 'https://api.brightspace.com/rels/assignment';
+export const submissionsRel = 'https://assignments.api.brightspace.com/rels/assignment-submission-list';
```
AlexBedley, line 11: Not the most important thing but https://github.com/Brightspace/d2l-hypermedia-constants is where we've put things like this in the past

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -91,16 +91,34 @@ export class ConsistentEvaluationHrefController {
 
 	async getSubmissionInfo() {
 		let root = await this._getRootEntity(false);
-		let submissionList = undefined;
-		let evaluationState = undefined;
+		let submissionList, evaluationState, submissionType, dueDate;
+		const getHref = (root, rel) => {
```
AlexBedley, line 95: This function seems pretty useful, it probably belongs at the top scope of this class (it's used in another function above)

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -91,16 +91,34 @@ export class ConsistentEvaluationHrefController {
...
+				return root.getLinkByRel(rel).href;
+			}
+			return undefined;
+		};
 		if (root && root.entity) {
 			root = root.entity;
-			submissionList = root.getSubEntityByClass('assignment-submission-list').links;
-			evaluationState = root.getSubEntityByClass('evaluation').properties.state;
+			submissionList = root.getSubEntityByRel(submissionsRel).links;
+			evaluationState = root.getSubEntityByRel(evaluationSubEntityRel).properties.state;
```
AlexBedley, line 104: I notice some of the other properties (like `dueDate`) have checks but this one doesn't

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -39,49 +60,54 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
 
-	set dateStr(str) {
-		this._date = new Date(str);
-	}
-
-	_initializeAttachments() {
-		const attachmentsListEntity = this._assignmentSubmission.getSubEntityByClass('attachment-list');
+	_initializeSubmissionProperties() {
+		this._date = new Date(this.dateStr);
+		const attachmentsListEntity = this._submissionEntity.getSubEntityByClass('attachment-list');
```
AlexBedley, line 84: not a huge deal but we could use the getter here instead (just `this.submissionEntity`)

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -39,49 +60,54 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+		if (this.submissionEntity.getSubEntityByClass('submission-comment')) {
+			this._comment = this.submissionEntity.getSubEntityByClass('submission-comment').properties.html;
 		}
 	}
 
 	_getIcon(filename) {
-		const ext = filename.split('.')[1];
-		if (ext === 'png' || ext === 'jpg') {
+		const ext = this._getFileType(filename);
+		if (ext === 'PNG' || ext === 'JPG') {
```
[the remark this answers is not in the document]
zhangium: yeah that's a good question and you're right I think we support more than just png and jpg for images. I'm not sure how other places determine conditional file icons/if that's done. should maybe ask backend if the API can send us a doctype property or something?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -98,48 +124,112 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+	_renderStatus() {
+		if (this.late) {
+			return html`<d2l-status-indicator state="alert" text="Late" bold></d2l-status-indicator>`;
+		} else {
+			return html ``;
+		}
+	}
+
+	_renderEvaluationState() {
+		// TODO: compare submission date vs evaluation published date to determine if new submission
```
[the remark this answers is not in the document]
zhangium: I was making a note for later but I can remove it, basically the backend changes need to happen first before we can do anything on frontend

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -98,48 +124,112 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+		}
+	}
+
+	_addActionsForTextSubmission() {
+		if (this.submissionType === 'Text submission') {
+			return html`
+			<div slot="actions">
+				<d2l-button-icon icon="tier1:more"></d2l-button-icon>
+			</div>
 			</d2l-list-item>`;
```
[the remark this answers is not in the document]
zhangium: 😢

components/controllers/constants.js
zhangium: should we be putting all the constants there?

zhangium: I mean should we just be importing constants from that resource instead of our own constants file?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -39,49 +60,54 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+		const ext = this._getFileType(filename);
+		if (ext === 'PNG' || ext === 'JPG') {
 			return 'file-image';
 		}
 		else {
 			return 'file-document';
 		}
 	}
 
+	_getFileTitle(filename) {
```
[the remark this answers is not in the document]
zhangium: I think `path` is a server-side only library. Should that maybe get worked on in a different new work item?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -171,63 +220,84 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+	}
 
-			<div slot="actions">
-				<d2l-button-icon icon="tier1:more"></d2l-button-icon>
-			</div>
-			</d2l-list-item>`)}`;
+	_addActionsForFileSubmission(downloadHref) {
+		// Placeholder for menu presentational
+		return html`<div slot="actions" style="z-index: inherit;">
+			<d2l-dropdown-more text="More Options">
```
jeroach, line 243: Can we move the menu html template code to a helper so it isn't duplicated for file and text submissions?

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -86,16 +88,27 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
...
+						<d2l-consistent-evaluation-submission-item
+							dateStr=${submissionDate}
+							.displayNumber=${this._submissionEntities.length - i}
+							evaluationState=${this.evaluationState}
+							?late=${new Date(this.dueDate) < new Date(submissionDate)}
+							.submissionEntity=${submissionEntity}
+							submissionType=${this.submissionType}
+						></d2l-consistent-evaluation-submission-item>`);
+				} else {
+					itemTemplate.push(html`
```
jeroach, line 102: Is this `else` block possible? e.g. can a submission exist without a date?

I like the pre-check - my suggestion is that we can just omit the `else` and just render nothing if no date is found, since we never expect this to happen in practice.

zhangium: good point I think the date should always be there as a property if the submission is there

zhangium:
@jeroach thanks! I updated the screenshots in my initial first comment. Goldens will need to be regenerated but the diff is not significant since I didn't add in visual diff tests for submissions list. Just the "rendering full app" golden test is broken.

zhangium:
Need to update travis ci to do `npm ci` instead of `npm i` but I don't remember how to set this config @AlexBedley do you remember how you fixed this last time?

PR #44, opened 2020-07-10
Update slot name to supporting-info

ciss1995:
context:

BrightspaceUI/core#681
This updates the slot name secondary so supporting-info


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -225,7 +225,7 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 				${this._renderReadStatus(file.properties.read)}
 				<div class="consistent-eval-submission-attachment-item">
 					<span>${this._getFileTitle(file.properties.name)}</span>
-					<div slot="secondary" class="d2l-body-small">
+					<div slot="supporting-info" class="d2l-body-small">
```
dbatiste, line 228: Looks like `d2l-body-small` should not be needed here.  The component already applies these styles for this slot (that's actually one of it's main objectives).

ciss1995: Yup good catch! Thanks

PR #45, opened 2020-07-13
change to proper property and attribute binding

zhangium:
Consolidated naming convention best practices and a working consistent eval app.

For _strings_ and _bools_, pass using **attributes**:
`some-attribute`
`?some-bool`
For _objects_, pass using **property** bindings:
`.someProperty`


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -295,6 +295,7 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 	}
 
 	render() {
+		console.log(this);
```
Scy-D2L, line 298: Logging left in

PR #46, opened 2020-07-13
cleanup

zhangium:
- got rid of unused feedback and footer files (`consistent-evaluation-feedback.js` and `ConsistentEvaluationFeedbackController.js` and `ConsistentEvaluationFeedbackController.test.js` and `consistent-evaluation-footer.js`)
- updated feedback demo page to use `feedback-presentational` 
- got rid of intermediary event dispatching for transient saves, listening directly on `consistent-evaluation-page.js` now
- ~re-enabling 2 previously skipped visual diff tests after regenerating goldens (changes were expected for full app, and a matter of pixels in hiding-outcomes)~
- ~bump up grade-result version in `package.json` and `package-lock.json`~


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -260,8 +260,8 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 				<d2l-list-item>
 				<d2l-list-item-content>
 					${this._renderCommentTitle()}
-					<div slot="supporting-info">
-						<d2l-more-less height=${peekHeight} h-align>${unsafeHTML(this._comment)}</d2l-more-less>
+					<div slot="secondary">
```
zhangium, line 263: I know recently there was a change in core's `d2l-list-item-content` to use the `supporting-info` slot but it's not rendering properly here. Changed it back to the `secondary` slot since there doesn't seem to be a major difference.

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -49,13 +49,13 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 
 		this._debounceJobs.feedback = Debouncer.debounce(
 			this._debounceJobs.feedback,
-			timeOut.after(500),
+			timeOut.after(800),
```
zhangium, line 52: I bumped this up pretty arbitrarily, but I was experiencing weird debouncer behaviour if I typed somewhat slower than what I usually do and 800ms seems to be a better timeout after some experimentation.

components/left-panel/consistent-evaluation-submission-item.js
zhangium: @Scy-D2L maybe you have more context on the update Gaudi is making?

dbatiste: Hey Annie, can you provide a bit more info about how it's not rendering correctly?  From what I gather, previous usages of the `secondary` slot should have been `supporting-info`, but perhaps that wasn't the case for all.

fyi: @ciss1995

zhangium: That's interesting. When using `supporting-info` nothing within that div renders, as in I don't see it in the UI.

dbatiste: That sounds odd - do you have the latest version of `core`?

It's possible that we changed this one in error.  It's not really clear when something should be `secondary` vs `supporting-info`.  If this still needs to be `secondary` then definitely switch it back.  😄 .  It would be good to know that the `supporting-info` slot works though.

ciss1995: Like Dave said, it could be related to the core version here https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/package-lock.json#L1103. Maybe try using the latest core version.

ciss1995: The new slot is available in `v1.58.0` https://github.com/BrightspaceUI/core/releases/tag/v1.58.0.

zhangium: Thanks for uncovering that we're not on the latest core version. I think I'll stick with `secondary` for now and change it when we upgrade core as a whole project.

m6jones: >Thanks for uncovering that we're not on the latest core version. I think I'll stick with secondary for now and change it when we upgrade core as a whole project.

Honestly, this statement scares me a bit. Core should be safe to upgrade minor and patch versions. Is there a reason this project is a bit scared to upgrade core?

m6jones: Also, using `d2l-more-less` within secondary tells me it should be `supporting-info` Secondary is not meant for description text.

zhangium: > Honestly, this statement scares me a bit. Core should be safe to upgrade minor and patch versions. Is there a reason this project is a bit scared to upgrade core?


@m6jones I meant that this PR was supposed to be small cleanup from another chunk of work and I didn't think upgrading should be done here. I feel I should have just left this file alone and addressed it in another PR. CC: @AlexBedley @jeroach what do you think?

jeroach: IHMO we should be on the latest of core, and I'd be fine with having it in this PR.

If that's a lot more work and it would be easier to do in a follow up, that works too.

So long as we get there soon :)

components/consistent-evaluation-page.js
```diff
@@ -188,7 +188,14 @@ export default class ConsistentEvaluationPage extends LitElement {
 	}
 
 	async _transientSaveGrade(e) {
-		const newGradeVal = e.detail;
+		let newGradeVal;
+		const type = e.detail.grade.scoreType;
```
jeroach, line 192: Is it safe to assume that `e.detail.grade` will always have a value?

zhangium: Good point. I do think it will always have a value since a default grade will always exist in consistent-eval-grade-result, but it doesn't make sense to transient save a default value

zhangium: Actually I go back on this. If a user really wants to give a default grade of 0, then so be it and they should be able to. `e.detail.grade` will always have a value because the `this.grade` value is always initialized in `consistent-evaluation-grade-result.js`

jeroach: Sounds good - I just wanted to ask the question :)

zhangium: Yep, still a good question

zhangium:
Reverted a bunch of commits related to re-enabling visual diff tests but those won't be relevant as per our team consensus to defer til later.

components/consistent-evaluation-page.js
```diff
@@ -188,7 +188,14 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+		if (type === GradeType.Letter) {
```
AlexBedley, line 193: I'm not a huge fan of consumers needing to care about what type of grade it is. I know this is just fallout from changing the controller to here though so no changes needed in this PR, but maybe we should change grades later

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -261,7 +261,7 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 				<d2l-list-item-content>
 					${this._renderCommentTitle()}
 					<div slot="supporting-info">
-						<d2l-more-less height=${peekHeight} h-align>${unsafeHTML(this._comment)}</d2l-more-less>
+						<d2l-more-less height=${peekHeight}>${unsafeHTML(this._comment)}</d2l-more-less>
```
AlexBedley, line 264: is this change because of feedback from Dan?

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -68,6 +68,7 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 				<d2l-activity-text-editor
 					.value="${this.feedbackText}"
 					.richtextEditorConfig="${this.richTextEditorConfig}"
+					@d2l-activity-text-editor-change="${this._saveOnFeedbackChange}"
```
AlexBedley, line 71: does this change mean it wasn't working before?

zhangium: It definitely was there when I merged the last PR, but I think it got lost in a merge conflict.

components/left-panel/consistent-evaluation-submission-item.js
zhangium: Nope, this was just cleanup because I realized that's not how the attribute is used.

PR #47, opened 2020-07-14
Best Practices Name Tweaks / Cleanup

Scy-D2L:
Cleaned up unused attributes, changed rich-text-editor-config to property name


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -37,7 +37,6 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 				type: Boolean
 			},
 			submissionEntity : {
-				attribute: 'submission-entity',
```
jeroach, line 40: Do we need `attribute: false` on these?

Scy-D2L: That is a good call, in theory that should stop us from trying to assign them as properties and ending up with strings

PR #48, opened 2020-07-15
Vy/video audio rich text

vix633:
image

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -30,7 +30,7 @@ export default class ConsistentEvaluationPage extends LitElement {
 				type: Boolean
 			},
 			richtextEditorConfig: {
-				attribute: 'rich-text-editor-config',
+				attribute: false,
 				type: Object
 			},
 			rubricAssessmentHref: {
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -29,6 +29,10 @@ export default class ConsistentEvaluationPage extends LitElement {
 				attribute: 'rich-text-editor-disabled',
 				type: Boolean
 			},
+			richtextEditorConfig: {
+				attribute: 'rich-text-editor-config',
+				type: Object
+			},
 			rubricAssessmentHref: {
 				attribute: 'rubric-assessment-href',
 				type: String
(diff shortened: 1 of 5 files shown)


components/consistent-evaluation-page.js
```diff
@@ -29,6 +29,10 @@ export default class ConsistentEvaluationPage extends LitElement {
 				attribute: 'rich-text-editor-disabled',
 				type: Boolean
 			},
+			richtextEditorConfig: {
+				attribute: 'rich-text-editor-config',
```
jeroach, line 33: I think we want `attribute: false` here since it's being set from within the class.

As opposed to being passed in via `consistent-evaluation.js`

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -43,6 +43,10 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 		this.feedbackText = '';
 
 		this._debounceJobs = {};
+		this.addEventListener('d2l-request-provider',
```
jeroach, line 46: Do we need to also remove the event listener?

vix633: Not sure, why do you think we do?

jeroach: There was a note on it in the Lit element best practices:
https://blog.d2l.dev/2020/05/11/LitElement-Best-Practices-Gotchas/#-do-add-event-listeners-properly

However, taking a second look - this might just be for the `connectedCallback` case, so I think you're okay here.

dbatiste: Correct, this handler will be cleaned up when this element is garbage collected.  

The case for manually removing the handler is: when the listener is being wired up to an element that is outside this components DOM scope, like the `window` or the `body`.  If we don't, those elements will still be attached to those elements, and those handlers will still hold reference to `this` element, preventing this element from being garbage collected, thereby creating a memory leak.  Typically, wiring handlers up to those elements (like `window` and `body`) needs to be done in `connectedCallback` and `disconnectedCallback` because those like cycle methods offer opportunity to manage the handler.

vix633: Thanks guys, TIL.

PR #50, opened 2020-07-17
Jroach/us117189

jeroach:
[US117189](https://rally1.rallydev.com/#/detail/userstory/395295867204?fdp=true): [Evidence] Wrap evidence in a web component

This has been ongoing for too long, so wanted to put up what's here since it's a "complete" unit of work.

Evidence renders for multiple file types, and we can navigate back and forth between files.

Still need to figure out why the 2nd scroll bar is there - might just be a demo page thing since the latest BSI on the CDN has the fix.

Also need to confirm working with the CDN version of the Annotations Viewer when that's ready.

Screenshot:
image


components/left-panel/consistent-evaluation-evidence.js
```diff
@@ -0,0 +1,71 @@
...
+		};
+	}
+
+	static get styles() {
+		return css`
+			.d2l-consistent-evaluation-evidence-top-bar {
+				height: 2.7rem;
+			}
+			iframe {
+				height: calc(100% - 2.7rem);
```
jeroach, line 21: This is ugly and I'm not sure if it will work for all cases, but it seems to get the `iframe` to fill the remainder of the left panel.

components/left-panel/consistent-evaluation-evidence.js
```diff
@@ -0,0 +1,71 @@
...
+		return {
+			url: { type: String },
+			token: { type: String }
+		};
+	}
+
+	static get styles() {
+		return css`
+			.d2l-consistent-evaluation-evidence-top-bar {
+				height: 2.7rem;
```
jeroach, line 18: Height specified by Dan

components/left-panel/consistent-evaluation-evidence.js
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	static async getLocalizeResources(langs) {
+		return await loadLocalizationResources(langs);
+	}
+
+	constructor() {
+		super();
+
+		window.addEventListener('message', e => {
```
jeroach, line 34: Needed for Annotations Viewer - not sure if there is a better way to do this.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -224,7 +235,9 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 				</d2l-icon>
 				${this._renderReadStatus(file.properties.read)}
 				<div class="consistent-eval-submission-attachment-item">
-					<span>${this._getFileTitle(file.properties.name)}</span>
+					<a href="#" @click="${() => this._dispatchRenderEvidenceEvent(file.properties.fileViewer)}">
```
jeroach, line 238: @zhangium not sure if this should just be a `span` instead of an href since the whole area should be "clickable"

I might need to move the click handler around.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -6,6 +6,7 @@ import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/list/list.js';
 import '@brightspace-ui/core/components/list/list-item.js';
 import '@brightspace-ui/core/components/list/list-item-content.js';
+import '@brightspace-ui/core/components/button/button.js';
```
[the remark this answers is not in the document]
jeroach: Ah yep, artifact of when I first setup the viewer click to a button for testing. Will remove.

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,90 @@
+import 'd2l-polymer-siren-behaviors/store/entity-store.js';
```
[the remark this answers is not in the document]
jeroach: yep, another artifact

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -218,21 +229,24 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
-				${this._renderReadStatus(file.properties.read)}
-				<div class="consistent-eval-submission-attachment-item">
-					<span>${this._getFileTitle(file.properties.name)}</span>
-					<div slot="supporting-info">
-						${this._renderFlaggedStatus(file.properties.flagged)}
-						${this._getFileType(file.properties.name)}
-						<d2l-icon icon="tier1:dot"></d2l-icon>${this._getReadableFileSizeString(file.properties.size)}
+			<d2l-list-item 
+				class="consistent-eval-submission-attachment-item-container"
+				@click="${() => this._dispatchRenderEvidenceEvent(file.properties.fileViewer)}"
```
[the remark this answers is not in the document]
jeroach: Sadly it is - seems to be the only way I could find that allows the proper passing of a parameter (in this case the `fileViewer` url) to an on-click method

components/left-panel/consistent-evaluation-evidence.js
```diff
@@ -0,0 +1,71 @@
...
+		return html`
+			<div class="d2l-consistent-evaluation-evidence-top-bar">
+				<d2l-button-subtle 
+					text="${this.localize('backToUserSubmissions')}" 
+					icon="tier1:chevron-left"
+					@click="${this._dispatchBackToUserSubmissionsEvent}"
+				></d2l-button-subtle>
+			</div>
+			<iframe 
+				src="${this.url}"
```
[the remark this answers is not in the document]
jeroach: Good question - we're formulating the URL ourselves so it should be valid - but the UI should still fail gracefully - let me think some on this.

components/left-panel/consistent-evaluation-evidence.js
dbatiste: I don't know a lot about the current use case, but this the targetOrigin `*` is a red flag.  See [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Security_concerns).  Is this something you are worried about?

components/left-panel/consistent-evaluation-submission-item.js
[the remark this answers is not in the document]
dbatiste: That aside, I think the Lit linter has a rule that recommends against arrow functions in the render context like this.  The rationale is that a new function will be created every time the render life cycle runs. How much of a perf difference it makes, I am not sure, but the alternative is:
```
<d2l-list-item data-attachment-id="..." @click="${this._handleListItemClick}">...</d2l-list-item>
```
And then something along the lines of...
```
_handleListItemClick(e) {
    this._dispatchRenderEvidenceEvent(this._getFileViewer(e.target.attachmentId));
}
```
Again, not sure how much of an issue it is, just that the Lit linter complains about it.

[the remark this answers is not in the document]
jeroach: Thanks Dave, I'll work on this!

components/left-panel/consistent-evaluation-evidence.js
jeroach: Updated to `targetOrigin = http://s.brightspace.com` which does what it needs to do since the annotations viewer is on the CDN.

Any other checks needed to ensure proper security? @dbatiste

components/left-panel/consistent-evaluation-submission-item.js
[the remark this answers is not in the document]
jeroach: I've fiddled around with this for much of today without getting anywhere. I did experiment with alternatives in the past as well before setting on this.

If there are no strong objections, I'm inclined to leave this as is for now, and if the perf issue becomes apparent later on, then we can revisit.

Thoughts?

components/left-panel/consistent-evaluation-evidence.js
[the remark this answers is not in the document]
jeroach: @AlexBedley and I chatted offline about this - so long as we address the token posting concern from @dbatiste above, then we should be able to blindly trust the url

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,89 @@
...
+	set displayEvidence(newVal) {
+		const oldVal = this._displayEvidence;
+		if (oldVal !== newVal) {
+			this._displayEvidence = newVal;
+			this.requestUpdate('displayEvidence', oldVal);
+		}
+	}
+
+	connectedCallback() {
+		this.addEventListener('d2l-consistent-evaluation-submission-item-render-evidence', this._renderEvidence);
```
AlexBedley, line 49: I wonder if we can put this on the `<d2l-consistent-evaluation-submissions-page>` element using `@event` bindings. It would make it more clear where we expect the event to come from at least (and be more idiomatic Lit)

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -218,21 +228,25 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
-				<div class="consistent-eval-submission-attachment-item">
-					<span>${this._getFileTitle(file.properties.name)}</span>
-					<div slot="supporting-info">
-						${this._renderFlaggedStatus(file.properties.flagged)}
-						${this._getFileType(file.properties.name)}
-						<d2l-icon icon="tier1:dot"></d2l-icon>${this._getReadableFileSizeString(file.properties.size)}
+			<d2l-list-item 
+				class="consistent-eval-submission-attachment-item-container"
+				@click="${
+	// eslint-disable-next-line lit/no-template-arrow
```
AlexBedley, line 234: To avoid this, you could put `file.properties.fileViewer` as an attribute (maybe a `data-*` attribute?) and use `e.target` to retrieve it on-click. Not sure if that's _better_ but it would avoid the eslint override

components/left-panel/consistent-evaluation-evidence.js
```diff
@@ -0,0 +1,73 @@
...
+				frameborder="0" 
+				scrolling="no" 
+				?allowfullscreen="${this._allowFullscreen}"
```
AlexBedley, line 66: Should we just put `true` here? Will `_allowFullscreen` ever be false?

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,89 @@
...
+
+		this._evidenceUrl = undefined;
+		this._displayEvidence = false;
+	}
+
+	get displayEvidence() {
+		return this._displayEvidence;
+	}
+
+	set displayEvidence(newVal) {
```
AlexBedley, line 40: Not really related to this PR: I see this cookie cutter code everywhere in our repo. It feels so clunky - I can't believe Lit just doesn't re-render if properties change. Do we remember where this came from? Is there another way to do this?

components/left-panel/consistent-evaluation-submission-item.js
AlexBedley: Ah I see I missed a conversation about this above. Not a deal breaker either way

components/left-panel/consistent-evaluation-submission-item.js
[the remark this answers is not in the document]
dbatiste: It should be possible to avoid the arrow function in render, but yeah, I'm not certain it's _that_ big of a deal.

components/left-panel/consistent-evaluation-left-panel.js
dbatiste: It also looks like it should be private to this class - it's only used internally ya?  If we need a private getter/setter we would normally name with a leading underscore, but in this case, we should be able to just have a private property `_displayEvidence` defined in the static properties object.  

The `_evidenceUrl` could be defined there as well since the output of `render` depends on it.  For example, currently, if the url were to change but not at the same time as `_displayEvidence`, the `render` method would not get rerun.  If that's a possible scenario, then it should definitely be defined as it's own private property.

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,89 @@
...
+		this._evidenceUrl = e.detail.url;
+	}
+
+	_renderSubmissionList() {
+		this.displayEvidence = false;
+		this._evidenceUrl = undefined;
+	}
+
+	render() {
+		return html`${this.displayEvidence ?
```
dbatiste, line 71: Looking at this further...  it looks like we should be able to define `_evidenceUrl` as a private property in the static properties object (which will automatically cause the lifecycle to run), and do this here:
```
if (this._evidenceUrl) {
    return html`<d2l-consistent-evaluation-evidence .url=${this._evidenceUrl} .token=${this.token}></d2l-consistent-evaluation-evidence>`;
} else {
    return html`<d2l-consistent-evaluation-submissions-page ...`;
}
```
This will avoid the need for `displayEvidence` property.  Also, with the render defined in this way (as opposed to `return html ``${ someBool ? html``stuff`` : html``other stuff``}``;`) , we eliminate a `lit-html` binding - while not a big deal, but they certainly not free and I think the end result is simpler.

components/left-panel/consistent-evaluation-evidence.js
jeroach: oh good idea - will do

components/left-panel/consistent-evaluation-evidence.js
jeroach: That is what I had originally, but it failed the linter since it was a boolean attribute without the prefix `?`

This seemed to be the only way to make the linter happy, but if I missed something then I'm all ears :)

components/left-panel/consistent-evaluation-left-panel.js
jeroach: So, the event is being fired from the `<d2l-consistent-evaluation-submissions-item>` - I was under the impression that if an even needs to be fired up more than 1 level, then we need to use the callbacks:
https://blog.d2l.dev/2020/05/11/LitElement-Best-Practices-Gotchas/#-do-add-event-listeners-properly

if I'm misunderstanding, then of course I'll fix it - @dbatiste could you please either confirm or correct my understanding? Thanks!

components/left-panel/consistent-evaluation-evidence.js
dbatiste: Note that `allowFullscreen` is deprecated ([see Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe)).  They are suggesting that `allow="fullscreen"` should be used.  This might help with the linting/analyzer.  (also, is it the linter or the analyzer that fails?)

components/left-panel/consistent-evaluation-left-panel.js
jeroach: Thanks Dave! working on this now

components/left-panel/consistent-evaluation-left-panel.js
jeroach: @AlexBedley I agree it is ugly - luckily it will go away by using Dave's suggestion :)

Something we can keep in mind/consider on future PRs (or perhaps a clean up opportunity for the project)

components/left-panel/consistent-evaluation-evidence.js
jeroach: Ah it was indeed the analyzer:
https://travis-ci.com/github/BrightspaceHypermediaComponents/consistent-evaluation/jobs/363988572#L260

Regardless, sounds like we can remove it. Thanks!

components/left-panel/consistent-evaluation-left-panel.js
jeroach: Fixed!

components/left-panel/consistent-evaluation-evidence.js
jeroach: Done.

components/left-panel/consistent-evaluation-left-panel.js
dbatiste: Alex is correct here.  I think the best practices docs covers this, but let me know if there's something we can improve there and I will make it so.  :)

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,79 @@
...
+			d2l-consistent-evaluation-submissions-page {
+				width: 100%;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+
+		this._evidenceUrl = undefined;
```
dbatiste, line 36: This assignment is unnecessary.  The `this._evidenceUrl` will already be `undefined`.  You'd only need to initialize the string if you need to assign a value, like `null` or an actual string.

jeroach: Thanks for the explanation :) I'll get rid of it.

components/left-panel/consistent-evaluation-left-panel.js
jeroach: Fixed!

PR #52, opened 2020-07-21
submissions style + general update

zhangium:
Branch name not that relevant, as I didn't make too many changes for accessibility in the end.

- styling changes requested by Dan
- added more langterms for submissions list
- updated submission-item to display corresponding icons for all the different kinds of attached files
- updates to submissions-page demo page since latest merges changed how submissions-page works


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -58,20 +66,53 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
 			margin: 0;
 		}
 		.consistent-eval-submission-attachment-item {
 			display: inline-block;
 			vertical-align: middle;
-			padding-left: 0.5rem;
 		}
 		.consistent-eval-submission-attachment-item-container:hover {
-			background-color: var(--d2l-color-regolith);
+			background-color: red; /*not working*/
```
zhangium, line 80: made a new story to deal with this US117849

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -58,20 +66,53 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 			flex-direction: column;
 			position: relative;
 		}
+		d2l-list-item, d2l-list-item:hover {
```
jeroach, line 69: Does this `hover` bit also fall under US117849? or is it working as intended?

zhangium:
Leaving a comment with updated screenshots!
image
image

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -217,61 +272,74 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
 			<d2l-dropdown-more text="More Options">
 			<d2l-dropdown-menu id="dropdown">
-				<d2l-menu label="Options">
-					<d2l-menu-item-link text="Download" href="${downloadHref}"></d2l-menu-item-link>
-					<d2l-menu-item-link text="Mark as Read" href="https://en.wikipedia.org/wiki/Universe"></d2l-menu-item-link>
-					<d2l-menu-item-link text="Flag" href="https://en.wikipedia.org/wiki/Universe"></d2l-menu-item-link>
-					<d2l-menu-item-link text="Edit a Copy" href="https://en.wikipedia.org/wiki/Universe"></d2l-menu-item-link>
+				<d2l-menu>
+					<d2l-menu-item-link text="${this.localize('download')}" href="${downloadHref}"></d2l-menu-item-link>
+					<d2l-menu-item-link text="${oppositeReadState}" href="#"></d2l-menu-item-link>
```
dbatiste, line 310: What's the purpose of the `href="#"` here?

components/left-panel/consistent-evaluation-submission-item.js
dbatiste: per conversation in Slack - we want the `d2l-list-item` to handle this. it currently does not support JS action as the primary action, but if we do need to support that, we'll figure out how to support it so apps don't need to try and override hover styles, etc.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -58,20 +66,53 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+		.d2l-submission-attachment-icon-container-inner {
+			margin: 0.5rem;
+			position: relative;
+			top: 0;
+			left: 0;
+		}
+		.d2l-attachment-read-status {
+			color: var(--d2l-color-carnelian);
+			position: absolute;
+			top: 0px;
```
dbatiste, line 101: Nit: not units on `0` value dimensions.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -217,61 +272,74 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 	}
 
 	_renderAttachments() {
+		// href placeholder on list-item
 		return html`${this._attachments.map((file) => html`
-			<d2l-list-item class="consistent-eval-submission-attachment-item-container">
-			<d2l-list-item-content>
-				<d2l-icon class="consistent-eval-submission-attachment-item" icon="tier2:${this._getIcon(file.properties.name)}">
-				</d2l-icon>
+			<d2l-list-item href="#" class="consistent-eval-submission-attachment-item-container">
```
dbatiste, line 277: What's the reason for the `href="#"` here and below?

components/left-panel/consistent-evaluation-submission-item.js
zhangium: Yeah this will probably change as well. `d2l-list-item` looks like they have built-in hover styling but it's only for a variant of d2l-list-item.

components/left-panel/consistent-evaluation-submission-item.js
zhangium: I was using it as more of a placeholder and also to test screenreaders/a11y. Is this where you suggested to replace as `href="javascript:void(0);"`? I can change that in this PR @dbatiste

dbatiste: While `#` would make it keyboard accessible, normally this would scroll the page (to the top) in the same way that `#some-id` would focus and scroll to the element with that id, and to avoid that behavior, a handler would normally need to be attached to `preventDefault`.  

The `href="javascript:void(0);"` is better and may be the quickest way to get your code working, but it still doesn't convey the correct semantic.  i.e. it's going to be announced as a link, which unless navigating to a different page or moving to a fragment, is the incorrect thing to use.

If we determine that we definitely need to make the primary action a JS onclick (as opposed to putting a button in the actions slot), then we need to update `d2l-list-item` to render a `button` element or `role="button"` internally.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -103,11 +102,12 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 				const submissionEntity = this._submissionEntities[i].entity;
 				if (submissionEntity.getSubEntityByClass(Classes.assignments.submissionDate)) {
 					const submissionDate = submissionEntity.getSubEntityByClass(Classes.assignments.submissionDate).properties.date;
+					const evaluationState = submissionEntity.properties.evaluationStatus;
 					itemTemplate.push(html`
 						<d2l-consistent-evaluation-submission-item
 							date-str=${submissionDate}
 							display-number=${this._submissionEntities.length - i}
-							evaluation-state=${this.evaluationState}
```
AlexBedley, line 110: Should we also delete the `evaluationState` property now? (is it used?)

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -18,6 +17,15 @@ import { loadLocalizationResources } from '../locale.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
 import { unsafeHTML } from 'lit-html/directives/unsafe-html.js';
 
+export const fileTypes = {
```
AlexBedley, line 20: This list makes me a bit nervous - I'm worried some other component will have a slightly different list and we'll introduce some inconsistencies. Ideally the LMS would just tell us "this file is a video" etc. But this is probably outside the scope of this PR.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -217,61 +272,74 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+				`;
 		}
 		return html``;
 	}
 
 	_renderCommentTitle() {
 		if (this.submissionType === fileSubmission) {
-			return html`<div class="d2l-label-text">${this.localize('comments')}</div>`;
+			return html`
+			<div class="d2l-label-text" style="padding-top: 0.5rem;">
```
AlexBedley, line 340: I know historically we avoided in-line styles. I don't remember why (maybe just to organize the CSS??). I'm not suggesting to change this, more of an open question if anyone remembers why we haven't been using in-line styles

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -116,12 +157,10 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
-		}
-		else {
-			return 'file-document';
+		if (ext in fileTypes) {
+			return fileTypes[ext];
 		}
+		return 'file-document';
 	}
 
 	_getFileTitle(filename) {
```
AlexBedley, line 166: it's out of range so I can't comment on the specific line. But `_getFileTitle` and `_getFileType` seem super fragile - I'm pretty sure `.` is a valid character in file names (and its also not required). The LMS might have rules I'm not aware of though.

components/left-panel/consistent-evaluation-submission-item.js
dbatiste: I think there has been much discussion about this on the internets over the years.  

In general, I think there are several benefits of CSS over inlines styles such as improved separation of content from presentation; reuse/sharing of styles; download & cache styles once; media queries and other features; special styles for print; and perhaps most importantly selector specificity rules.  

Inlines styles can really bloat the HTML at times; only supports of subset of style capabilities (ex. no media queries, keyframes, et.al.); and basically trump everything in terms of specificity, which makes it difficult to override without the use of `!important`.

Performance is a bit of a debate.  There are trade-offs with both in terms of performance, and the end-result depends on a lot of factors.  What may be faster in one case may not be faster in another case, and browsers have different algorithms for matching selectors to DOM, so there are inevitable differences between them when it comes to performance.

That's not to say inline styles aren't useful though.  We definitely use them when we need to specify a style value that's calculated at runtime.  That being said, we tend to define all the styles in CSS when we can.

components/left-panel/consistent-evaluation-submission-item.js
dbatiste: should be able to use `lastIndexOf` to make this a little better.  still possible for no extension, or no known extension. fwiw: [file handling in MVC-land](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web/UI/Html/Files/FileUpload/Default/FileUploadManager.cs?r=51e73ae2#173), uses this approach.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -58,20 +66,53 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
...
+			width: 10px;
+			height: 10px;
+			padding: 0.2rem;
+		}
+		d2l-more-less p, ul {
+			margin: 0.5rem 0rem;
+		}
+		d2l-status-indicator {
+			text-transform: none;
+			margin-right: 0.5rem;
```
dbatiste, line 114: Presumably this will need to be flipped for RTL.  If so, remember to add `RtlMixin` for the class, and flip the required styles.  Same thing for others... like `right: -4px;` above.  Recommend checking this visually to identify any other layout issues.

```
:host([dir="rtl"]) d2l-status-indicator {
    margin-left: 0.5rem;
    margin-right: 0;
}
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -175,36 +218,48 @@ export class ConsistentEvaluationSubmissionItem extends LocalizeMixin(LitElement
 		return html`
 		<d2l-list-item>
 		<d2l-list-item-content>
-			<h3 class="d2l-heading-3">${this.localize('textSubmission')} ${this.displayNumber}${this._renderReadStatus(read)}</h3>
-			<div class="d2l-body-small">
+			<div class="d2l-submission-attachment-icon-container">
+				<h3 style="padding-right: 0.5rem;" class="d2l-heading-3">${this.localize('textSubmission')} ${this.displayNumber}</h3>
```
dbatiste, line 222: This padding will probably need to be flipped for RTL.  To address this, use a CSS instead (class such as `.d2l-heading-3` or some other selector)... ex.
```
.d2l-heading-3 {
    padding-right: 0.5rem;
}
:host([dir="rtl"]) .d2l-heading-3 {
    padding-left: 0.5rem;
    padding-right: 0;
}
```
I think this is a good example where CSS excels over inline styles, per the question/comment further below.

zhangium: Actually this heading specifically is somewhere I think the padding should stay on the right. For context, it's to maintain space between the little "read" icon and the title for text submissions (image below). Changing the padding direction moves the orange dot to cover the top of the text.
image

zhangium:
**Update**: added in humanized lateness and more options menu alignment

image

components/left-panel/consistent-evaluation-submission-item.js
Scy-D2L: Is the dot going to remain on the right-hand side when the text is RTL?

components/left-panel/consistent-evaluation-submissions-page.js
zhangium: it's still needed to be displayed on the submission item yes! we just fetch it from the submission entity properties now instead of passing it down from fetching the evaluation entity

components/left-panel/consistent-evaluation-submission-item.js
zhangium: Fixed this with styling! Dot should be on the left hand side when text is RTL.

Updated screenshots: 
image
image

components/left-panel/consistent-evaluation-submission-item.js
zhangium: updated this to CSS instead of inline style!

components/left-panel/consistent-evaluation-submission-item.js
zhangium: I'm not sure how this was handled for quick eval but yeah it maybe shouldn't be up to the frontend to do the logic.

zhangium:
> I share Alex's concerns about the file type. This mapping is stored in the LMS in db tables.

I see. I'll make an ask to have it available in the response


> I assume that the `href="#"` will be addressed in a follow-up PR.

Yes, there should be another task to address adding the menu actions

components/left-panel/consistent-evaluation-submission-item.js
AlexBedley: Thanks for the rundown Dave!

PR #53, opened 2020-07-23
Empty States - Us117461

Scy-D2L:
Added messages that display when a file/text assignment being evaluated has no submissions and when an on-paper/in-person assignment is being evaluated (as they have no evidence).


components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -1,9 +1,25 @@
...
+function getSubmissionTypeName(type) {
+	switch (type) {
+		case fileSubmission:
+			return 'fileSubmission';
+		case textSubmission:
+			return 'textSubmission';
+		case observedInPerson:
+			return 'observedInPerson';
+		case onPaperSubmission:
+			return 'onPaperSubmission';
```
jeroach, line 18: Should we have a default case here?

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -24,12 +40,46 @@ export class ConsistentEvaluationLeftPanel extends LitElement {
 			d2l-consistent-evaluation-evidence {
 				overflow: hidden;
 			}
+
+			.d2l-consistent-evaluation-no-evidence {
+				align-items: center;
+				display: flex;
+				flex-direction: column;
+				height: calc(100% - 2.7rem);
```
jeroach, line 48: nit: I moved the `2.7rem` to a CSS variable in my PR - feel free to update now, or we can take care of it when the merge happens.

components/left-panel/consistent-evaluation-left-panel.js
Scy-D2L: I believe yes, typically there should be a default on every case, but I don't know what it should be. If we run into the case where the submission type isn't on this list I believe something else must have gone critically wrong and none of the options would make sense. So... maybe not? Or maybe some string to tip off to the fact that there's an error here? Hmm.

jeroach: Yeah I wasn't exactly sure myself - I mean since we're building out the logic, it's a case that really shouldn't happen.

Scy-D2L: It seems to me that throwing an error makes the most sense, because anything else would be a misrepresentation of what we're looking at. I can `console.warn('unknown submission type: <type>')`

components/left-panel/consistent-evaluation-left-panel.js
Scy-D2L: Possible larger issue: That 2.7 rem exists to compensate for the back button up top (I think?), but this PR (and the design it's based on) doesn't include that bar. I think we do need it though, right?

jeroach: Yeah good call out. I think we can just do 100% here. Since for these cases, there is no list of submissions to go back to :)

PR #54, opened 2020-07-27
remove moment import from submissions page

zhangium:
Related to user story: https://rally1.rallydev.com/#/detail/userstory/408052509400?fdp=true

### DIFF
--- a/components/left-panel/consistent-evaluation-submissions-page.js
+++ b/components/left-panel/consistent-evaluation-submissions-page.js
@@ -4,7 +4,6 @@ import '@brightspace-ui/core/components/colors/colors.js';
 import './consistent-evaluation-submission-item.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { Classes } from 'd2l-hypermedia-constants';
-import moment from 'moment/src/moment';
 
 export class ConsistentEvaluationSubmissionsPage extends LitElement {
 	static get properties() {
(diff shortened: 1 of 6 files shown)


components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -44,6 +43,13 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 		this._submissionList = [];
 		this._token = undefined;
 		this._submissionEntities = [];
+		/* global moment:false */
+		moment.relativeTimeThreshold('s', 60);
+		moment.relativeTimeThreshold('m', 60);
+		moment.relativeTimeThreshold('h', 24);
+		moment.relativeTimeThreshold('d', 7);
+		moment.relativeTimeThreshold('w', 4);
+		moment.relativeTimeThreshold('M', 12);
```
AlexBedley, line 52: Are these thresholds from Dan? Just wanna make sure we're consistent here with the rest of the LMS

zhangium: Ah no they are not. I went off of my own assumptions here. Will double check with Dan.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -44,6 +43,11 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 		this._submissionList = [];
 		this._token = undefined;
 		this._submissionEntities = [];
+		/* global moment:false */
+		moment.relativeTimeThreshold('s', 60);
+		moment.relativeTimeThreshold('m', 60);
+		moment.relativeTimeThreshold('h', 24);
+		moment.relativeTimeThreshold('d', Number.MAX_SAFE_INTEGER);
```
zhangium, line 50: After speaking with Dan, the desired outcome is for everything to be counted in units of `days`, upwards of days.

PR #57, opened 2020-07-28
fix scrolling

jeroach:
Fix for issue from this thread: https://d2l.slack.com/archives/CSL64JP9U/p1595852000335200

scrollbars


jeroach:
> This approach looks good. However, did we look into why we need `overflow: hidden` in the first place? Can't we just make the evidence display `height: 100%` so no scrollbars are displayed?

The scrollbars for evidence are inside the iframe - CSS outside of that won't touch it

PR #59, opened 2020-07-30
Toast implementation

JamieTran:
Screen Shot 2020-07-30 at 4 15 49 PM
Screen Shot 2020-07-30 at 4 15 55 PM
Screen Shot 2020-07-30 at 4 15 43 PM
Screen Shot 2020-07-30 at 4 15 38 PM

Notes: 

1. Toast is covering up the footer (was told to leave this for now)
2. I think the toast will reappear when something is rendered to the screen
3. Toasts should be displayed _once the associated web request is completed_. To me, this means on receiving the status code from the web request but there is no (to my knowledge) a simple way to do this without an API refactor. Instead, I am checking if the entity returned is an error or not.


components/consistent-evaluation-page.js
```diff
@@ -52,6 +55,12 @@ export default class ConsistentEvaluationPage extends LitElement {
 			token: {
 				type: String
 			},
+			displayToast: {
```
AlexBedley, line 58: `displayToast` is internal to this component, so by convention it should start with an underscore

components/consistent-evaluation-page.js
```diff
@@ -209,29 +223,44 @@ export default class ConsistentEvaluationPage extends LitElement {
 	async _saveEvaluation() {
 		const entity = await this._controller.fetchEvaluationEntity(false);
 		this.evaluationEntity = await this._controller.save(entity);
+		if (!(this.evaluationEntity instanceof Error))  {this._toastMessage = this.localize('saved');}
+		this.toastMessage = this.localize(this._toastMessage);
```
AlexBedley, line 227: looks like this line can be deleted

components/consistent-evaluation-page.js
```diff
@@ -209,29 +223,44 @@ export default class ConsistentEvaluationPage extends LitElement {
 	async _saveEvaluation() {
 		const entity = await this._controller.fetchEvaluationEntity(false);
 		this.evaluationEntity = await this._controller.save(entity);
+		if (!(this.evaluationEntity instanceof Error))  {this._toastMessage = this.localize('saved');}
```
AlexBedley, line 226: nit: style-wise I'd prefer the if-block to be expanded, this makes the coding style consistent with the rest of the repository.

In general, we avoid commenting on style things because the linter should be enforcing them. But in this case the linter is not, I can look into adding it.

components/consistent-evaluation-page.js
```diff
@@ -209,29 +223,44 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+		this.toastMessage = this.localize(this._toastMessage);
 		this.evaluationState = this.evaluationEntity.properties.state;
 	}
 
 	async _updateEvaluation() {
 		const entity = await this._controller.fetchEvaluationEntity(false);
 		this.evaluationEntity = await this._controller.update(entity);
+		if (!(this.evaluationEntity instanceof Error))  {this._toastMessage = this.localize('updated');}
```
AlexBedley, line 234: I sorta mentioned this to you before, but the presence of an entity isn't necessarily enough to know the action succeeded. We don't have a very robust design around HM error states (usually every present action is a legal action)

But for this PR, I think this is fine.

components/consistent-evaluation-page.js
```diff
@@ -209,29 +223,44 @@ export default class ConsistentEvaluationPage extends LitElement {
...
 		this.evaluationState = this.evaluationEntity.properties.state;
 	}
 
 	async _publishEvaluation() {
 		const entity = await this._controller.fetchEvaluationEntity(false);
 		this.evaluationEntity = await this._controller.publish(entity);
 		this.evaluationState = this.evaluationEntity.properties.state;
+		if (!(this.evaluationEntity instanceof Error))  {this._toastMessage = this.localize('published');}
```
AlexBedley, line 242: What happens if we do get an `Error` back? Should we display an error state toast?

components/consistent-evaluation-page.js
```diff
@@ -209,29 +223,44 @@ export default class ConsistentEvaluationPage extends LitElement {
...
 	async _retractEvaluation() {
 		const entity = await this._controller.fetchEvaluationEntity(false);
 		this.evaluationEntity = await this._controller.retract(entity);
+		if (!(this.evaluationEntity instanceof Error))  {this._toastMessage = this.localize('retracted');}
 		this.evaluationState = this.evaluationEntity.properties.state;
 		this.submissionInfo.evaluationState = draftState;
 	}
 
+	_renderToast() {
+		this.displayToast = (this.displayToast === true) ? false : true; // idk why but this resets the toast timer
```
AlexBedley, line 255: I find the handling of whether the toast is displayed or not confusing. I would just expect it to be an internal boolean property like `_displayToast` that controls the `open` attribute on the d2l-alert-toast.

Something like:

```js
_renderToast() {
  return html`<d2l-alert-toast open=${this._displayToast}>${this._toastMessage}</d2l-alert-toast>`;
}
```

And then when you want to open it and set the message just do:

```js
this._displayToast = false; // reset in case the toast was already open and not manually closed
this._toastMessage = this.localize('langtermId');
this._displayToast = true; // open the toast message
```

components/consistent-evaluation-page.js
```diff
@@ -209,29 +223,44 @@ export default class ConsistentEvaluationPage extends LitElement {
...
+
+		if (this._toastMessage === '') return '';
```
AlexBedley, line 257: I would expect this case to be covered when `_displayToast` is `false`

components/consistent-evaluation-page.js
```diff
@@ -263,8 +292,9 @@ export default class ConsistentEvaluationPage extends LitElement {
 						@on-d2l-consistent-eval-feedback-edit=${this._transientSaveFeedback}
 						@on-d2l-consistent-eval-grade-changed=${this._transientSaveGrade}
 					></consistent-evaluation-right-panel>
-				</div>
+					</div>
```
AlexBedley, line 295: nit: the indent was correct before

components/consistent-evaluation-page.js
dbatiste: I _suspect_ that a `await this.requestUpdate()` call would be needed after the first update to set `this._displayToast = false`, otherwise by the time the life cycle runs it will just be true and `render` won't be called again if it was already `true`.  Alternatively, `_displayToast` should be kept in sync with the toast state, probably by listening to a close event, but it looks like we're not exposing such an event for toasts yet.

components/consistent-evaluation-page.js
JamieTran: An `Error` should display an error message but that has not been investigated yet and is related to a different story: US118570.

components/consistent-evaluation-page.js
JamieTran: Thanks for the feedback and I totally agree. I did run into the where the toast wouldn't render again due to it always being `true`. I'm testing `await this.requestUpdate()` and it seems that the toast timer does not reset, do you have any thoughts on how that works?

dbatiste: I'm not super familiar with it, but it's animation could be causing trouble.  I think it would be most appropriate if we add a close event of time kind to the component, then your private bool can be kept properly in sync.  I've created https://github.com/BrightspaceUI/core/issues/759 for this.  It should be fairly easy to add.

dbatiste: FYI: I've add a `d2l-alert-toast-close` to the toast component in `core@v1.66.0`.  So the preferred way to keep in sync would be:
```
<d2l-alert-toast open="${this._displayToast}" @d2l-alert-toast-close="${this._onToastClose}"  >...</d2l-alert-toast>
```
```
_onToastClose() {
    this._displayToast = false;
}
```

components/consistent-evaluation-page.js
```diff
@@ -209,29 +224,68 @@ export default class ConsistentEvaluationPage extends LitElement {
...
 
 	async _retractEvaluation() {
 		const entity = await this._controller.fetchEvaluationEntity(false);
 		this.evaluationEntity = await this._controller.retract(entity);
+		this._getToastMessage(retractActionName, this.evaluationEntity);
 		this.evaluationState = this.evaluationEntity.properties.state;
 		this.submissionInfo.evaluationState = draftState;
 	}
 
+	_getToastMessage(action, entity) {
```
AlexBedley, line 254: I have a few nitpicky design comments around this function. First I think the name should give a hint that it opens the toast. Secondly, I'd just take the langterm as a parameter and eliminate the action-name-middle-man completely. Lastly, this one is mostly my opinion, but I'd just make the function show the toast (rather than determine if the toast should be shown a la `instanceof Error`)

Like:

```js
if (!(this.evaluationEntity instanceof Error)) {
  this.showToast(this.localize('retracted'));
} else {
  this.showToast(this.localize('retractError'));
}
```

JamieTran: I totally agree with you on the first two comments and will make these changes, but can you explain to me the reasoning for the last one? From my point of view, it cuts down on code repetition (I guess if we change it to `_showToast`, it's breaking the single responsibility principle?).

AlexBedley: Agreed, its a tradeoff between code repetition and simpler functions. I'm ok with the explicit repetition because I think its makes the code easier to read (the original code made me ask "why does the toast function need the entity?"). Additionally, if we ever needed to use toasts for other purposes on this page we can simply call `showToast` and not worry about the 2nd parameter (which may or may not be relevant for the other use-case) which you already called out re: single responsibility

PR #60, opened 2020-08-04
Add text truncation for attachment item filenames

zhangium:
These changes should be able to take effect after this on core gets merged in: https://github.com/BrightspaceUI/core/pull/752
**Screenshot:**
image

### DIFF
--- a/components/left-panel/consistent-evaluation-submission-item.js
+++ b/components/left-panel/consistent-evaluation-submission-item.js
@@ -57,6 +57,10 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			submissionType: {
 				attribute: 'submission-type',
 				type: String
+			},
+			tooltips : {
+				attribute: false,
+				type: Object
 			}
 		};
 	}
@@ -151,6 +155,7 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		this._date = undefined;
 		this._attachments = [];
 		this._comment = '';
+		this._tooltips = [];
 	}
 
 	get submissionEntity() {
@@ -164,6 +169,27 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			this._initializeSubmissionProperties();
 			this.requestUpdate();
 		}
+	}
+
+	get tooltips() {
+		return this._tooltips;
+	}
+
+	set tooltips(newVal) {
+		const oldVal = this._tooltips;
+		if (oldVal !== newVal) {
+			this._tooltips = newVal;
+			this.requestUpdate();
+		}
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('load', this.insertTooltipsAfterUpdateComplete());
+	}
+	disconnectedCallback() {
+		window.removeEventListener('load', this.insertTooltipsAfterUpdateComplete());
+		super.disconnectedCallback();
 	}
 
 	_initializeSubmissionProperties() {
@@ -320,10 +346,12 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 
 	_renderAttachments() {
 		// href placeholder on list-item
-		return html`${this._attachments.map((file) => html`
-			<d2l-list-item @mouseover=${
-	// eslint-disable-next-line lit/no-template-arrow
-	() => this.insertTooltipsAfterUpdateComplete()}>
+		const template = [];
+		for (let i = 0; i < this._attachments.length; i++) {
+			const file = this._attachments[i];
+			const hasTooltip = this.tooltips[i];
+			template.push(html`
+			<d2l-list-item>
 			<div slot="illustration" class="d2l-submission-attachment-icon-container">
 				<d2l-icon class="d2l-submission-attachment-icon-container-inner"
 					icon="tier2:${this._getIcon(file.properties.name)}"
@@ -344,7 +372,43 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			</d2l-list-item-content>
 			${this._addMenuOptions(file.properties.read, file.properties.flagged, file.properties.href, file.properties.name)}
 			</d2l-list-item>
-			`)}`;
+			${this._renderTooltip(hasTooltip, file.properties.name)}
+			`);
+		}
+		return html`${template}`;
+	}
+
+	_renderTooltip(id, text) {
+		if (id) {
+			return html`<d2l-tooltip for="${id}" class="d2l-truncated-filename-tooltips"
+				offset="0" align="start">${text}</d2l-tooltip>`;
+		}
+	}
+
+	isClamped(e) {
+		return e.clientHeight < e.scrollHeight;
+	}
+
+	async insertTooltipsAfterUpdateComplete() {
+		await this.updateComplete;
+		this.insertFilenameTooltips();
+	}
+
+	insertFilenameTooltips() {
+		const items = this.shadowRoot.querySelectorAll('.truncate');
+		const keys = [];
+		items.forEach(element => {
+			if (this.isClamped(element)) {
+				const uniqueId = getUniqueId();
+				keys.push(uniqueId);
+				// attach to the parent d2l-list-item component for visibility
+				const target = element.closest('d2l-list-item');
+				target.id = uniqueId;
+			} else {
+				keys.push(false);
+			}
+		});
+		this.tooltips = keys;
 	}
 
 	_addMenuOptions(read, flagged, downloadHref, name) {
@@ -420,26 +484,8 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		}
 	}
 
-	isClamped(e) {
-		return e.clientHeight < e.scrollHeight;
-	}
-
-	async insertTooltipsAfterUpdateComplete() {
-		this.updateComplete.then(this.insertFilenameTooltips());
-	}
-
-	insertFilenameTooltips() {
-		const items = this.shadowRoot.querySelectorAll('.truncate');
-		items.forEach(element => {
-			if (this.isClamped(element)) {
-				const uniqueId = getUniqueId();
-				// attach to the d2l-list-item component for visibility
-				element.parentElement.parentElement.id = uniqueId;
-				element.parentElement.parentElement.insertAdjacentHTML('afterend',
-					`<d2l-tooltip for="${uniqueId}" class="d2l-truncated-filename-tooltips"
-					offset="0" align="start">${element.innerText}</d2l-tooltip>`);
-			}
-		});
+	async _getUpdateComplete() {
+		await super._getUpdateComplete();
 	}
 }
 customElements.define('d2l-consistent-evaluation-submission-item', ConsistentEvaluationSubmissionItem);
--- a/components/left-panel/consistent-evaluation-submission-item.js
+++ b/components/left-panel/consistent-evaluation-submission-item.js
@@ -9,10 +9,12 @@ import '@brightspace-ui/core/components/menu/menu.js';
 import '@brightspace-ui/core/components/menu/menu-item-link.js';
 import '@brightspace-ui/core/components/more-less/more-less.js';
 import '@brightspace-ui/core/components/status-indicator/status-indicator.js';
+import '@brightspace-ui/core/components/tooltip/tooltip.js';
 import { bodySmallStyles, heading3Styles, labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { fileSubmission, textSubmission } from '../controllers/constants';
 import { formatDate, formatTime } from '@brightspace-ui/intl/lib/dateTime.js';
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
 import { loadLocalizationResources } from '../locale.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
@@ -316,7 +318,7 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	_renderAttachments() {
 		// href placeholder on list-item
 		return html`${this._attachments.map((file) => html`
-			<d2l-list-item>
+			<d2l-list-item @mouseover=${this.handle()}>
 			<div slot="illustration" class="d2l-submission-attachment-icon-container">
 				<d2l-icon class="d2l-submission-attachment-icon-container-inner"
 					icon="tier2:${this._getIcon(file.properties.name)}"
@@ -327,7 +329,7 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			@click="${
 	// eslint-disable-next-line lit/no-template-arrow
 	() => this._dispatchRenderEvidenceEvent(file.properties.fileViewer)}">
-				<div class="truncate">${this._getFileTitle(file.properties.name)}</div>
+				<div class="truncate" aria-label="heading">${this._getFileTitle(file.properties.name)}</div>
 				<div slot="supporting-info">
 					${this._renderFlaggedStatus(file.properties.flagged)}
 					${this._getFileExtension(file.properties.name)}
@@ -395,7 +397,6 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		${this._renderAttachments()}
 		</d2l-list>
 		${this._renderComment()}
-
 		`;
 	}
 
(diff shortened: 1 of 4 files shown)


zhangium:
Tooltips notes:
- they're attached to a "superparent" d2l-list-item because if attached within a div in d2l-list-item-content, it gets cut off
- they get attached to the list item if a filename is too long (> 3 lines) and gets clamped/truncated


image

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -123,6 +125,18 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
 			margin-right: 0;
 		}
+		.truncate {
+			text-overflow: ellipsis;
+			white-space: break-spaces;
+			overflow: hidden;
+			overflow-wrap: break-word;
+			display: -webkit-box;
+			-webkit-line-clamp: 3;
+			-webkit-box-orient: vertical;
```
AlexBedley, line 135: do these work on non-webkit browsers?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -307,7 +321,9 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	_renderAttachments() {
 		// href placeholder on list-item
 		return html`${this._attachments.map((file) => html`
-			<d2l-list-item>
+			<d2l-list-item @mouseover=${
```
AlexBedley, line 324: what's the performance of this? seems like we're adding lots of things to the dom every mouseover. What happens if you mouseover more than once?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -307,7 +321,9 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	// eslint-disable-next-line lit/no-template-arrow
+	() => this.insertTooltipsAfterUpdateComplete()}>
```
AlexBedley, line 326: can we avoid the arrow function and just set mouseover to `this.insertTooltipsAfterUpdateComplete`?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -404,5 +419,27 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			return html`${this._renderTextSubmission()}`;
 		}
 	}
+
+	isClamped(e) {
+		return e.clientHeight < e.scrollHeight;
+	}
+
+	async insertTooltipsAfterUpdateComplete() {
+		this.updateComplete.then(this.insertFilenameTooltips());
```
AlexBedley, line 428: nit: the `async` way to do this is:

```js
await this.updateComplete;
this.insertFilenameTooltips();
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -404,5 +419,27 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	}
+
+	insertFilenameTooltips() {
+		const items = this.shadowRoot.querySelectorAll('.truncate');
+		items.forEach(element => {
+			if (this.isClamped(element)) {
+				const uniqueId = getUniqueId();
+				// attach to the d2l-list-item component for visibility
+				element.parentElement.parentElement.id = uniqueId;
```
AlexBedley, line 437: I wonder if there's a more robust way to get the `d2l-list-item` (you might not expect moving the DOM around to break tooltips). What if we used something like [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -404,5 +419,27 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+				element.parentElement.parentElement.insertAdjacentHTML('afterend',
```
AlexBedley, line 438: I wonder what the performance of this is vs doing it the `document.createElement` and `insertBefore` way

components/left-panel/consistent-evaluation-submission-item.js
zhangium: you're right, this keeps adding tooltips each time whereas we just want to add it once in the beginning. I didn't notice any lag when I tried to mouseover multiple times so that was overlooked

components/left-panel/consistent-evaluation-submission-item.js
zhangium: no, I don't think they'd work for non-webkit browsers. is that something we need to support? just tested in IE and this definitely breaks along with a lot of other styling. I've found webkit styling in existing Brightspace code as well.

components/left-panel/consistent-evaluation-submission-item.js
zhangium: not too sure about performance. `insertAdjacentHTML` seems simpler as I can set all the attributes and the innertext in one go.

components/left-panel/consistent-evaluation-submission-item.js
AlexBedley: it's ok if it doesn't work in IE, we don't support it anymore. but I think we're still supporting legacy edge which might cause problems

components/left-panel/consistent-evaluation-submission-item.js
dbatiste: Note: this is modifying the DOM outside of `render` which should not be done.  `UI = f(data/properties)`, and inversion should be true also.  This makes the UI state very predictable, and it also allows Lit to batch/schedule updates.  To accomplish that, we'd set some property here, which would cause render to run again, where DOM changes can be reflected.

components/left-panel/consistent-evaluation-submission-item.js
dbatiste: The line-clamp is sort of weird - the other browsers (excluding IE11) actually implemented this features with the `-webkit` vendor prefix.  Eventually it's supposed to be replaced by actual `line-clamp` that doesn't have the issues that `-webkit-line-clamp` has.  https://caniuse.com/#search=line-clamp

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -17,6 +17,7 @@ import { formatDate, formatTime } from '@brightspace-ui/intl/lib/dateTime.js';
 import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
 import { loadLocalizationResources } from '../locale.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
+import { default as ResizeObserver } from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
```
dbatiste, line 20: I _think_ this import can just be:
```
import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -57,6 +58,10 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			submissionType: {
 				attribute: 'submission-type',
 				type: String
+			},
+			tooltips : {
```
dbatiste, line 62: I think this should be:
```
_tooltips: { type: Array }
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -166,6 +172,36 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		}
 	}
 
+	get tooltips() {
```
dbatiste, line 175: I don't think this custom getter/setter is needed.  It doesn't appear to be doing anything beyond what the existing default getter/setter will do.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -320,10 +356,12 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 
 	_renderAttachments() {
 		// href placeholder on list-item
-		return html`${this._attachments.map((file) => html`
-			<d2l-list-item @mouseover=${
-	// eslint-disable-next-line lit/no-template-arrow
-	() => this.insertTooltipsAfterUpdateComplete()}>
+		const template = [];
+		for (let i = 0; i < this._attachments.length; i++) {
```
dbatiste, line 360: The `for` loop is fine, but if you prefer the `map`, it can provide the index.
```
this._attachments.map((file, index) => html`...`);
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -320,10 +356,12 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+			const file = this._attachments[i];
+			const hasTooltip = this.tooltips[i];
+			template.push(html`
+			<d2l-list-item>
 			<div slot="illustration" class="d2l-submission-attachment-icon-container">
```
dbatiste, line 365: Nit: we should fix up the indentation here to make this easier to read.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -344,7 +382,38 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	}
+
+	_renderTooltip(id, text) {
+		if (id) {
+			return html`<d2l-tooltip for="${id}" class="d2l-truncated-filename-tooltips"
+				offset="0" align="start">${text}</d2l-tooltip>`;
+		}
+	}
+
+	_isClamped(e) {
```
dbatiste, line 398: Nit: `e` is usually exclusively used as the variable for events, so I'd probably name this more descriptively.  ex. `elem` or `fileNameElem`, the latter probably making it easier for the reader to understand what specifically is being measured.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -344,7 +382,38 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			</d2l-list-item-content>
 			${this._addMenuOptions(file.properties.read, file.properties.flagged, file.properties.href, file.properties.name)}
 			</d2l-list-item>
-			`)}`;
+			${this._renderTooltip(hasTooltip, file.properties.name)}
```
dbatiste, line 385: This appears to render the tooltip outside of `d2l-list-item`, but I think we want it inside the list-item.  It would be a little dicey rendering in between the list-items because the `d2l-list` expects its children to be list-items.  Besides that, I think the use of `closest` (further below) relies on the tooltip being inside of a list-item, ya?

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -344,7 +382,38 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+			return html`<d2l-tooltip for="${id}" class="d2l-truncated-filename-tooltips"
+				offset="0" align="start">${text}</d2l-tooltip>`;
+		}
+	}
+
+	_isClamped(e) {
+		return e.clientHeight < e.scrollHeight;
+	}
+
+	_insertFilenameTooltips() {
```
dbatiste, line 402: I'd consider renaming this method to better reflect what it's doing.  Perhaps `_updateFileNameTooltips`.  The "insert" makes it sound like it's still inserting into the DOM, but maybe it's just me.  :)

Aside from that, in this file we are using variables `key`, `hasTooltip`, and `id` all to refer to the same thing, which makes things harder to read.  We should be consistent in the naming.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -320,10 +356,12 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
 	_renderAttachments() {
 		// href placeholder on list-item
-		return html`${this._attachments.map((file) => html`
-			<d2l-list-item @mouseover=${
-	// eslint-disable-next-line lit/no-template-arrow
-	() => this.insertTooltipsAfterUpdateComplete()}>
+		const template = [];
+		for (let i = 0; i < this._attachments.length; i++) {
+			const file = this._attachments[i];
+			const hasTooltip = this.tooltips[i];
```
dbatiste, line 362: The `hasTooltip` variable name is confusing here, since it is named like a boolean, but in the case where it will have a tooltip, it's actually the id of the tooltip target.  I might consider renaming this to be `tooltipTargetId` to avoid reader confusion.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -344,7 +382,38 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			</d2l-list-item-content>
 			${this._addMenuOptions(file.properties.read, file.properties.flagged, file.properties.href, file.properties.name)}
 			</d2l-list-item>
-			`)}`;
+			${this._renderTooltip(hasTooltip, file.properties.name)}
+			`);
+		}
+		return html`${template}`;
```
dbatiste, line 388: `template` is an array of templates - I don't think you need to wrap this in another template, as it would just create an unnecessary binding (see https://lit-html.polymer-project.org/guide/template-reference#looping-statements)  So I think you can just go:
```
return template;
```
Aside, I might consider renaming to be plural to avoid reader confusion.

components/left-panel/consistent-evaluation-submission-item.js
zhangium: It looked like the tooltip had to exist as a next sibling of the target to me, which is why I put the tooltip outside of the list item. https://github.com/BrightspaceUI/core/tree/master/components/tooltip

dbatiste: Oh I see.  I'm not sure if that's the required placement.  @AllanKerr ... does the tooltip need to exist beside the element it is targeting, or is it valid to put it inside?

Annie, if putting beside, we need to be careful of a11y and how this announces.  Also, if the items could ever be removed or reordered, it would be better for the tooltip to be within.

zhangium: I think the attachments list will be purely presentational with no reordering or insertion/deletion of items capability on the page. Additionally, I think some styling on the list item is covering up part of the tooltip if the component is placed inside. That might require more custom styling or a change to the core component. I do think it's weird WRT list and list items, but it looks like accessibility nodes don't include tooltip isn't a list item. How do you think I should proceed?

AllanKerr: There are two options for tooltips. Usually its encourages to use `for` and link explicitly via the ID but if you don't give a tooltip a `for` attribute then it will use its parent element as its target.

dbatiste: So sounds like placing inside should be doable but there are some list-styles that are causing some issue.  Maybe related to overflow?  @m6jones do you have any thoughts on this?  cc'ing you since i know you are looking at overflow for list items currently, and this is a potential use-case.

dbatiste: @zhangium : feel free to put the tooltip beside for the time being if necessary.  I am away on vacation next week so I didn't want you waiting in limbo for a decision, but also we want to make sure this use case works the way we want and is accessible, so we have clear way to do this for the next person.  So will be interested in the work that Maya is doing with overflow.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -327,8 +365,35 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	_updateFileNameTooltips() {
+		const filenames = this.shadowRoot.querySelectorAll('.truncate');
+		const uniqueIds = [];
+		filenames.forEach(element => {
+			if (this._isClamped(element)) {
+				const uniqueId = getUniqueId();
+				uniqueIds.push(uniqueId);
+				element.id = uniqueId;
+			} else {
+				uniqueIds.push(false);
```
AlexBedley, line 393: I do worry a bit about relying on the ordering of `querySelectorAll` and `render` being the same (could we tie them together using something on the `file`?) but it's probably fine

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -327,8 +365,35 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	}
+
+	_isClamped(element) {
+		return element.clientHeight < element.scrollHeight;
+	}
+
+	_updateFileNameTooltips() {
+		const filenames = this.shadowRoot.querySelectorAll('.truncate');
+		const uniqueIds = [];
+		filenames.forEach(element => {
```
AlexBedley, line 387: nit: it would be more idiomatic to just use `map` like:

```js
this._tooltips = filenames.map(element => {
  if (clamped) {
    return uniqueId;
  }
  return false;
});
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -152,6 +172,25 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	disconnectedCallback() {
+		window.removeEventListener('load', this._updateFileNameTooltips());
+		super.disconnectedCallback();
+	}
+
+	firstUpdated() {
+		super.firstUpdated();
+		const filenames = this.shadowRoot.querySelectorAll('.truncate');
+		for (const filename of filenames) {
+			this._resizeObserver.observe(filename);
```
AlexBedley, line 190: do we need to un-observe? or does the library handle that

zhangium: I could add that in `disconnectedCallback`

zhangium:
Closing and creating a new PR regarding updated input from Dan. We don't want to use the d2l-tooltip component, rather we want to use the built-in HTML `title` attribute to display tooltips

PR #62, opened 2020-08-05
Multiple attachments add

vix633:
Remove is still only doing one file at a time, getting this out for testing to start happening.


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -63,9 +66,26 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 			detail: feedback
 		}));
 	}
-	async saveAttachment() {
-		this.shadowRoot.querySelector('d2l-activity-attachments-editor').save();
+	async saveAttachment(e) {
+		const files = e.detail.files;
+		for (let i = 0; files.length > i; i++) {
```
AlexBedley, line 71: nit: `files.forEach` would be more idiomatic, ngl I had to reread `files.length > i` like 5 times before I figured out what it was doing

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -63,9 +66,26 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
...
+			const fileSystemType = files[i].m_fileSystemType;
+			const fileId = files[i].m_id;
+
+			const sirenEntity = await window.D2L.Siren.EntityStore.get(this.href, this.token);
+			const link = sirenEntity.getLinkByRel('https://activities.api.brightspace.com/rels/attachments');
```
AlexBedley, line 76: nit: the rel should be a constant somewhere. You could add it to https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/hypermedia-constants.js#L47

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -63,9 +66,26 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
...
+			const entitytemp = await window.D2L.Siren.EntityStore.fetch(link, this.token);
+
+			const entity = new AttachmentCollectionEntity(entitytemp.entity, this.token, { remove: () => { } });
```
AlexBedley, line 79: Looks like the cleanup is safe, you don't need the 3rd parameter

https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/f357e6ff7b51ccd7153d3d1794399683312ccb95/src/es6/Entity.js#L46

components/right-panel/consistent-evaluation-feedback-presentational.js
vix633: I'll follow up with this

PR #63, opened 2020-08-07
- Adding support for grade item buttons

jmurray-d2l:
- Looking for feedback, tests will be updated once we're comfortable with the solution


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -108,4 +108,32 @@ export class ConsistentEvaluationHrefController {
 			submissionType
 		};
 	}
+
+	async getGradeItemInfo() {
```
AlexBedley, line 112: This function was written to fail silently it seems. What's the impact if any of the 3 properties are missing?

jmurray-d2l: Yeah, fail silently since its valid for the grade item link to not be present - I've tried to make sure I handle the absence of any of the properties in the downstream components

zhangium: I think you'll still need some of the changes from the now closed PR in grade-result to handle the absences

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -109,5 +123,80 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
 			</d2l-consistent-evaluation-right-panel-block>
 		`;
 	}
+
+	_openGradeEvaluationDialog() {
+
+		const dialogUrl = this.gradeItemInfo && this.gradeItemInfo.evaluationUrl
```
zhangium, line 129: nit: missing semicolon? same for below

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -109,5 +123,80 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
 			</d2l-consistent-evaluation-right-panel-block>
 		`;
 	}
+
+	_openGradeEvaluationDialog() {
```
zhangium, line 127: is this being called anywhere? I see `_openGradeStatisticsDialog` but not this.

locales/en.js
```diff
@@ -27,6 +30,8 @@ export const val = {
 	recordAudio: 'Record Audio',
 	recordVideo: 'Record Video',
 	saveDraft: 'Save Draft',
+	saveBtn: 'Save',
+	statistics: 'Class and user statistics',
```
zhangium, line 34: nit: everything else looks like it's title case so just wondering if Dan wanted that?

jmurray-d2l: I think his is intended, I took it right from the design spec, but I can confirm with him

components/right-panel/consistent-evaluation-grade-result.js
jmurray-d2l: It should be!  I'll fix it

karma.sauce.conf.js
```diff
@@ -47,6 +47,7 @@ module.exports = config => {
 			customLaunchers: customLaunchers,
 			browsers: Object.keys(customLaunchers),
 			reporters: ['dots', 'saucelabs'],
+			browserNoActivityTimeout: 100000,
```
jmurray-d2l, line 50: Any thoughts on this change?  This fixes the test failures, but not sure if its just a symptom of another problem with the test?

zhangium: I'm not opposed to upping the timeout. That Safari disconnected failure has happens every once in a while and I don't think we've been able to figure out why. Usually a re-run of the test will pass.

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -82,25 +94,27 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
...
+				?includeReportsButton=${this.gradeItemInfo && this.gradeItemInfo.statsUrl}
 
 				?isGradeAutoCompleted=${this._isGradeAutoCompleted}
 				?isManualOverrideActive=${this._manuallyOverriddenGrade !== undefined}
 
 				?readOnly=${this.readOnly}
 				?hideTitle=${this.hideTitle}
 
+				@d2l-grade-result-reports-button-click=${this._openGradeStatisticsDialog}
+				@d2l-grade-result-grade-button-click=${this._openGradeEvaluationDialog}
```
zhangium, line 117: maybe match the event name to the handle method name or vice versa

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -109,5 +123,80 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
 			</d2l-consistent-evaluation-right-panel-block>
 		`;
 	}
+
+	_openGradeEvaluationDialog() {
+
+		const dialogUrl = this.gradeItemInfo && this.gradeItemInfo.evaluationUrl;
+
+		if (!dialogUrl) {
+			return;
```
AlexBedley, line 132: We might consider logging here

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -109,5 +123,80 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
...
+		];
+
+		D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
+			/*               opener: */ document.body,
+			/*             location: */ location,
+			/*          srcCallback: */ 'SrcCallback',
+			/*       resizeCallback: */ '',
+			/*      responseDataKey: */ 'result',
+			/*                width: */ 1920,
+			/*               height: */ 1080,
```
AlexBedley, line 160: Do we have plans to accommodate different screen sizes at some point?

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -109,5 +123,80 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
...
+
+	_openGradeEvaluationDialog() {
+
+		const dialogUrl = this.gradeItemInfo && this.gradeItemInfo.evaluationUrl;
+
+		if (!dialogUrl) {
+			return;
+		}
+
+		const location = new D2L.LP.Web.Http.UrlLocation(dialogUrl);
```
AlexBedley, line 135: What happens if this global (and the others) aren't present? I just want to make sure the app doesn't crash.

In the past we've handled missing LMS globals gracefully, but it was also more common to run outside of the LMS (I think with the onset of unbundled BSI it's more common to test with a real LMS). I'd be happy with a one-time check during app load that warns to the console that globals are missing and some safety here so we don't blow up the app.

jmurray-d2l: Good question,  I'll find a way to test this

components/right-panel/consistent-evaluation-grade-result.js
jmurray-d2l: I didn't have plans of making this configurable, but we might have to for responsiveness

PR #65, opened 2020-08-10
Truncate filenames over 3 lines and add tooltip if truncated

zhangium:
Truncates filenames with ellipses after 3 lines:

image

After load, a `title` attribute is added if text clamping is applied. On hover, this attribute makes a tooltip show up with the full filename text:

image


zhangium:
This PR is branched off from https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/60 but that one had used `d2l-tooltip` which was not requested afterall. I'm leaving that PR there as closed for transparency and for future reference if we need to do anything with tooltips like that.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -152,6 +162,29 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+		const filenames = this.shadowRoot.querySelectorAll('.truncate');
+		for (const filename of filenames) {
+			this._resizeObserver.unobserve(filename);
+		}
+		super.disconnectedCallback();
+	}
+
+	firstUpdated() {
+		super.firstUpdated();
+		const filenames = this.shadowRoot.querySelectorAll('.truncate');
```
jeroach, line 182: Why is this in `firstUpdated` instead of `connectedCallback`?

zhangium: I think because `connectedCallback` gets called multiple times and we don't want to observe multiple times, whereas `firstUpdated` happens once at the beginning

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -152,6 +162,29 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		}
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('load', this._updateFilenameTooltips());
```
AlexBedley, line 167: I don't think you meant to invoke `_updateFilenameTooltips`:

```js
window.addEventListener('load', this._updateFilenameTooltips);
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -152,6 +162,29 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+		this._resizeObserver = new ResizeObserver(() => this._updateFilenameTooltips());
```
AlexBedley, line 168: nit: there might be some `this` weirdness but you could probably eliminate the outer lambda:

```js
this._resizeObserver = new ResizeObserver(this._updateFilenameTooltips);
```

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -152,6 +162,29 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	}
+
+	disconnectedCallback() {
+		window.removeEventListener('load', this._updateFilenameTooltips());
```
AlexBedley, line 172: Similar to above, I assume you didn't intend to invoke `_updateFilenameTooltips`

PR #68, opened 2020-08-14
US119141: Use new property provided by the API to determine file type

zhangium:
- use the getFileIconTypeFromExtension that was added to core/icons
- use the extension property now returned by the API for file attachment
- re-add in the text submission view that somehow got erased (probably because of me not merging master to a prev branch frequently enough 😿 )
Note: no UI changes!

### DIFF
--- a/components/left-panel/consistent-evaluation-submission-item.js
+++ b/components/left-panel/consistent-evaluation-submission-item.js
@@ -220,10 +220,34 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		return Math.max(fileSizeBytes, 0.1).toFixed(1) + unit;
 	}
 
-	_dispatchRenderEvidenceEvent(url) {
-		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence', {
+	_dispatchRenderEvidence(extension, fileViewer) {
+		if (extension === 'txt') {
+			this._dispatchRenderEvidenceTextEvent();
+		}
+		else if (fileViewer !== undefined) {
+			this._dispatchRenderEvidenceFileEvent(fileViewer);
+		}
+	}
+
+	_dispatchRenderEvidenceFileEvent(url) {
+		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence-file', {
 			detail: {
 				url: url
+			},
+			composed: true
+		});
+		this.dispatchEvent(event);
+	}
+
+	_dispatchRenderEvidenceTextEvent() {
+		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence-text', {
+			detail: {
+				textSubmissionEvidence: {
+					title: `${this.localize('textSubmission')} ${this.displayNumber}`,
+					date: this._formatDateTime(),
+					downloadUrl: this._attachments[0].properties.href,
+					content: this._comment
+				}
 			},
 			composed: true
 		});
@@ -323,14 +347,11 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	}
 
 	_renderAttachments() {
-		// href placeholder on list-item
 		return html`${this._attachments.map((file) => {
-			const read = file.properties.read;
-			const flagged = file.properties.flagged;
-			const href = file.properties.href;
-			const name = file.properties.name;
-			const extension = file.properties.extension;
-			const size = file.properties.size;
+			const read = file.properties.read, flagged = file.properties.flagged,
+				href = file.properties.href, name = file.properties.name,
+				extension = file.properties.extension, size = file.properties.size,
+				fileViewer = file.properties.fileViewer;
 			return html`
 			<d2l-list-item>
 				<div slot="illustration" class="d2l-submission-attachment-icon-container">
@@ -342,11 +363,11 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 				<d2l-list-item-content
 				@click="${
 	// eslint-disable-next-line lit/no-template-arrow
-	() => this._dispatchRenderEvidenceFileEvent(file.properties.fileViewer)}">
+	() => this._dispatchRenderEvidence(extension, fileViewer)}">
 					<div class="truncate" aria-label="heading">${this._getFileTitle(name)}</div>
 					<div slot="supporting-info">
 						${this._renderFlaggedStatus(flagged)}
-						${extension}
+						${extension.toUpperCase()}
 						<d2l-icon class="d2l-separator-icon" aria-hidden="true" icon="tier1:dot"></d2l-icon>
 						${this._getReadableFileSizeString(size)}
 					</div>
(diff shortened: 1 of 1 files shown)


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -236,8 +220,17 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		return Math.max(fileSizeBytes, 0.1).toFixed(1) + unit;
 	}
 
-	_dispatchRenderEvidenceEvent(url) {
-		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence', {
+	_dispatchRenderEvidence(extension, fileViewer) {
+		if (extension === 'txt') {
+			this._dispatchRenderEvidenceTextEvent();
+		}
+		else if (fileViewer !== undefined) {
```
AlexBedley, line 227: nit: Not sure if we're just specifically worried about `undefined`, but if we also want to protect against `null` etc. we could just do `else if (fileViewer)`

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -236,8 +220,17 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		return Math.max(fileSizeBytes, 0.1).toFixed(1) + unit;
 	}
 
-	_dispatchRenderEvidenceEvent(url) {
-		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence', {
+	_dispatchRenderEvidence(extension, fileViewer) {
+		if (extension === 'txt') {
```
AlexBedley, line 224: nit: this string does appear twice in this file, might be worth a constant

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -246,6 +239,21 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		this.dispatchEvent(event);
 	}
 
+	_dispatchRenderEvidenceTextEvent() {
+		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence-text', {
+			detail: {
+				textSubmissionEvidence: {
+					title: `${this.localize('textSubmission')} ${this.displayNumber}`,
+					date: this._formatDateTime(),
+					downloadUrl: this._attachments[0].properties.href,
```
AlexBedley, line 248: Is attachment order guaranteed? It might be safer to find the attachment via `class` or `rel`

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -339,29 +347,34 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
-	// eslint-disable-next-line lit/no-template-arrow
-	() => this._dispatchRenderEvidenceEvent(file.properties.fileViewer)}">
-				<div class="truncate" aria-label="heading">${this._getFileTitle(file.properties.name)}</div>
-				<div slot="supporting-info">
-					${this._renderFlaggedStatus(file.properties.flagged)}
-					${this._getFileExtension(file.properties.name)}
-					<d2l-icon class="d2l-separator-icon" aria-hidden="true" icon="tier1:dot"></d2l-icon>
-					${this._getReadableFileSizeString(file.properties.size)}
+		return html`${this._attachments.map((file) => {
+			const read = file.properties.read, flagged = file.properties.flagged,
```
AlexBedley, line 351: nit: you can use the [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) for this like:

```js
const {read, flagged, href, name, extension, size, fileViewer} = file.properties;
```

components/left-panel/consistent-evaluation-submission-item.js
zhangium: for text submissions it's guaranteed that there's only one single attachment (an html file)

zhangium: so I _think_ this is okay now if I `getSubEntitiesByClass("attachment")` earlier when the list is initiated @AlexBedley

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -379,27 +390,30 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+	_addMenuOptions(read, flagged, downloadHref, extension) {
 		// Placeholder for menu presentational
 		const oppositeReadState = read ? this.localize('markUnread') : this.localize('markRead');
 		const oppositeFlagState = flagged ? this.localize('unflag') : this.localize('flag');
-		const fileType = this._getFileExtension(name);
 		return html`<div slot="actions" style="z-index: inherit;">
 			<d2l-dropdown-more text="More Options">
 			<d2l-dropdown-menu id="dropdown" boundary="{&quot;right&quot;:10}">
 				<d2l-menu>
+				${this.submissionType === textSubmission ? html`
```
jeroach, line 401: Thanks for adding this back :)

zhangium: yep! sorry for getting rid of it in the first place :octocat:

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -379,27 +390,30 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
-	_addMenuOptions(read, flagged, downloadHref, name) {
+	_addMenuOptions(read, flagged, downloadHref, extension) {
 		// Placeholder for menu presentational
 		const oppositeReadState = read ? this.localize('markUnread') : this.localize('markRead');
 		const oppositeFlagState = flagged ? this.localize('unflag') : this.localize('flag');
-		const fileType = this._getFileExtension(name);
 		return html`<div slot="actions" style="z-index: inherit;">
 			<d2l-dropdown-more text="More Options">
 			<d2l-dropdown-menu id="dropdown" boundary="{&quot;right&quot;:10}">
 				<d2l-menu>
```
dbatiste, line 400: Just noticed this a11y issue: the root `d2l-menu` itself should have a `label` ([example usage](https://github.com/BrightspaceUI/core/tree/master/components/dropdown#d2l-dropdown-menu)).  This will get announced along with the menu semantics to screen readers.  Many times this is the same as the text for the opener, in the case where the menu is in a dropdown, but it doesn't necessarily have to be the same.  For nested menus (menus inside of menus), the label is automatically assigned based on the menu-item that opens the nested menu.

zhangium: thanks for pointing that out, I missed it when adding the menu.

components/left-panel/consistent-evaluation-submission-item.js
AlexBedley: seems a bit safer - it's probably fine. I get alarms bells in my head when I read an unprotected array lookup (What if _attachments is empty? what if someone modifies _attachments after initialization? what if someone changes the init method and not this function?) these are the things that keep me up at night 🤣

PR #69, opened 2020-08-18
Add StyleLint to make styling more consistent

JamieTran:
Added cool 3rd-party linting tool for styles embedded in HTML, markdown, JS, and most conveniently, template literals.
Rules are imported from @brightspace-ui/stylelint-config
This will hopefully keep consistent linting of styles in our components.

For more information view this blog post: https://blog.d2l.dev/2020/08/04/style-linting/

Example output (npm run lint):
image

### DIFF
--- a/components/left-panel/consistent-evaluation-evidence-text.js
+++ b/components/left-panel/consistent-evaluation-evidence-text.js
@@ -21,18 +21,18 @@ export class ConsistentEvaluationEvidenceText extends LocalizeMixin(LitElement) 
 
 	static get styles() {
 		return [bodyStandardStyles, heading2Styles, css`
-			.consistent-eval-text-evidence {
+			.d2l-consistent-eval-text-evidence {
+				padding-bottom: 2rem;
 				padding-left: 0.8rem;
 				padding-right: 0.8rem;
-				padding-bottom: 2rem;
 			}
-			.consistent-eval-text-evidence-options {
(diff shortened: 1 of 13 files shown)


components/left-panel/consistent-evaluation-evidence-text.js
```diff
@@ -21,18 +21,18 @@ export class ConsistentEvaluationEvidenceText extends LocalizeMixin(LitElement)
 
 	static get styles() {
 		return [bodyStandardStyles, heading2Styles, css`
-			.consistent-eval-text-evidence {
+			.d2l-consistent-eval-text-evidence {
```
jeroach, line 24: Don't we need to change the usages of the styles in this file as well?

JamieTran: Yes, you're right. Good catch 😅

JamieTran:
> looks good from what I can tell. Might be worth setting up a full-app vdiff test locally and running that before/after your changes to verify it looks exactly the same

JamieTran:
> looks good from what I can tell. Might be worth setting up a full-app vdiff test locally and running that before/after your changes to verify it looks exactly the same

This sounds like a good idea, I'll follow https://github.com/BrightspaceUI/visual-diff to verify the changes didn't affect the UI

JamieTran:
Tests Passed 
image

PR #70, opened 2020-08-19
Vy/use for loop

vix633:


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -68,17 +69,17 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
...
+			const fileSystemType = files[i].m_fileSystemType;
+			const fileId = files[i].m_id;
 
 			const sirenEntity = await window.D2L.Siren.EntityStore.get(this.href, this.token);
-			const link = sirenEntity.getLinkByRel('https://activities.api.brightspace.com/rels/attachments');
+			const link = sirenEntity.getLinkByRel(Rels.Activities.feedbackAttachments);
 			const entitytemp = await window.D2L.Siren.EntityStore.fetch(link, this.token);
 
-			const entity = new AttachmentCollectionEntity(entitytemp.entity, this.token);
+			const entity = new AttachmentCollectionEntity(entitytemp.entity, this.token, { remove: () => { } });
```
AlexBedley, line 80: I don't think the 3rd parameter is needed

PR #71, opened 2020-08-19
Add outcomes to right panel

zhangium:
- use the `d2l-activity-alignments` component in place of `d2l-outcomes-level-of-achievements` (still more work to be done to make it match design)
- fix the outcomes demo page and add demo data
- relies on https://github.com/Brightspace/d2l-activity-alignments/pull/131 to make the outcomes show up properly, otherwise I think it shows up as readonly for now


zhangium:
Screenshot! You may see a readonly version of outcomes if the PR at d2l-activity-alignments isn't merged to accept the actor route.

image

PR #72, opened 2020-08-19
Remove .skip from visualdiff tests

JamieTran:
Some tests were skipped and needed to be checked for the addition of StyleLint changes


AlexBedley:
one thing to note is we're unskipping them because they aren't run in CI. so if they are ever re-activated in CI, we want these tests to run as well.

PR #73, opened 2020-08-25
Us118894/grade timing

zhangium:
- add debouncer to grade item
- will need grades repo change https://github.com/BrightspaceUILabs/grade-result/pull/new/US118894/grade-timing
- branched from @AlexBedley 's branch with the Awaiter


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -48,6 +48,13 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 			e => {
 				if (e.detail.key === 'd2l-provider-html-editor-enabled') e.detail.provider = true;
 			});
+		window.addEventListener('d2l-flush', this.flush.bind(this));
```
AlexBedley, line 51: the event listener needs to be removed at some point, also not sure if it should be added here or in `connected` (same in the grades file)

dbatiste: Yeah any wire-up of event handlers outside the local DOM scope needs to be cleaned up or it will be a leak - the handler will continue to exist and hold reference to everything in its closure, prevent it all from being garbage collected.

To facilitate that, using `connectedCallback` and `disconnectedCallback` is preferred.  The `connectedCallback` runs later that the `constructor` though - hopefully that would not be problematic, otherwise the add/removal becomes more complicated.

components/consistent-evaluation-page.js
```diff
@@ -236,9 +243,17 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 			newGradeVal = e.detail.grade.score;
 		}
 		this.evaluationEntity = await this._controller.transientSaveGrade(entity, newGradeVal);
+
+		unlock();
 	}
 
 	async _saveEvaluation() {
+		window.dispatchEvent(new CustomEvent('d2l-flush', {
```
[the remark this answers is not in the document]
zhangium: I didn't end up doing a refactor as I think the way we have it is more straightforward

AlexBedley:
Gonna close this PR and open a new one where I am the author

PR #74, opened 2020-08-26
add consistent eval telemetry mixin

zhangium:
Made a mixin to do telemetry stuff!
I'm not too sure if using onload and awaiting updateComplete is a good measurement so it would be good to get thoughts on that.


zhangium:
relates to this guy: https://github.com/Brightspace/lms/pull/724 (change to pass the endpoint in)

PR #75, opened 2020-08-26
Add visual diff tests for grade results

JamieTran:
Added component-based visual diff tests for grade results:

- number grade
- letter grade


JamieTran:
squashed all visual-diff tests into 1PR

PR #76, opened 2020-08-27
Added visual diff tests for right panel components

JamieTran:
Added component-based visual diff tests for right panel components.
Grade results:

- number grade
- letter grade

Outcomes:

- default
- with description

Rubric:

- default


test/perceptual/consistent-evaluation-right-panel.visual-diff.js
```diff
@@ -18,42 +18,42 @@ describe('d2l-consistent-evaluation', () => {
 
 	after(async() => await browser.close());
 
-	it('renders the right panel', async function() {
+	it.skip('renders the right panel', async function() {
```
jeroach, line 21: Do we still want to skip all of these?

test/perceptual/feedback-actions.visual-diff.js
```diff
@@ -19,7 +19,7 @@ describe('d2l-consistent-evaluation-feedback-actions', () => {
 	after(async() => await browser.close());
 
 	[ 'empty', 'all', 'some-with-all', 'some' ].forEach((name) => {
-		it(name, async function() {
+		it.skip(name, async function() {
```
jeroach, line 22: Same comment re: skip

test/perceptual/consistent-evaluation.visual-diff.js
```diff
@@ -18,7 +18,7 @@ describe('d2l-consistent-evaluation', () => {
 
 	after(async() => await browser.close());
 
-	it('renders the full app', async function() {
+	it.skip('renders the full app', async function() {
```
jeroach, line 21: Same comment re: skip

test/perceptual/consistent-evaluation-right-panel.visual-diff.js
JamieTran: This one was skipped due to the story of fixing the spacing in the right panel.
I can turn these ones back on 😄

test/perceptual/feedback-actions.visual-diff.js
JamieTran: This component is not used in consistent-eval, it was replaced by d2l-activity-attachments-editor, so I don't think testing this component is necessary.

test/perceptual/consistent-evaluation.visual-diff.js
JamieTran: IMO visual diff tests should be run on components that will not have a lot of change and the full-app render is subject to the most change.

test/perceptual/feedback-actions.visual-diff.js
jeroach: Should we just remove the tests then?

test/perceptual/consistent-evaluation.visual-diff.js
jeroach: that's fair 👍

test/perceptual/feedback-actions.visual-diff.js
JamieTran: Seeing that it isn't being used I'll remove the tests 👌

test/perceptual/consistent-evaluation-right-panel.visual-diff.js
JamieTran: Just kidding, I am going keep these tests off because there seems to be an issue with how outcomes are being rendered in the right-panel-component
image
image

PR #80, opened 2020-09-04
Jroach/us119316

jeroach:
All of: [US119316](https://rally1.rallydev.com/#/detail/userstory/411840195584?fdp=true): [LCB] Create learner context bar component

Some of: [US119916](https://rally1.rallydev.com/#/detail/userstory/422748376172?fdp=true): [LCB] [User Context] Create sub-component for user context within LCB

In Consistent Evaluation:

image

Demo page:

image


components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -0,0 +1,61 @@
...
+				attribute: 'display-name',
+				type: String
+			}
+		};
+	}
+
+	static get styles() {
+		return [bodyStandardStyles, css`
+			.d2l-consistent-evaluation-lcb-user-name {
+				font-weight: bold;
```
jeroach, line 35: Much of this style borrowed from the styling of the user profile image in the top-level navbar of the LMS.

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -0,0 +1,79 @@
...
+			}
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: inline-block;
+				height: 100%;
+				margin: 1rem;
```
jeroach, line 21: Margin to be confirmed with Dan

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -0,0 +1,79 @@
...
+
+			if (lastNameEntity) {
+				return lastNameEntity.properties.name;
+			}
+		}
+		return '';
+	}
+
+	get _colourId() {
+		return 9;
```
jeroach, line 52: This will be set from the `userId` when that is available from the API - placeholder for now.

Behind the scenes, it's just used as the input to a mod of about 20 colours - userId is how it's done elsewhere so figured it would be good to be consistent.

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -0,0 +1,79 @@
...
+				return displayNameEntity.properties.name;
+			}
+		}
+		return undefined;
+	}
+
+	render() {
+		return html`
+			<d2l-consistent-evaluation-lcb-user-context
+				profile-image-href=""
```
jeroach, line 69: Profile image href to come from the API in:
[US120258](https://rally1.rallydev.com/#/detail/userstory/425516084284?fdp=true): [LCB][API][User Context] Add profile image href and user id to user API route

demo/consistent-evaluation-lcb-user-context.html
```diff
@@ -0,0 +1,32 @@
...
+		<title>d2l-consistent-evaluation-lcb-user-context</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<meta charset="UTF-8">
+	</head>
+	<body class="d2l-typography">
+		<d2l-demo-page page-title="d2l-consistent-evaluation-lcb-user-context">
+
+			<d2l-demo-snippet>
+				<d2l-consistent-evaluation-lcb-user-context
```
jeroach, line 21: Demo page using the subcomponent instead of top level since I couldn't quite figure out how to define the `userInfo` object for the demo page.

Open to suggestions if this isn't ideal.

components/consistent-evaluation.js
```diff
@@ -48,6 +50,7 @@ export class ConsistentEvaluation extends MobxLitElement {
 			this._childHrefs = await controller.getHrefs();
 			this._submissionInfo = await controller.getSubmissionInfo();
 			this._gradeItemInfo = await controller.getGradeItemInfo();
+			this._userInfo = await controller.getUserInfo();
```
AllanKerr, line 53: Is there a reason we need to resolve the `userInfo` then pass it up? I would expect to pass up the `userHref` and only resolve the entity in the actual component that needs it.

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -0,0 +1,79 @@
...
+	get _displayName() {
+		if (this.userInfo) {
+			const displayNameEntity = this.userInfo.getSubEntityByRel('https://api.brightspace.com/rels/display-name');
+
+			if (displayNameEntity) {
+				return displayNameEntity.properties.name;
+			}
+		}
+		return undefined;
+	}
```
AllanKerr, line 64: Is there a reason we're opting to not use the `siren-sdk` to handle things like this rather than interfacing with the entities directly?

demo/consistent-evaluation-lcb-user-context.html
AllanKerr: I'd recommend using the top level component. I think you _should_ be able to pass in a JSON string `myData="{&quot;hello&quot;: &quot;world&quot;}"`. Alternatively if you have issues with that we will sometimes create a component specifically for the demo page that wraps the component we want to demo and passes in any context it needs.

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -0,0 +1,61 @@
...
+				margin-left: 0.5rem;
```
AllanKerr, line 36: `rtl` support is needed here

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -0,0 +1,79 @@
...
+				attribute: false,
+				type: Object
+			}
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: inline-block;
```
AllanKerr, line 19: `block` or `flex` might be more appropriate here since I assume we want the context bar to always take up the full page width?

demo/consistent-evaluation-lcb-user-context.html
jeroach: I did try the JSON string but that didn't seem to work.

However, if I go with your suggestion of passing in the href, and then resolving the object in the end class, then that should allow for the top level component to be demoed here.

components/consistent-evaluation.js
jeroach: Good point. I wasn't entirely sure about this, and passing the href also allows for the demo to cover the top level component - will fix.

components/header/d2l-consistent-evaluation-learner-context-bar.js
jeroach: Nope, will fix

components/consistent-evaluation.js
AlexBedley: This is actually the source of some bugs we have in the right panel (passing entity instead of href). We should try and not fetch the entity for as long as possible, so the entity is fresh when we want to use it

components/header/d2l-consistent-evaluation-learner-context-bar.js
AlexBedley: ya https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/users/UserEntity.js looks pretty good

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -0,0 +1,106 @@
...
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+
+		this._displayName = undefined;
+		this._firstName = undefined;
+		this._lastName = undefined;
```
AllanKerr, line 67: Properties start as `undefined` so we normally don't initialize them unless its a different type like a number or boolean. But feel free to leave it as is if you'd rather be explicit.

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -0,0 +1,73 @@
...
+		};
+	}
+
+	static get styles() {
+		return [bodyCompactStyles, css`
+			:host {
+				display: flex;
+			}
+			.d2l-consistent-evaluation-lcb-profile-image {
+				align-self: center;
```
AllanKerr, line 40: You can put `align-items: center;` on the host then you don't need to `align-self` for each child.

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -0,0 +1,73 @@
...
+				margin-left: 0.5rem;
+				max-width: 10rem;
+				min-width: 2rem;
+				overflow: hidden;
+				text-overflow: ellipsis;
+				white-space: nowrap;
+				width: 100%;
+			}
+			:host([dir="rtl"]) .d2l-consistent-evaluation-lcb-user-name {
+				margin-right: 0.5rem;
```
AllanKerr, line 53: You'll want to clear the left margin with `margin-left: 0;`

package.json
```diff
@@ -52,9 +52,11 @@
     "d2l-hypermedia-constants": "^6.36.0",
     "d2l-rubric": "github:Brightspace/d2l-rubric#semver:^3",
     "d2l-tooltip": "github:BrightspaceUI/tooltip#semver:^3",
+    "d2l-users": "github:BrightspaceHypermediaComponents/users#semver:^2",
     "lit-element": "^2",
     "mobx": "^5.15.4",
     "resize-observer-polyfill": "^1.5.1",
-    "stylelint": "^13.6.1"
+    "stylelint": "^13.6.1",
```
AllanKerr, line 59: `stylelint` should be under `devDependencies`

dbatiste: Should move `@brightspace-ui/stylelint-config` as well.

AllanKerr (approved):

PR #81, opened 2020-09-11
Jtran/nav bar

JamieTran:
Apologies for the large PR but the nav-bar feature for consistent eval has been added.

The navbar contains the assignment name
The navbar contains the course name
The Navbar is hooked up to the iterator API (and iterates when you click)

Linked to story [US119994](https://rally1.rallydev.com/#/294005552584d/custom/236803223728?detail=%2Fuserstory%2F422977019400)

image

An issue that came up with the addition of the iterator is that submissions are being appended to the submissions list and are not being cleared between students
image


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -139,4 +139,53 @@ export class ConsistentEvaluationHrefController {
...
+				const organizationResponse = await this._getEntityFromHref(organizationLink, false);
+				if (organizationResponse && organizationResponse.entity) {
+					organizationName = organizationResponse.entity.properties.name;
+				}
+			}
+		}
+		return organizationName;
+	}
+
+	async getIteratorTotal() {
```
JamieTran, line 174: I am not a fan of `getIteratorTotal() `and `getIteratorIndex()` being two different methods. I tried doing something like `getIteratorInfo()` and returning an object with both the `total` and the `index` but when it got down to the nav-bar component, I could not get the object to update with the correct values.

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -1,74 +1,76 @@
...
-
-					<div slot="left" class="d2l-navigation-header-left">
-						<d2l-navigation-link-back 
-							class="d2l-label-text"
-							@click="${this._backButtonClicked}"
-							text="Back to Quick Eval">
-						</d2l-navigation-link-back>
-					</div>
+			<d2l-navigation-immersive 
+				back-link-href="https://www.d2l.com/"
```
JamieTran, line 54: The component is set up to take the return-href that @amauryperez has developed. It is a matter of passing it down through the multiple tiers and using it here.

amauryperez:
Isn't `package-lock.json` supposed to be excluded from PRs?

components/consistent-evaluation-page.js
```diff
@@ -303,7 +334,16 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 	render() {
 		return html`
 			<d2l-template-primary-secondary primary-overflow="${this._scrollbarStatus}">
-				<div slot="header"><h1>Hello, consistent-evaluation!</h1></div>
+				<div slot="header">
```
amauryperez, line 337: In my local copy I can see `d2l-consistent-evaluation-learner-context-bar` inside the header slot, how come it is not here?

JamieTran: I have not pulled master into my branch in a while, I will change this so it is up to date.

amauryperez:
Looks good to me. My only concern is the `package-lock.json` file, maybe @vix633 can confirm if it is ok to have it as part of the PR or if it should be excluded

vix633:
> Looks good to me. My only concern is the `package-lock.json` file, maybe @vix633 can confirm if it is ok to have it as part of the PR or if it should be excluded

This is totally fine. It's not necessarily a part of the PR but we typically sneak an update of modules whenever we can.

amauryperez:
but isn't this particular file something that gets created locally when working on the repo? This is not `package.json` is the `lock` file

vix633:
> but isn't this particular file something that gets created locally when working on the repo? This is not `package.json` is the `lock` file

This file is basically how we lock our dependencies to certain versions. When you delete this file and run `npm install` locally, it will create this file with new module versions if there are any based on what `package.json` specifies. Essentially, it prevents any automated installs from pulling any package versions that we didn't deliberately commit like Jamie is doing here.

amauryperez (approved):

components/consistent-evaluation.js
```diff
@@ -48,14 +56,21 @@ export class ConsistentEvaluation extends MobxLitElement {
 			this._childHrefs = await controller.getHrefs();
 			this._submissionInfo = await controller.getSubmissionInfo();
 			this._gradeItemInfo = await controller.getGradeItemInfo();
+			this._assignmentName = await controller.getAssignmentName();
+			this._organizationName = await controller.getOrganizationName();
+			this._iteratorTotal = await controller.getIteratorTotal();
+			this._iteratorIndex = await controller.getIteratorIndex();
```
ChrisKraljevicD2L, line 62: Nit: can these controller `get` calls be `Promise.all`'d together? Seems like fetching these in parallel would net a performance gain. Unless they need to be `await`ed serially?

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -142,4 +142,53 @@ export class ConsistentEvaluationHrefController {
...
+
+				if (assignmentResponse && assignmentResponse.entity) {
+					assignmentName = assignmentResponse.entity.properties.name;
+				}
+			}
+		}
+		return assignmentName;
+	}
+
+	async getOrganizationName() {
```
ChrisKraljevicD2L, line 162: This and `getAssignmentName` look very similar in terms of logic. Could the shared code and logic be put in its own function that accepts a parameter to differentiate between getting an assignment name and an organization name?

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	_emitPreviousStudentEvent() { this._dispatchButtonClickEvent('d2l-consistent-evaluation-on-previous-student');}
+	_emitNextStudentEvent() { this._dispatchButtonClickEvent('d2l-consistent-evaluation-on-next-student'); }
+
+	render() {
+		return html`
+			<d2l-navigation-immersive 
+				back-link-href="https://www.d2l.com/"
+				back-link-text="Back to D2L"
```
ChrisKraljevicD2L, line 55: In case this isn't placeholder text, this string should be a langterm.

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -0,0 +1,78 @@
...
+				</div>
+
+				<d2l-navigation-iterator 
+					slot="right"
+					@previous-click=${this._emitPreviousStudentEvent} 
+					@next-click=${this._emitNextStudentEvent}
+					?previous-disabled=${(this.iteratorIndex === 1)}
+					?next-disabled=${(this.iteratorIndex === this.iteratorTotal)}
+					hide-text>
+					<span class="d2l-student-iterator d2l-label-text">User ${this.iteratorIndex} of ${this.iteratorTotal}</span>
```
ChrisKraljevicD2L, line 70: langterm here too

components/consistent-evaluation-page.js
```diff
@@ -309,6 +340,14 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		return html`
 			<d2l-template-primary-secondary primary-overflow="${this._scrollbarStatus}">
 				<div slot="header">
+					<d2l-consistent-evaluation-nav-bar
+						assignment-name=${this.assignmentName}
```
CodeBaboon, line 344: I think these properties should start with `.` (if nothing else to be consistent with https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/81/files#diff-c380577c4ac5f9e72316100b9a3a8a6eR85)

components/header/consistent-evaluation-nav-bar.js
amauryperez: In this case `back-link-text` should only be present if the respective property exist in the consistent-eval web component, if it does not exist then there is no need to set a default text value. Maybe don't add `back-link-text` and `back-link-href` as part of this PR

components/consistent-evaluation.js
```diff
@@ -38,6 +42,10 @@ export class ConsistentEvaluation extends MobxLitElement {
 		this._childHrefs = undefined;
 		this._submissionInfo = undefined;
 		this._gradeItemInfo = undefined;
+		this._assignmentName = ' ';
+		this._organizationName = ' ';
```
CodeBaboon, line 46: Why are these defaulted to `' '`? Could the just be `undefined` or at least empty string?

components/consistent-evaluation.js
```diff
@@ -48,14 +56,21 @@ export class ConsistentEvaluation extends MobxLitElement {
...
 		}
-
 	}
 
 	onNextStudentClick() {
 		this.href = this._childHrefs.nextHref;
 	}
 
+	onPreviousStudentClick() {
+		this.href = this._childHrefs.previousHref;
```
CodeBaboon, line 71:
```suggestion
		this.href = this._childHrefs?.previousHref;
```
Suggest a `?` just in case `_childHrefs` were to be undefined.

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -142,4 +142,53 @@ export class ConsistentEvaluationHrefController {
 			gradeItemName
 		};
 	}
+
+	async getAssignmentName() {
+		const root = await this._getRootEntity(false);
+		let assignmentName;
```
CodeBaboon, line 148: I don't think this variable is necessary. Instead you could just:

`return assignmentResponse.entity.properties.name;` on line 155, and `return undefined;` on line 159.

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -142,4 +142,53 @@ export class ConsistentEvaluationHrefController {
...
+		return iteratorTotal;
+	}
+
+	async getIteratorIndex() {
+		const root = await this._getRootEntity(false);
+		let iteratorIndex;
+		if (root && root.entity) {
+			iteratorIndex = root.entity.properties.iteratorIndex;
+		}
+		return iteratorIndex;
```
CodeBaboon, line 192: Something to consider: if this ever returns `null` here will that cause problems with the `previous-disabled` logic in the nav bar?

CodeBaboon:
Probably this PR should update or add some unit tests.

components/consistent-evaluation.js
dbatiste: It is also unnecessary to assign `undefined` - the property values will already be undefined.  To avoid rendering "undefined" strings, use the `ifDefined` directive when necessary.

demo/consistent-evaluation-nav-bar.html
```diff
@@ -0,0 +1,26 @@
...
+	<head>
+		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script type="module">
+			import '@brightspace-ui/core/components/demo/demo-page.js';
+			import '@brightspace-ui/core/components/typography/styles.js';
+			import '@brightspace-ui/core/components/typography/typography.js';
+			import '../components/header/consistent-evaluation-nav-bar.js';
+		</script>
+		<title>d2l-consistent-evaluation-nav-bar</title>
```
dbatiste, line 12: It is not necessary to render a `<title>` element here.  The `d2l-demo-page` will automatically render the title element based on the value provided for `page-title`.

components/consistent-evaluation.js
JamieTran: Could this also apply to the `onNextStudentClick()` as well?

components/controllers/ConsistentEvaluationHrefController.js
JamieTran: I totally agree, I applied the same fix to `getIteratorTotal()` & `getIteratorIndex()`

components/header/consistent-evaluation-nav-bar.js
JamieTran: You're correct, this is a placeholder text

components/consistent-evaluation.js
JamieTran: I didn't want the component to render in with the text 'undefined' before it was done fetching from the API. Will change to an empty string 😄 .

dbatiste: Normally we would use `ifDefined` directive to avoid that.

components/consistent-evaluation.js
```diff
@@ -48,14 +56,21 @@ export class ConsistentEvaluation extends MobxLitElement {
...
+			this._iteratorIndex = await controller.getIteratorIndex();
 		}
-
 	}
 
 	onNextStudentClick() {
 		this.href = this._childHrefs.nextHref;
 	}
 
+	onPreviousStudentClick() {
```
dbatiste, line 70: Are these handlers meant to be private?  If so, they should be prefixed with a leading `_` for clarity.

JamieTran: updated 👍

components/controllers/ConsistentEvaluationHrefController.js
JamieTran: Good catch, the previous button remains enabled when the iteratorIndex is undefined (which we don't want). Fixed for `next-disabled` as well

CodeBaboon (changes requested):
Code looks good now but still no tests.

CodeBaboon (approved):
I think this LGTM. Not sure if visual diff tests are required or not.

dbatiste:
I'm not sure what the state of visual-diff tests are in this repo, but in core we generally find that a few visual-diff tests provide far better coverage.  We generally only use the unit tests to cover what's not covered visually.

PR #84, opened 2020-09-18
Fix expected annotations viewer host

AlexBedley:
We just forgot to make it https when we originally made this file (couldn't test it since we didn't have a version until now)


JamieTran (approved):
LGTM 👍

PR #85, opened 2020-09-21
Fix submission list appending when iterating through students

JamieTran:
Related to [US120545](https://rally1.rallydev.com/#/294005552584d/custom/236803223728?detail=%2Fuserstory%2F435781871716)

Fixed issue where iterating through students, all evidence for the students would append onto the submissions list. Did this by clearing the list of submissions when initializing the submission entity.

Also fixed a bug where the comments would not update.

submission-list-fix

### DIFF
--- a/components/left-panel/consistent-evaluation-submission-item.js
+++ b/components/left-panel/consistent-evaluation-submission-item.js
@@ -42,7 +42,7 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 				type: String
 			},
 			attachments: {
-				type: Object
+				type: Array
 			},
 			comment: {
 				type: String
@@ -233,13 +233,13 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	}
 
 	_formatDateTime() {
-		this.dateStr = this.dateStr ? new Date(this.dateStr) : undefined;
+		const date = this.dateStr ? new Date(this.dateStr) : undefined;
 
-		const formattedDate = (this.dateStr) ? formatDate(
-			this.dateStr,
+		const formattedDate = (date) ? formatDate(
+			date,
 			{format: 'full'}) : '';
-		const formattedTime = (this.dateStr) ? formatTime(
-			this.dateStr,
+		const formattedTime = (date) ? formatTime(
+			date,
 			{format: 'short'}) : '';
 		return `${formattedDate} ${formattedTime}`;
 	}
--- a/components/left-panel/consistent-evaluation-submission-item.js
+++ b/components/left-panel/consistent-evaluation-submission-item.js
@@ -13,7 +13,6 @@ import { bodySmallStyles, heading3Styles, labelStyles } from '@brightspace-ui/co
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { fileSubmission, textSubmission } from '../controllers/constants';
 import { formatDate, formatTime } from '@brightspace-ui/intl/lib/dateTime.js';
-import { Classes } from 'd2l-hypermedia-constants';
 import { getFileIconTypeFromExtension } from '@brightspace-ui/core/components/icons/getFileIconType';
 import { loadLocalizationResources } from '../locale.js';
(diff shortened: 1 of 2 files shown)


vix633 (approved):

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -182,6 +182,7 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	}
 
 	_initializeSubmissionProperties() {
+		this._comment = '';
```
AlexBedley, line 185: what about resetting `this._attachments`?

This file might be simpler if you get rid of `submissionEntity` and  just pass `_comment` and `_attachments` from `consistent-evaluation-submissions-page.js` (it's already passing half of it's properties). That way this file doesn't need to know about hypermedia

JamieTran: Oh I didnt notice attachments but that needs to be reset as well.
I also agree with simplifying `submission-item` by getting rid of `submissionEntity`

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -138,26 +139,10 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	constructor() {
 		super();
 		this.late = false;
-		this._submissionEntity = undefined;
 		this._date = undefined;
-		this._attachments = [];
-		this._comment = '';
```
AlexBedley, line 144: We might still want the defaults here (haven't been burned one way or the other yet, but for consistencies sake)

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -258,6 +232,8 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	}
 
 	_formatDateTime() {
+		this._date = this.dateStr ? new Date(this.dateStr) : undefined;
```
AlexBedley, line 235: `_date` probably doesn't need to exist anymore (at least as a class variable), it's only used in this function as far as I can tell

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -18,6 +18,12 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 			},
 			token: {
 				type: String
+			},
+			_comment: {
+				type: String
+			},
+			_attachments: {
```
AlexBedley, line 25: Similarly, these don't need to exist as lit properties, or even class properties. They're only used in scope of submission entity below. I would separate `_initializeSubmissionProperties` into `getComment(submissionEntity)` and `getAttachments(submissionEntity)` and just use the values locally in render

AlexBedley (changes requested):

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -258,11 +233,13 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 	}
 
 	_formatDateTime() {
-		const formattedDate = (this._date) ? formatDate(
-			this._date,
+		this.dateStr = this.dateStr ? new Date(this.dateStr) : undefined;
```
AlexBedley, line 236: This won't work (I don't think the Date constructor can take a Date object the second time this function is called). Instead of re-using `this.dateStr` just make a local variable `date`. (also has the advantage of keeping `dateStr` a string)

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -42,10 +41,12 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 			lateness: {
 				type: String
 			},
-			submissionEntity : {
-				attribute: false,
+			attachments: {
 				type: Object
```
AlexBedley, line 45: type should be `Array`

AlexBedley (approved):

PR #86, opened 2020-09-21
Jmurray/sync rubric and grade

jmurray-d2l:


components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -83,7 +87,40 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
...
 		this.hideGrade = false;
 		this.hideFeedback = false;
-		this.hideOutcomes = false;
+		this.hideOutcomes = false;	
+		this.eventcount = 0;	
+
+		this.addEventListener('d2l-rubric-total-score-changed',
+		e => {
+
+			if(this.eventcount <= 1){
```
jmurray-d2l, line 96: I haven't been able to figure this part out yet.  The d2l-rubric component fires `d2l-rubric-total-score-changed` twice before any actual user changes.  So if I leave this part out, any score that has been manually overwritten will get replaced by the current rubric score.  Thoughts on how to fix this?

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -84,6 +88,53 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
...
+				}
+
+				if (this.rubricFirstLoad) {
+					this.rubricFirstLoad = false;
+					return;
+				}
+
+				let score = this.grade.score;
+				let letterGrade = this.grade.letterGrade;
+				if (this.grade.scoreType === GradeType.Letter && this.grade.entity.properties.letterGradeSchemeRanges) {
```
AlexBedley, line 107: This much grade logic in the right-panel makes me nervous, but I guess this is the short term solution because the rubric stuff doesn't do transient autosave?

Ideally, when (if?) rubrics does transient autosave, I'd expect the grade value to come back from the API and we just blindly render it there. But for now this seems like our best bet.

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -84,6 +88,53 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
 		this.hideGrade = false;
 		this.hideFeedback = false;
 		this.hideOutcomes = false;
+		this.rubricFirstLoad = true;
+
+		this.addEventListener('d2l-rubric-total-score-changed',
```
AlexBedley, line 93: In general, we prefer to add listeners in the connectedCallback (and unlisten in the disconnectedCallback to clean up memory) https://lit-element.polymer-project.org/guide/events#add-event-listners-in-connectedcallback

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -68,6 +69,9 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
 			},
 			token: {
 				type: String
+			},
+			rubricFirstLoad:{
```
AlexBedley, line 73: This doesn't need to be a property (we won't re-render based on its value)

components/right-panel/consistent-evaluation-right-panel.js
jmurray-d2l: Agreed, its not ideal, but is the same functionality as the legacy experience.  It's something we'll want to address if we ever move the rubric evaluation into our domain

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -84,6 +85,26 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
...
+		super.connectedCallback();
+		window.addEventListener('d2l-rubric-total-score-changed',
+			e => {
+				this._syncRubricGrade(e);
+			}
+		);
+	}
+
+	disconnectedCallback() {
+		window.removeEventListener('d2l-rubric-total-score-changed',
```
AlexBedley, line 101: this won't work, since the lambda makes a "new" function, it needs to be a reference to the function you added in connected. I think you can just remove the lambda wrapper and pass `_syncRubricGrade` (you'll need to bind it to `this` in the constructor as well)

AlexBedley: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/ce0ac04546295d1366e7fecea7ae12dcac5a3b7e/components/left-panel/consistent-evaluation-submission-item.js#L145 for an example

AlexBedley (approved):

amauryperez (approved):

PR #88, opened 2020-09-23
Fix bug where feedback does not update on iteration

JamieTran:
Related to [US120640](https://rally1.rallydev.com/#/388869855380d/dashboard?detail=%2Fuserstory%2F436813346276)

feedback-fix

### DIFF
--- a/components/right-panel/consistent-evaluation-feedback-presentational.js
+++ b/components/right-panel/consistent-evaluation-feedback-presentational.js
@@ -31,6 +31,9 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 			},
 			token: {
 				type: String
+			},
+			_key: {
+				type: String
 			}
 		};
 	}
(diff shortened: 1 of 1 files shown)


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -95,13 +98,22 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 		await this.shadowRoot.querySelector('d2l-activity-attachments-editor').save();
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (changedProperties.has('feedbackText')) {
+			this._key = this.href;
+		}
```
JamieTran, line 106: When iterating to the next student, the component would receive the href before the feedbackText and would display the previous student's feedback on the current student. This ensures that the feedbackText is updated when we want to re-render the `d2l-activity-text-editor`  component

AlexBedley (approved):

vix633 (approved):

PR #90, opened 2020-09-25
Jtran/nav bar responsiveness

JamieTran:
Related to [US120544](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F435780339580)

Added truncation and tool-tips of course & assignment names
Added responsive-ness to the back button and iterator component

Note: we want the tool-tip to only show on truncation (Milestone 2)

nav-bar-responsiveness

### DIFF
--- a/components/header/consistent-evaluation-nav-bar.js
+++ b/components/header/consistent-evaluation-nav-bar.js
@@ -53,9 +53,6 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 				padding-top: 0.25rem;
 			}
 			.d2l-truncate {
-				-webkit-box-orient: vertical;
-				display: -webkit-box;
-				-webkit-line-clamp: 1;
 				overflow: hidden;
 				overflow-wrap: break-word;
 				text-overflow: ellipsis;
--- a/components/header/consistent-evaluation-nav-bar.js
+++ b/components/header/consistent-evaluation-nav-bar.js
@@ -145,11 +145,11 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 					hide-text>
 
 					<div class="d2l-iterator-space"> 
-						<span class="d2l-iterator-text d2l-label-text">${this.localize('user')} ${this.iteratorIndex} ${this.localize('of')} ${this.iteratorTotal}</span>
+						<span class="d2l-iterator-text d2l-label-text">${this.localize('iteratorText', { num: this.iteratorIndex, total: this.iteratorTotal }) }</span>
 					</div>
 
 				</d2l-navigation-iterator>
-
+				
 			</d2l-navigation-immersive>
 		`;
 	}
--- a/components/header/consistent-evaluation-nav-bar.js
+++ b/components/header/consistent-evaluation-nav-bar.js
@@ -118,8 +118,7 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 
 				<d2l-navigation-link-back 
 					class="d2l-short-back"
-					href=${this.returnHref}
-					text="Back">
+					href=${this.returnHref}>
 				</d2l-navigation-link-back>
 			`;
 		}
--- a/components/header/consistent-evaluation-nav-bar.js
+++ b/components/header/consistent-evaluation-nav-bar.js
@@ -1,6 +1,8 @@
+
 import 'd2l-navigation/d2l-navigation-immersive.js';
 import 'd2l-navigation/components/d2l-navigation-iterator/d2l-navigation-iterator.js';
 import 'd2l-navigation/d2l-navigation-link-back.js';
+import '@brightspace-ui/core/components/tooltip/tooltip.js';
 
 import { css, html, LitElement } from 'lit-element';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
@@ -47,6 +49,15 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 			.d2l-heading-3 {
 				padding-top: 0.25rem;
 			}
+			.d2l-truncate {
+				-webkit-box-orient: vertical;
+				display: -webkit-box;
+				-webkit-line-clamp: 1;
+				overflow: hidden;
+				overflow-wrap: break-word;
+				text-overflow: ellipsis;
+				white-space: nowrap;
+			}
 		`];
 	}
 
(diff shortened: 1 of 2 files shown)


components/header/consistent-evaluation-nav-bar.js
```diff
@@ -73,9 +111,16 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
...
 				<d2l-navigation-link-back 
+					class="d2l-full-back"
 					href=${this.returnHref}
 					text=${ifDefined(this.returnHrefText)} >
 				</d2l-navigation-link-back>
+
+				<d2l-navigation-link-back 
+					class="d2l-short-back"
+					href=${this.returnHref}
+					text="Back">
```
svanherk, line 122: So this would need to be localized, but you actually get that for free with this component: https://github.com/BrightspaceUI/navigation/blob/master/d2l-navigation-link-back.js#L32-L44
If you just remove `text` altogether, it should still say "Back" and handle everything for you

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -40,13 +42,49 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
...
+			.d2l-iterator-text {
 				padding-left: 4rem;
 				padding-right: 4rem;
 			}
 			.d2l-heading-3 {
 				padding-top: 0.25rem;
 			}
+			.d2l-truncate {
+				-webkit-box-orient: vertical;
+				display: -webkit-box;
```
svanherk, line 57: This is non-standard and deprecated… it seems like what you actually want here is flexbox (https://caniuse.com/flexbox)?  `display: flex;` and `flex-direction: column`?  You may need to wrap them in a div and place the `display: flex;` at that level, but there's tons of info online - it's super powerful and will definitely do what you need

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -40,13 +42,49 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
...
+				-webkit-line-clamp: 1;
```
svanherk, line 58: This won't work on all the browsers clients are using (it's fairly new in Firefox and a lot of schools lock it to older versions), but I don't think you actually need it?  You have `overflow: hidden;`, `white-space: nowrap;` and `text-overflow: ellipsis;` already, which I would expect will work nicely with flexbox.  It can be super finicky though, I can help debug if it's not working as you expect without `-webkit-line-clamp: 1;`

JamieTran: You're right, it turns out the `-webkit` stuff isn't needed at all. Adding `display: flex` removes the ellipsis on the truncation. Not defining `display` fixes it.

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -101,7 +144,10 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 					?previous-disabled=${(this.iteratorIndex === 1 || this.iteratorIndex === undefined)}
 					?next-disabled=${(this.iteratorIndex === this.iteratorTotal || this.iteratorIndex === undefined || this.iteratorIndex === undefined)}
 					hide-text>
-					<span class="d2l-student-iterator d2l-label-text">${this.localize('user')} ${this.iteratorIndex} ${this.localize('of')} ${this.iteratorTotal}</span>
+
+					<div class="d2l-iterator-space"> </div>
```
svanherk, line 148: Is this to make more space between the iterators when the text disappears?  I'm surprised this isn't causing things to be slightly off-center (or maybe it is and I can't tell from the gif lol).
If it is, seems like it would be good to put the `span` inside the `div` to add the padding to each side?
```
<div class="d2l-iterator-space">
     <span class="d2l-iterator-text d2l-label-text">${this.localize('user')} ${this.iteratorIndex} ${this.localize('of')} ${this.iteratorTotal}</span>
</div>
```
If you do that, I expect you could also apply a min-width to the `div` of `1rem` instead of the padding, but the padding should work too

JamieTran: Yeah exactly, the div is there to make it so there is a gap between the iterators (without it they would be touching). I don't notice any sort of off-center-ness but I do like the idea of putting the `span` inside the `div` to tidy things up and for peace of mind.

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -101,7 +143,11 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 					?previous-disabled=${(this.iteratorIndex === 1 || this.iteratorIndex === undefined)}
 					?next-disabled=${(this.iteratorIndex === this.iteratorTotal || this.iteratorIndex === undefined || this.iteratorIndex === undefined)}
 					hide-text>
-					<span class="d2l-student-iterator d2l-label-text">${this.localize('user')} ${this.iteratorIndex} ${this.localize('of')} ${this.iteratorTotal}</span>
+
+					<div class="d2l-iterator-space">
```
dbatiste, line 147: If we want this to be consistent, this `div` could be added [here](https://github.com/BrightspaceUI/navigation/blob/master/components/d2l-navigation-iterator/d2l-navigation-iterator.js#L58).

components/header/consistent-evaluation-nav-bar.js
```diff
@@ -101,7 +143,11 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 					?previous-disabled=${(this.iteratorIndex === 1 || this.iteratorIndex === undefined)}
 					?next-disabled=${(this.iteratorIndex === this.iteratorTotal || this.iteratorIndex === undefined || this.iteratorIndex === undefined)}
 					hide-text>
-					<span class="d2l-student-iterator d2l-label-text">${this.localize('user')} ${this.iteratorIndex} ${this.localize('of')} ${this.iteratorTotal}</span>
+
+					<div class="d2l-iterator-space"> 
+						<span class="d2l-iterator-text d2l-label-text">${this.localize('user')} ${this.iteratorIndex} ${this.localize('of')} ${this.iteratorTotal}</span>
```
svanherk, line 148: This is unrelated to what you're doing here, but just wanted to call it out... I don't think this will translate nicely in a lot of languages, especially RTL languages that might expect the order switched.  Usually for something like this we'd have one langterm with built-in variables:
```
"iteratorText": "User {num} of {total}"
```
And then you use it like:
```
${this.localize('iteratorText', { num: this.iteratorIndex, total: this.iteratorTotal } }
```
That way, if the translation is something like `Of {total} viewing {num}`, it'll just "work".  This is a kinda silly example, but hopefully it makes sense - usually translating individual words in a defined order leads to awkward phrases in other languages.

svanherk (approved):
This looks good to me!  Also, just a heads up, there is a story in the Gaudi backlog to eventually handle the shortening of the back text for you: https://rally1.rallydev.com/#/15545167705d/dashboard?detail=%2Fuserstory%2F402825544948&fdp=true

It's a while away, but once that's in I can let you know!  I don't think it will be a breaking change.

components/header/consistent-evaluation-nav-bar.js
svanherk: Yeah, that probably does make sense... if someone is hiding the iterator text, us handling the minimum space consistently would be good.  FYI @geurts we could add this to the … although I don't think we have a design for the nav iterators, should we?

PR #92, opened 2020-09-29
Vy/lcb file context

vix633:
Initial pass on the LCBF
image


components/consistent-evaluation.js
```diff
@@ -24,10 +24,12 @@ export class ConsistentEvaluation extends MobxLitElement {
 			_childHrefs: { type: Object },
 			_submissionInfo: { type: Object },
 			_gradeItemInfo: { type: Object },
+			_submissionFiles: {type: Object},
```
jeroach, line 27: Where is this used?

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -144,19 +116,16 @@ export class ConsistentEvaluationLeftPanel extends LocalizeMixin(LitElement) {
...
 	}
 
 	_renderTextEvidence() {
 		return html`
 		<d2l-consistent-evaluation-evidence-text
-			title=${this._textEvidence.title}
-			date=${this._textEvidence.date}
-			download-url=${this._textEvidence.downloadUrl}
-			.content=${this._textEvidence.content}
-			@d2l-consistent-evaluation-evidence-back-to-user-submissions=${this._showSubmissionList}
```
jeroach, line 159: Don't we still need this event listener? I assume we still want to be able to return to the list of submissions from a text based submission

demo/consistent-evaluation-lcb-file-context.html
```diff
@@ -0,0 +1,28 @@
...
+		<title>d2l-consistent-evaluation-lcb-file-context</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<meta charset="UTF-8">
+	</head>
+	<body class="d2l-typography">
+		<d2l-demo-page page-title="d2l-consistent-evaluation-lcb-file-context">
+
+			<d2l-demo-snippet>
+				<d2l-consistent-evaluation-lcb-file-context
```
jeroach, line 21: I don't think we want a demo for the subcomponent (similar to how we didn't want one for the user-context subcomponent).

I would just expect us to update the existing demo for the parent `d2l-consistent-evaluation-learner-context-bar`

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -86,40 +86,12 @@ export class ConsistentEvaluationLeftPanel extends LocalizeMixin(LitElement) {
...
-	}
-
 	_renderFileEvidence() {
 		return html`
 		<d2l-consistent-evaluation-evidence-file
-			.url=${this._fileEvidenceUrl}
+			.url=${this.fileEvidenceUrl}
 			.token=${this.token}
-			@d2l-consistent-evaluation-evidence-back-to-user-submissions=${this._showSubmissionList}
+			@d2l-consistent-evaluation-evidence-back-to-user-submissions=${this.showSubmissionList}
```
jeroach, line 94: Where is `showSubmissionList` defined?

I see that we delete `_showSubmissionList`, but I don't see a `showSubmissionList` either added or passed in.

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -1,63 +1,77 @@
...
+	}
+
+	render() {
+		return html`
+			<d2l-consistent-evaluation-lcb-user-context
+				href="${this.href}"
+				.token="${this.token}"
+			></d2l-consistent-evaluation-lcb-user-context>
+			<d2l-consistent-evaluation-lcb-file-context
+			.submissionInfo="${this.submissionInfo}"
```
jeroach, line 69: nit: indentation

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,131 @@
...
+				type: String
+			}
+		};
+	}
+
+	static get styles() {
+		return [selectStyles,
+			css`
+				:host {
+					display: inline-block;
```
jeroach, line 33: Since both `file-context` and `user-context` are both setting `display: inline-block;` can we just set it on the parent `learner-context-bar` ?

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,131 @@
...
+
+	render() {
+		return html`
+        <select
+		class="d2l-input-select"
+		@change=${this._onSelectChange}
+		>
+		<option label='User Submissions' value=${submissions} ?selected=${this.selectedItemName === submissions}></option>
+                    ${this._files && this._files.map(submission => html`
+						<optgroup label=${`Submission ${submission.submissionNumber}`}>
```
jeroach, line 120: The word `Submission` needs to be a lang term

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,131 @@
...
+		this.dispatchEvent(event);
+	}
+
+	render() {
+		return html`
+        <select
+		class="d2l-input-select"
+		@change=${this._onSelectChange}
+		>
+		<option label='User Submissions' value=${submissions} ?selected=${this.selectedItemName === submissions}></option>
```
jeroach, line 118: `User Submissions` needs to be a lang term

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,131 @@
...
+				i++;
+				return file;
+			});
+			return Promise.all(submissionEntities);
+		}
+
+	}
+
+	getSubmissionFiles(submission) {
+		const attachments = submission.entity.getSubEntityByRel('https://assignments.api.brightspace.com/rels/attachment-list');
```
jeroach, line 74: nit: use const for rel

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,131 @@
...
+				bubbles: true
+			});
+			this.dispatchEvent(event);
+			return;
+		}
+		const submissionFile = JSON.parse(e.target.value);
+		this._dispatchRenderEvidenceFileEvent(submissionFile);
+	}
+
+	_dispatchRenderEvidenceFileEvent(sf) {
```
jeroach, line 100: Do we not need another method to dispatch `d2l-consistent-evaluation-submission-item-render-evidence-text` ?

How does `_onSelectChange()` know that the selected file in question was a file and not a text submission. Right now it seems to just assume it's a file submission, unless I'm missing something.

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,131 @@
...
+			let i = 1;
+			const submissionEntities = await this.submissionInfo.submissionList.map(async sub => {
+				const file = await window.D2L.Siren.EntityStore.fetch(sub.href, this.token, false);
+				file.submissionNumber = i;
+				i++;
+				return file;
+			});
+			return Promise.all(submissionEntities);
+		}
+
```
jeroach, line 70: Do we need to return a failing promise if `this.submissionInfo === undefined` ?

components/left-panel/consistent-evaluation-left-panel.js
jeroach: Actually I think this is fine, and we actually want to remove it from the other render method.

components/left-panel/consistent-evaluation-left-panel.js
jeroach: As per my other comment, I think this line needs to go, as we want the event to bubble up and be caught and handled in `consistent-evaluation-page`

components/header/d2l-consistent-evaluation-lcb-file-context.js
jeroach: Used `flex` instead

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -1,63 +1,77 @@
...
+
+	render() {
+		return html`
+			<d2l-consistent-evaluation-lcb-user-context
+				href="${this.href}"
+				.token="${this.token}"
+			></d2l-consistent-evaluation-lcb-user-context>
+			<d2l-consistent-evaluation-lcb-file-context
+				.submissionInfo="${this.submissionInfo}"
+				.selectedItemName=${this.selectedItemName}>
```
AllanKerr, line 70: Similar comment as above; is there a reason we don't want this to be an attribute?

components/header/d2l-consistent-evaluation-lcb-file-context.js
jeroach: Fixed

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,148 @@
...
+			const event = new Event('d2l-consistent-evaluation-evidence-back-to-user-submissions', {
+				composed: true,
+				bubbles: true
+			});
+			this.dispatchEvent(event);
+			return;
+		}
+
+		const submission = JSON.parse(e.target.value);
+		if (submission.comment === undefined) {
```
jeroach, line 96: Not sure if this is the best way to identify submission type (file or text) - but `comment` is a property that exists on text submissions that doesn't exist on file submissions.

components/consistent-evaluation-page.js
```diff
@@ -357,14 +394,16 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
+						.selectedItemName=${this._selectedFile}
 					></d2l-consistent-evaluation-learner-context-bar>
 				</div>
 				<div slot="primary" class="d2l-consistent-evaluation-page-primary-slot">
 					<d2l-consistent-evaluation-left-panel
 						.submissionInfo=${this.submissionInfo}
 						.token=${this.token}
-						@d2l-consistent-evaluation-left-panel-render-evidence=${this._hideScrollbars}
-						@d2l-consistent-evaluation-left-panel-render-submission-list=${this._showScrollbars}
+						.fileEvidenceUrl=${this._fileEvidenceUrl}
```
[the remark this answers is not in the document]
jeroach: For some reason when I try to rename this one I get this super odd error:
image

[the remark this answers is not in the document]
AllanKerr: Is it possible it wasn't renamed everywhere or `attribute: false` wasn't changed causing it to be unexpectedly `undefined`? The error looks like its from trying to query a bad route in the LMS.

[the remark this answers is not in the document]
jeroach: I've tried this twice now...maybe I'm just missing something

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,125 @@
...
+			this._files = await this.getSubmissions();
+		}
+	}
+
+	async getSubmissions() {
+		if (this.submissionInfo) {
+			let i = 1;
+			const submissionEntities = await this.submissionInfo.submissionList.map(async sub => {
+				const file = await window.D2L.Siren.EntityStore.fetch(sub.href, this.token, false);
+				file.submissionNumber = i;
```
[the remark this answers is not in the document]
jeroach: What you say makes sense :)

Any thoughts to an alternative?

components/consistent-evaluation-page.js
```diff
@@ -357,14 +394,16 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 					<d2l-consistent-evaluation-learner-context-bar
 						href=${ifDefined(this.userHref)}
 						.token=${this.token}
+						.submissionInfo=${this.submissionInfo}
```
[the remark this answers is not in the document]
jeroach: Thoughts on this, given we were okay to leave it in the other PR?

components/header/d2l-consistent-evaluation-lcb-file-context.js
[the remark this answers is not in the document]
AllanKerr: I _think_ you'll be able to get away with just moving the `submissionNumber` assignment and increment above the `await`. Although it'd probably be preferable to just get the index number from `map` and add `1` to it.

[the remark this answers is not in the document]
jeroach: Awesome, thanks! This worked nicely.

components/consistent-evaluation-page.js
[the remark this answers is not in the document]
AllanKerr: I'm okay leaving it as but you might have more context on the right decision. If its something that is going to need to be mutated in the future then it should probably be fixed now. If its only ever going to be read-only, then leaving it as is is fine as long as everyone watches out to make sure it doesn't get mutated.

components/header/d2l-consistent-evaluation-lcb-file-context.js
vix633: I don't think we need to "handle" the text submissions differently anymore. I talked with Dan and we expect that the text subs use the same file viewer as the rest. So I previously removed all these handlers.

jeroach: Discussed offline - was some confusion between text assignments and text submissions to a file based assignment

components/consistent-evaluation-page.js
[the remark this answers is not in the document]
jeroach: Finally fixed :)

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,148 @@
...
+		const submission = JSON.parse(e.target.value);
+		if (submission.comment === undefined) {
+			this._dispatchRenderEvidenceFileEvent(submission);
+		} else {
+			this._dispatchRenderEvidenceTextEvent(submission);
+		}
+	}
+
+	_dispatchRenderEvidenceFileEvent(sf) {
+		const event = new CustomEvent('d2l-consistent-evaluation-submission-item-render-evidence-file', {
```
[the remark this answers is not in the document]
jeroach: As discussed, this will not be fixed here

components/consistent-evaluation-page.js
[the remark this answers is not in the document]
jeroach: As discussed, leaving this as is for now

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -0,0 +1,146 @@
...
+		super.updated(changedProperties);
+
+		if (changedProperties.has('submissionInfo')) {
+			this._files = await this.getSubmissions();
+		}
+	}
+
+	async getSubmissions() {
+		if (this.submissionInfo) {
+			const submissionEntities = await this.submissionInfo.submissionList.map(async(sub, index) => {
```
AllanKerr, line 60: This `await` should be remove since `map` returns an array of promises rather than a promise itself.

AllanKerr (approved):

PR #94, opened 2020-09-30
Update lcb to display profile image and use user href

jeroach:
[US119916](https://rally1.rallydev.com/#/detail/userstory/422748376172?fdp=true): [LCB] [User Context] Create sub-component for user context within LCB

Profile image:
image

No profile image:
image

### DIFF
--- a/components/header/d2l-consistent-evaluation-lcb-user-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-user-context.js
@@ -1,22 +1,17 @@
 import 'd2l-users/components/d2l-profile-image.js';
 import { css, html, LitElement } from 'lit-element';
 import { bodyCompactStyles } from '@brightspace-ui/core/components/typography/styles.js';
+import { EntityMixinLit } from 'siren-sdk/src/mixin/entity-mixin-lit.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { UserEntity } from 'siren-sdk/src/users/UserEntity.js';
 
-export class ConsistentEvaluationLcbUserContext extends RtlMixin(LitElement) {
+export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(LitElement)) {
 
 	static get properties() {
 		return {
-			displayName: {
-				attribute: 'display-name',
-				type: String
-			},
-			profileImageHref: {
-				attribute: 'profile-image-href',
-				type: String
-			},
-			token: {
+			_displayName: {
+				attribute: false,
 				type: String
 			}
 		};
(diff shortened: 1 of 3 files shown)


components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -55,13 +46,11 @@ export class ConsistentEvaluationLcbUserContext extends RtlMixin(LitElement) {
 
 	render() {
 		return html`
-			<d2l-profile-image-base
+			<d2l-profile-image
```
jeroach, line 49: By using `d2l-profile-image` instead of the base class, we can simply pass the user href, and the component handles the profile image, first name, last name, and colour id

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -55,13 +46,11 @@ export class ConsistentEvaluationLcbUserContext extends RtlMixin(LitElement) {
 
 	render() {
 		return html`
-			<d2l-profile-image-base
+			<d2l-profile-image
 				href="${this.profileImageHref}"
-				first-name="${ifDefined(this.firstName)}"
-				last-name="${ifDefined(this.lastName)}"
-				colour-id="${this.colourId}"
+				token="${this.token}"
```
AllanKerr, line 51: You'll need to use the property syntax here: `.token="${this.token}"`

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -78,22 +70,14 @@ export class ConsistentEvaluationLearnerContextBar extends (EntityMixinLit(RtlMi
...
 
 	render() {
 		return html`
 			<d2l-consistent-evaluation-lcb-user-context
-				profile-image-href=""
-				first-name="${ifDefined(this._firstName)}"
-				last-name="${ifDefined(this._lastName)}"
-				colour-id="${this._colourId}"
+				profile-image-href="${this.href}"
 				display-name="${ifDefined(this._displayName)}"
```
AllanKerr, line 79: You're welcome to leave it as is but it might make more sense to have the `EntityMixinLit` on `d2l-consistent-evaluation-lcb-user-context` instead of the `d2l-consistent-evaluation-learner-context-bar`. That way, all it needs is an `href` and `token` and can then resolve `displayName` itself.

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -1,24 +1,15 @@
...
-import { UserEntity } from 'siren-sdk/src/users/UserEntity.js';
 
-export class ConsistentEvaluationLearnerContextBar extends (EntityMixinLit(RtlMixin(LitElement))) {
+export class ConsistentEvaluationLearnerContextBar extends RtlMixin(LitElement) {
 
 	static get properties() {
 		return {
-			_displayName: {
-				attribute: false,
+			href: {
```
JamieTran, line 9: nit: rename to userHref to keep things clear

AllanKerr (approved):

JamieTran (approved):

PR #95, opened 2020-10-01
Jmurray/token receiver

jmurray-d2l:


jmurray-d2l:
Relies on this PR for backend https://github.com/Brightspace/lms/pull/1936

JamieTran:
Would all the subcomponents of the right & left panel need to have the `token` type changed from `String` to `Object` as well? (e.g. rubrics, outcomes, submission pages, etc.)

jmurray-d2l:
That's a good point, I will update them.  Another thing I don't understand, the whole app works as expected without changing them.  I think I will wait for Alex to take a look before merging this

jmurray-d2l:
Okay, looking to put this in as is, please review! :)

JamieTran (approved):

components/controllers/ConsistentEvaluationController.js
```diff
@@ -29,14 +29,6 @@ export class ConsistentEvaluationController {
 			throw new Error(ConsistentEvaluationControllerErrors.INVALID_TYPE_EVALUATION_HREF);
 		}
 
-		if (!token) {
```
AlexBedley, line 32: Won't we still want this check?

components/controllers/ConsistentEvaluationController.js
```diff
@@ -29,14 +29,6 @@ export class ConsistentEvaluationController {
...
-			throw new Error(ConsistentEvaluationControllerErrors.INVALID_TOKEN);
-		}
-
-		if (typeof token !== 'string') {
-			throw new Error(ConsistentEvaluationControllerErrors.INVALID_TYPE_TOKEN);
```
AlexBedley, line 37: Can we delete this constant then?

AlexBedley (approved):

AlexBedley:
> That's a good point, I will update them. Another thing I don't understand, the whole app works as expected without changing them. I think I will wait for Alex to take a look before merging this

The property `type` is only really used when serializing/deserializing from attribute to property. I suspect we use `.token` everywhere (as opposed to just `token`) so we're not running into the converter code

components/controllers/ConsistentEvaluationController.js
jmurray-d2l: I couldn't get it to work, this would crash the app so I had to remove it.  Open to thoughts on how to keep it, but I'm stumped at this point

AlexBedley: that seems pretty suspicious, it implies that `this._token` changes before `_initializeController` calls the constructor

PR #96, opened 2020-10-01
US120317 - LCB - Exemptions and Late Subs - Handle if Learner is exempt

amauryperez:


components/consistent-evaluation-page.js
```diff
@@ -357,6 +357,7 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 					<d2l-consistent-evaluation-learner-context-bar
 						href=${ifDefined(this.userHref)}
 						.token=${this.token}
+						.submissionInfo=${this.submissionInfo}
```
jeroach, line 360: @AllanKerr had a similar comment on Viktor's PR around passing the `submissionInfo` entity like this:
https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/92/files#r498311067

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -30,15 +36,26 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
...
 				text-overflow: ellipsis;
 				white-space: nowrap;
-				width: 100%;
 			}
 			:host([dir="rtl"]) .d2l-consistent-evaluation-lcb-user-name {
 				margin-left: 0;
 				margin-right: 0.5rem;
 			}
+			.d2l-consistent-evaluation-lcb-is-exempt {
+				font-style: italic;
```
jeroach, line 45: @AllanKerr is this approach fine for text, or is there a style we should be applying from https://github.com/BrightspaceUI/core/tree/master/components/typography ?

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -55,6 +59,7 @@ export class ConsistentEvaluationLearnerContextBar extends RtlMixin(LitElement)
 			<d2l-consistent-evaluation-lcb-user-context
 				href="${this.href}"
 				.token="${this.token}"
+				?is-exempt="${this.submissionInfo.isExempt}"
```
jeroach, line 62: If we just need the boolean value of `isExempt`, can we just pass that in instead of the `submissionInfo` object? Not sure if this would impact the issue mentioned above around the passing of `submissionInfo`

jeroach (commented):
Some questions - looking for perspective from @AllanKerr as some of the issues were the same mentioned on the other PR.

components/header/d2l-consistent-evaluation-learner-context-bar.js
amauryperez: I'm passing the whole thing because I saw Viktor's PR where he needs more stuff

components/consistent-evaluation-page.js
amauryperez: I just copied from a different place in the repo that is doing the same thing. @AllanKerr what is the recommended approach here?

AllanKerr: This will work so I'm hesitant to recommend significant refactoring. Everyone will just need to be careful not to mutate `submissionInfo` since any changes to the `submissionInfo` object won't trigger re-renders unless careful copying and / or `requestUpdate` calls are used.

Normally I'd encourage doing this with the siren-sdk and passing down hrefs rather than objects. That way we can resolve the entity internally from the `evaluationRel` `href` inside components that need information about it.

components/header/d2l-consistent-evaluation-lcb-user-context.js
AllanKerr: We definitely want to be using typography for styling text so that font styles are consistent across tools. You can then apply your `font-style: italic` in addition to whichever style you use from `typography` just by applying both classes.

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -60,6 +77,14 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
 		this._displayName = userEntity.getDisplayName();
 	}
 
+	_getExemptText() {
+		if (this.isExempt) {
+			return html`<span class="d2l-consistent-evaluation-lcb-is-exempt">(${this.localize('exempt')})</span>`;
+		} else {
+			return html``;
```
AllanKerr, line 84: This should work fine but you can also just return `null` if you don't want anything to be rendered.

components/header/d2l-consistent-evaluation-lcb-user-context.js
amauryperez: no sure what this means :-), is there a class somewhere that already has the italic style that I should used? Sorry if the question makes no sense, I have pretty much zero knowledge of UI development :-)

AllanKerr: No need to apologize, I'm always happy to help! For some context, in BrightspaceUI core we define a bunch of different font styles that can be imported to keep fonts consistent. So generally we want to use these fonts.

To start off you'll need to import the styles you want:
```javascript
import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
```
Then you add them to the `styles` block so you can actually use them:
```javascript
  static get styles() {
    return [ bodyStandardStyles , css`
      // the CSS you already have will be here
    ` ];
  }
```

Now you can use the `d2l-body-standard` class on your `span` to give it the standard styles. But since you also need it to be italicized we'll need that to. So we'd add both classes:
```javascript
<span class="d2l-body-standard d2l-consistent-evaluation-lcb-is-exempt">(${this.localize('exempt')})</span>
```

amauryperez: cool, I'll give that a try

components/header/d2l-consistent-evaluation-learner-context-bar.js
jeroach: Good point :)

jeroach (approved):

AllanKerr:
These changes look good to me. My only concern is that I think a couple of the `lcb` demo pages might be broken since `submissionInfo` will be `undefined`

amauryperez:
I'll take a look

amauryperez:
@AllanKerr any ideas why the on looks like it is still queued, but it actually seems to be completed?

AllanKerr:
> @AllanKerr any ideas why the on looks like it is still queued, but it actually seems to be completed?

It looks like just an issue on the Travis end syncing to Github. You'll likely need to retrigger; although, it does look like it has passed so merging is probably fine with the current status.

PR #98, opened 2020-10-05
US117787 - Add Confirmation on Unsaved Changes

max-hao-d2l:
Add a dialog that confirms unsaved changes.
* The dialog will show up if the user has changed any editable component and has not updated, saved, or published the evaluation.
* For now, this component is only hooked up to the feedback and coa-eval-override components and only shows up in the context of evaluating COA outcomes.
* Note that this PR will only get merged once `brightspace-integration` and `d2l-activity-alignments` both get updated to use the latest major version of `d2l-outcomes-level-of-achievement` (v3). Otherwise, there will be subdependency conflicts (as outlined by the failing CI).
https://rally1.rallydev.com/#/detail/userstory/398566207612?fdp=true


components/consistent-evaluation.js
```diff
@@ -99,6 +103,7 @@ export class ConsistentEvaluation extends MobxLitElement {
 				.token=${this.token}
 				?rubric-read-only=${this._rubricReadOnly}
 				?rich-text-editor-disabled=${this._richTextEditorDisabled}
+				?confirm-unsaved-changes=${this._shouldConfirmUnsavedChanges()}
```
jeroach, line 106: Parenthesis not needed here

jeroach (approved):
Looks good to me - just one small item

components/consistent-evaluation.js
max-hao-d2l: `confirm-unsaved-changes` in this case is a boolean attribute. If I removed the parentheses, wouldn't it set the attribute to always true? (since `this._shouldConfirmUnsavedChanges` without the parentheses will never be undefined)

jeroach: My mistake - you are right :)

PR #99, opened 2020-10-06
Jmurray/reset permissions on publish retract

jmurray-d2l:
We are using the existence of actions the determine write and delete permissions, but the existence of actions can also depend on published state, so we need to check the permissions any time the published state changes


amauryperez (approved):

AlexBedley:
This change is fine but I think we'll keep running into these dirty state problems unless we refactor. We really need a single point that refreshes the cached entity (along with all the extra little helper variables). For example, does an update or save need to update these cached variables?

AlexBedley (approved):

jmurray-d2l:
They don't, only because they don't alter the published state.  I agree though, I expected to get this for free upon publishing or retracting

PR #101, opened 2020-10-07
Modify Navbar to align with COA Design

max-hao-d2l:
For COA manual eval, the navbar should contain the following:

* The student's name should be the title
* The words "Overall Achievement" should be in the subtitle
* Both the iterator buttons and the LCB should be hidden (for now)

Functionality outside the COA manual eval context should be unaffected
https://rally1.rallydev.com/#/detail/userstory/408254862364?fdp=true


jeroach (approved):
LGTM

PR #102, opened 2020-10-07
- Fixing issue with letter grades not updating for holistic rubrics

jmurray-d2l:


jeroach (approved):

AlexBedley:
hey got any more context on this change?

jmurray-d2l:
Sure yeah, holistic rubrics are percentage based, so when they are graded the event does not contain ` e.detail.outOf` and we just set that to 100.  My previous fix only had this inside the scope of numeric grade items, but we need it for both letter grade and numeric grade items, this just move that fix outside of the scope of checking the grade item scoring type

AlexBedley (approved):

AlexBedley:
Thanks! I needed to change to split view to make it more obvious what was going on

PR #103, opened 2020-10-07
Truncate file name in file context and keep extension

JamieTran:
SC showing truncation in dropdown: 
image
SC showing truncation on selection: 
image

### DIFF
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -39,6 +39,12 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 					margin-left: 0;
 					margin-right: 0.7rem;
 				}
+				.d2l-truncate {
+					overflow: hidden;
+					overflow-wrap: break-word;
+					text-overflow: ellipsis;
+					white-space: nowrap;
+				}
 			`
 		];
 	}
@@ -129,7 +135,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 
 	_truncateFileName(fileName) {
 		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
-		const maxFileLength = 25;
+		const maxFileLength = 50;
 		if (fileName.length <= maxFileLength) {
 			return fileName;
 		}
@@ -138,7 +144,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 
 	render() {
 		return html`
-			<select class="d2l-input-select" @change=${this._onSelectChange}>
+			<select class="d2l-input-select d2l-truncate" @change=${this._onSelectChange}>
 				<option label=${this.localize('userSubmissions')} value=${submissions} ?selected=${this.selectedItemName === submissions}></option>
 				${this._files && this._files.map(submission => html`
 					<optgroup label=${this.localize('submissionNumber', 'number', submission.submissionNumber)}>
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -134,12 +134,13 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 	}
 
 	_truncateFileName(fileName) {
-		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
 		const maxFileLength = 50;
 		if (fileName.length <= maxFileLength) {
 			return fileName;
 		}
-		return  `${fileName.substring(0, maxFileLength)  }...${  ext}`;
+
+		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
+		return  `${fileName.substring(0, maxFileLength)}…${ext}`;
 	}
 
 	render() {
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -127,6 +127,15 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		this.dispatchEvent(event);
 	}
 
+	_truncateFileName(fileName) {
+		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
+		const maxFileLength = 25;
+		if (fileName.length <= maxFileLength) {
+			return fileName;
+		}
+		return  `${fileName.substring(0, maxFileLength)  }...${  ext}`;
+	}
+
 	render() {
 		return html`
 			<select class="d2l-input-select" @change=${this._onSelectChange}>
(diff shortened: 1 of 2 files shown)


components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -127,14 +127,23 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		this.dispatchEvent(event);
 	}
 
+	_truncateFileName(fileName) {
+		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
+		const maxFileLength = 25;
```
jeroach, line 132: I assume Dan confirmed 25 characters as the max desired length?

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -127,14 +127,23 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
+
 	render() {
 		return html`
 			<select class="d2l-input-select" @change=${this._onSelectChange}>
 				<option label=${this.localize('userSubmissions')} value=${submissions} ?selected=${this.selectedItemName === submissions}></option>
 				${this._files && this._files.map(submission => html`
 					<optgroup label=${this.localize('submissionNumber', 'number', submission.submissionNumber)}>
 						${this.getSubmissionFiles(submission).map(sf => html`
-							<option value=${JSON.stringify(sf)} label=${sf.name} ?selected=${sf.name === this.selectedItemName} class="select-option"></option>
+							<option value=${JSON.stringify(sf)} label=${this._truncateFileName(sf.name)} ?selected=${sf.name === this.selectedItemName} class="select-option"></option>
```
jeroach, line 146: More of a Dan question, but did we want a tooltip (or similar) that shows the full name?

jeroach (commented):
A few questions, otherwise looks good

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -127,14 +127,23 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		this.dispatchEvent(event);
 	}
 
+	_truncateFileName(fileName) {
+		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
+		const maxFileLength = 25;
+		if (fileName.length <= maxFileLength) {
```
AlexBedley, line 139: This should be at the top of the function

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -127,14 +127,23 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
+			return fileName;
+		}
+		return  `${fileName.substring(0, maxFileLength)  }...${  ext}`;
```
AlexBedley, line 142: I think if the file has no extension this will just print the filename twice in a row

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -127,14 +133,23 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		this.dispatchEvent(event);
 	}
 
+	_truncateFileName(fileName) {
+		const ext = fileName.substring(fileName.lastIndexOf('.') + 1, fileName.length);
```
AlexBedley, line 137: What if the extension is really long? or something like `.tar.gz`? are we relying on some whitelist further up the chain to filter out unreasonable file types?

JamieTran:
SC Update: selected item doesnt contain file extention
image

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: Dan mentioned that we can revisit this later and we don't need to worry about that at this point

components/header/d2l-consistent-evaluation-lcb-file-context.js
AlexBedley: nit: spacing inside the template expressions

nit: should we use an actual ellipsis (`…`) instead of 3 dots (`...`)?

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: Got his approval for 50 but had to add `.d2l-truncation` to the input component

vix633: We should have a pass with dan about this, I talked with him about the bar length and it was a bit longer than what you have in the screenshot. My talk with Dan last Tuesday, he suggested the bar is as long as it needs to be to fit file file name.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: In the instance of `.tar.gz`, only the `.gz` is displayed which I think should be fine as there are no other file extensions that are similar to it. 

As for unreasonable file types, it looks like if the instructor chooses submission type `file Submission` anything goes, even files with fake extensions.
Eg. 
image
Is this something that we'll have to handle in this component?

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: TIL there is an actual `…`, students are not able to submit a file with no extension

AlexBedley: > correct me if I'm wrong but students are not able to submit a file with no extension

I'm not sure. Dan or Megan might know. It seems reasonable to me to expect an extension however, it's pretty cheap to check on our end and just display the truncated file name.

components/header/d2l-consistent-evaluation-lcb-file-context.js
AlexBedley: > Is this something that we'll have to handle in this component?

I'm not sure, I'd try Dan or Megan or even @amauryperez

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: I've been keeping Dan up-to-date today with the screenshots and he said the current iteration of the component (SC below) is the better option as "It's not too important to know the filetype of the file you are already viewing"

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: > As for unreasonable file types, it looks like if the instructor chooses submission type file Submission anything goes, even files with fake extensions.

I was wrong about this most invalid filetypes are caught here
image

But for the ones that do pass through: here's a quote from dan

> Hmm. I'm not too worried about this. Probably an edge case that we can ignore so long as it's failing gracefully (which it appears to be). In general it's kind of users' faults if they upload really messed up files.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: You cannot upload a file with no extension

AlexBedley (approved):

jeroach (approved):

PR #106, opened 2020-10-08
Us118894/grade timing

AlexBedley:
**Problem:** We were noticing weird timing issues when actions were executed *quickly*. Some examples:

* quickly edit feedback text and save draft
* quickly edit, save draft, and iterate

The reason is because each button press fires an async action that takes a certain amount of time. We need these actions to happen in the same order they were executed in.

**Example:**

In this example, let's perform the following actions:

* 1st action - edit feedback text from "good job" to "needs improvement"
* 2nd action - before the debouncer triggers the transient save (so under 800ms) click save draft button
* 3rd action - before the draft is saved (during the lifetime of the API request, say <200ms), iterate to the next student

Before this PR the following could happen:

1. 2nd action finishes first, we save draft with "good job"
2. 3rd action finishes, UI loads the next student
2. 1st action finishes, we transiently save "needs improvement" but for the wrong student

**Solution:** There are 2 root problems we need to address:

1. debouncer - if you did another action (such as save draft) before the debouncer triggered, the feedback would save with a previous value (but still display the new value). To address this, we need to flush the debouncer when a save is triggered.

2. save order - even if the debouncer is flushed, the transient save could still take longer than the saving of the draft. We need to ensure things happen in the correct order. To address this, I added a mutex to protect certain sections of code. The sections that are protected:

* transient saves
* save/publish
* next/previous student

**Example continued:**

Ok so now we're flushing the debouncers and using the mutex to ensure ordering. Referring to our example above, now the 2nd and 3rd actions don't occur until the 1st action has completed. In a similar fashion, the 3rd action isn't started until the 2nd has completed. Thus we will now correctly save "needs improvement" for the first student and then load the evaluation for the next student.

Sorry for the super huge PR, I'll try and give a good walkthrough


components/awaiter.js
```diff
@@ -0,0 +1,26 @@
+export class Awaiter {
```
AlexBedley, line 1: This class is a bit complicated. I'd focus more on the interface to start, and loop back to reading this later. I plan on extending the unit test coverage so we feel more confident about this implementation.

`lock` is public but it's only used internally. `dispatch` simply takes a function and ensures it runs to completion before the next function you called `dispatch` with.

components/consistent-evaluation-page.js
```diff
@@ -297,18 +299,22 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		return this.evaluationEntity.properties.state === publishedState;
 	}
 
-	_onNextStudentClick() {
-		this.dispatchEvent(new CustomEvent('d2l-consistent-evaluation-next-student-click', {
-			composed: true,
-			bubbles: true
-		}));
+	async _onNextStudentClick() {
```
AlexBedley, line 302: For the next, previous, and transient save functions, I simply wrapped them in a dispatch

components/consistent-evaluation-page.js
```diff
@@ -320,73 +326,118 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
+				}
+				else if (type === GradeType.Number) {
+					newGradeVal = e.detail.grade.score;
+				}
+				this.evaluationEntity = await this._controller.transientSaveGrade(entity, newGradeVal);
+			}
+		);
 	}
 
 	async _saveEvaluation() {
```
AlexBedley, line 356: The save/update/publish/retract functions are also wrapped in a dispatch. However they also must signal to their children to flush their debouncer queues (otherwise for example the save will happen with old values - remember the debouncer batches changes in 800ms increments)

components/consistent-evaluation-page.js
```diff
@@ -432,8 +483,8 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 	}
 
 	_renderToast() {
-		return html`<d2l-alert-toast 
-			?open=${this._displayToast} 
+		return html`<d2l-alert-toast
```
AlexBedley, line 486: the rest of this file is just atom being nitpicky about whitespace. this made the PR by accident

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -55,10 +55,31 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
...
 		this.canEditFeedback = false;
 		this._debounceJobs = {};
-		this.addEventListener('d2l-request-provider',
-			e => {
-				if (e.detail.key === 'd2l-provider-html-editor-enabled') e.detail.provider = true;
-			});
+		this.flush = this.flush.bind(this);
+	}
+
+	htmlEditorEnabled(e) {
```
AlexBedley, line 61: I moved this from the lambda so it could be correctly garbage-collected in `disconnectedCallback`

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -55,10 +55,31 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
...
+		window.addEventListener('d2l-flush', this.flush);
+	}
+
+	disconnectedCallback() {
+		this.removeEventListener('d2l-request-provider', this.htmlEditorEnabled);
+		window.removeEventListener('d2l-flush', this.flush);
+		super.disconnectedCallback();
+	}
+
+	flush() {
```
AlexBedley, line 79: This class uses a debouncer so it is interested in `d2l-flush` events. It just simply calls the debouncer's built-in `flush` function to tell the debouncer to not wait 800ms and transiently save their data immediately.

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -66,8 +71,32 @@ export class ConsistentEvaluationGradeResult extends LocalizeMixin(LitElement) {
...
+
+	flush() {
+		if (this._debounceJobs.grade && this._debounceJobs.grade.isActive()) {
+			this._debounceJobs.grade.flush();
+		}
+	}
+
 	onGradeChanged(e) {
 		const score = e.detail.value;
+		this._debounceJobs.grade = Debouncer.debounce(
```
AlexBedley, line 92: Grades did not already use a debouncer like feedback, so this diff includes adding the debouncer stuff, along with the same things we did in feedback-presentational

test/awaiter.test.js
```diff
@@ -0,0 +1,38 @@
+import { assert } from '@open-wc/testing';
```
AlexBedley, line 1: I plan on extending the test coverage tomorrow morning to include `dispatch` and make it a bit simpler (by deleting MyObject)

jeroach (approved):
Thanks for the detailed notes! Looks good to me.

I'll do a check of the updated unit tests once ready.

jmurray-d2l (approved):
Nice work!  Tough problem to solve

jmurray-d2l (approved):

PR #107, opened 2020-10-09
add bottom border to lcb

jeroach:
[US121323](https://rally1.rallydev.com/#/detail/userstory/441082552956?fdp=true): [LCB] Bottom border

image

### DIFF
--- a/components/header/d2l-consistent-evaluation-learner-context-bar.js
+++ b/components/header/d2l-consistent-evaluation-learner-context-bar.js
@@ -1,5 +1,6 @@
 import './d2l-consistent-evaluation-lcb-user-context.js';
 import './d2l-consistent-evaluation-lcb-file-context.js';
+import '@brightspace-ui/core/components/colors/colors.js';
 import { css, html, LitElement } from 'lit-element';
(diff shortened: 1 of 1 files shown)


components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -27,33 +28,34 @@ export class ConsistentEvaluationLearnerContextBar extends RtlMixin(LitElement)
 	static get styles() {
 		return css`
 			:host {
+				border-bottom: 0.05rem solid var(--d2l-color-gypsum);
 				display: flex;
 				height: 100%;
-				margin: 0.75rem 0 0.75rem 1.5rem;
+				padding: 0.75rem 0 0.75rem 1.5rem;
```
jeroach, line 34: Changed all the `margin` to `padding` since padding is inside the border, where as margin is outside the border.

We want all of the correct spacing between the border and the profile image/submission list.

vix633 (approved):

AlexBedley (approved):
looks good. I dunno what d2l's stance is on css logical properties, but we might be able to avoid all the rtl selectors if we use them. ie https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start

for another PR though

PR #108, opened 2020-10-13
Abedley/feedback/cumulative state

vix633:
This makes the feedback actions component work again.


components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,150 @@
...
+				</li>`;
+		});
+
+		return html`
+			<d2l-labs-attachment-list
+				editing="true"
+				@d2l-attachment-removed="${this.removeAttachment}">
+				${a}
+			</d2l-labs-attachment-list>
+			<div ?hidden="${!this.canEditFeedback}">
```
vix633, line 138: The default hidden behaviour on the picker still shows a grey bar without buttons, I talked with Dan and we prefer hiding it altogether. 

image

jmurray-d2l: This was previously using this https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/AttachmentCollectionEntity.js#L34 to determine read-only state.  Using the existence of those actions is probably more correct than using `!this.canEditFeedback`

vix633: This actually does both, the second part is here. https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/108/files#diff-e9a2b8eca2a2fd97cf0739278ae72f9f78219b7b2e25e403b879f432b6bd013cR91

jmurray-d2l: ahh cool, so you shouldn't need this then?

jmurray-d2l: Okay, right now though its possible the attach buttons are visible but the attach actions are not present on the api response and the user will get an error when attempting to attach something.  If we use the presence of the attach actions to toggle the visibility of the buttons, we wouldn't have that problem

jmurray-d2l: We'll be fine for now as is, it would be more for reliability later if we were to decouple permissions, so up to you, just letting you know how it originally worked

vix633: Good point! I just checked via api browser and it does look like we get a collection with no actions when we don't have permissions to give feedback.

```
{
  "class": [
    "collection",
    "attachments"
  ],
  "links": [
    {
      "rel": [
        "https://files.api.brightspace.com/rels/files"
      ],
      "href": "https://snip.files.api.proddev.d2l/6609"
    }
  ]
}
```
So yeah, it would totally be viable to do the check via `getActionByName`

jmurray-d2l: yeah, we don't do it for grades or feedback, and we probably should at some point.  So up to you, but the `AttachmentCollectionEntity` has it built in already, as you pointed out, so up to you!

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -128,15 +105,12 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
...
+							href="${this.href}/attachments"
 							.token="${this.token}"
-							@d2l-activity-attachments-picker-files-uploaded="${this.saveAttachment}"
-							@d2l-activity-attachments-picker-video-uploaded="${this.saveAttachment}"
-							@d2l-activity-attachments-picker-audio-uploaded="${this.saveAttachment}"
-							@d2l-attachment-removed="${this.deleteAttachment}"
-							?disabled="${!this.canEditFeedback}">
-						</d2l-activity-attachments-editor>
+							destinationHref="${this.href}"
+							.canEditFeedback="${this.canEditFeedback}">
```
AlexBedley, line 112: nit: since you've linked it to an attribute already, we can use it like `?can-edit-feedback="${this.canEditFeedback}"`

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -128,15 +105,12 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 						ariaLabel="${this.localize('overallFeedback')}">
 					</d2l-activity-text-editor>
 					<div>
-						<d2l-activity-attachments-editor
-							.href="${this.href}/attachments"
+						<d2l-consistent-evaluation-attachments-editor
+							href="${this.href}/attachments"
```
AlexBedley, line 109: instead of url hacking we should follow the rel to find this href

components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,150 @@
...
+					?deleted="${deleted}"
+					?creating="${creating}"
+					?editing="${a.canDeleteAttachment()}">
+				</d2l-labs-attachment>
+				</li>`;
+		});
+
+		return html`
+			<d2l-labs-attachment-list
+				editing="true"
```
AlexBedley, line 134: will this always be true? what if you don't have permission?

components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,150 @@
...
+				url: a.href()
+			};
+			const deleted = false;
+			const creating = false;
+			return html`<li slot="attachment" class="panel">
+				<d2l-labs-attachment
+					attachmentId="${a.self()}"
+					.attachment="${attachment}"
+					?deleted="${deleted}"
+					?creating="${creating}"
```
AlexBedley, line 126: I just put these `false`s as placeholders, they need to be replaced with the actual values

components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,150 @@
...
+		const a = await window.D2L.Siren.EntityStore.fetch(e.detail, this.token);
+		const attachmentEntity = new AttachmentEntity(a.entity, this.token);
+		await attachmentEntity.deleteAttachment();
+
+		const evaluationEntity = await window.D2L.Siren.EntityStore.get(this.destinationHref, this.token);
+		this.href = evaluationEntity.getLinkByRel(Rels.Activities.feedbackAttachments).href;
+	}
+
+	render() {
+		const a = this.attachments.map(a => {
```
AlexBedley, line 113: nit: I know I named them `a` but it was supposed to be temporary. It's a bit confusing here since there are 2 `a`'s

components/right-panel/consistent-evaluation-attachments-editor.js
vix633: Good question. I'm going to see if this controls the delete button and change it.

components/right-panel/consistent-evaluation-attachments-editor.js
vix633: i'll rename

components/right-panel/consistent-evaluation-feedback-presentational.js
vix633: I'll squeak this in my next pr

components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+
+	async _init(href, token) {
+		const tempAttachments = [];
+		let promises = [];
+		const entity = await window.D2L.Siren.EntityStore.fetch(href, token);
+		if (entity && entity.entity && entity.entity.entities) {
+			const entities = entity.entity.entities;
+			promises = entities.map(attachmentEntity => {
```
AlexBedley, line 73: normally I'm a huge fan of using the built-in array functions, but with Promises it just makes it more confusing. I'd just use a for-loop in this case, then you can get rid of `promises` and `vals` and `Promise.all`

components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,152 @@
...
+				return window.D2L.Siren.EntityStore.fetch(attachmentEntity.href, token);
+			});
+		}
+		const vals = await Promise.all(promises);
+		vals.map(a => {
```
AlexBedley, line 78: If you do stick with this approach, this should be `forEach`

components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -0,0 +1,152 @@
...
+		const a = await window.D2L.Siren.EntityStore.fetch(e.detail, this.token);
+		const attachmentEntity = new AttachmentEntity(a.entity, this.token);
+		await attachmentEntity.deleteAttachment();
+
+		const evaluationEntity = await window.D2L.Siren.EntityStore.get(this.destinationHref, this.token);
+		this.href = evaluationEntity.getLinkByRel(Rels.Activities.feedbackAttachments).href;
+	}
+
+	render() {
+		console.log('render', this._attachments.length);
```
AlexBedley, line 117: I'm surprised this passed linting

components/right-panel/consistent-evaluation-feedback-presentational.js
AlexBedley: if you don't wanna do this, then you should delete the unused attribute

components/right-panel/consistent-evaluation-attachments-editor.js
vix633: I definitely want to keep this. Promise.All saves us many renders. What I should also do is use `updated()` rather than the setter and getter for token and href. Not in this PR tho.

dbatiste: Yeah the `Promise.all` was something I'd suggested to Viktor to parallelize the fetches for the attachment entities.  Previous to that, they were waiting in succession - something we should generally try to avoid unless the calls depend on one another.

AlexBedley (approved):

PR #109, opened 2020-10-13
Handle late submission in LCB

JamieTran:
Added a subtle button to indicate how late the submission is, the button will open a special-access dialog for that particular use
late-submission-special-access

### DIFF
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -98,7 +98,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 			if (submission.entity.getSubEntityByClass(Classes.assignments.submissionDate)) {
 				sf.properties.latenessTimespan = submission.entity.properties.lateTimeSpan;
 			}
-			// this is when the assignment was submitted
+
 			sf.properties.date = submission.entity.getSubEntityByClass(Classes.assignments.submissionDate).properties.date;
 			sf.properties.displayNumber = displayNum;
 			return sf.properties;
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -74,12 +74,6 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 	constructor() {
 		super();
 		/* global moment:false */
-		moment.relativeTimeThreshold('s', 60);
-		moment.relativeTimeThreshold('m', 60);
-		moment.relativeTimeThreshold('h', 24);
-		moment.relativeTimeThreshold('d', Number.MAX_SAFE_INTEGER);
-		moment.relativeTimeRounding(Math.floor);
-
 		this._showFiles = false;
 	}
 
@@ -144,11 +138,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 			this._dispatchRenderEvidenceTextEvent(submission);
 		}
 
-		if (submission.latenessTimespan !== undefined) {
-			this._submissionLateness = submission.latenessTimespan;
-		} else {
-			this._submissionLateness = undefined;
-		}
+		this._submissionLateness = submission.latenessTimespan;
 	}
 
 	_dispatchRenderEvidenceFileEvent(sf) {
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -177,6 +177,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		} else {
 			return html`
 				<d2l-button-subtle
+					id="lateButton"
 					text="${moment.duration(Number(this._submissionLateness), 'seconds').humanize()} ${this.localize('late')}"
 					icon="tier1:access-special"
 					@click="${this._openSpecialAccessDialog}"
@@ -211,7 +212,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		];
 
 		const delayedResult = D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
-			/*               opener: */ document.body,
+			/*               opener: */ this.shadowRoot.querySelector('d2l-button-subtle'),
 			/*             location: */ location,
 			/*          srcCallback: */ 'SrcCallback',
 			/*       resizeCallback: */ '',
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -3,7 +3,6 @@ import '@brightspace-ui/core/components/button/button-subtle.js';
 import { attachmentListRel, submissions } from '../controllers/constants';
 import { css, html, LitElement } from 'lit-element';
 import { Classes } from 'd2l-hypermedia-constants';
-import { labelStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { loadLocalizationResources } from '../locale.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
@@ -37,7 +36,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 	}
 
 	static get styles() {
-		return [labelStyles, selectStyles,
+		return [selectStyles,
 			css`
 				:host {
 					margin-left: 0.7rem;
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -4,7 +4,6 @@ import '@brightspace-ui/core/components/button/button-subtle.js';
 import { attachmentListRel, submissions } from '../controllers/constants';
 import { css, html, LitElement } from 'lit-element';
 import { Classes } from 'd2l-hypermedia-constants';
-import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
 import { loadLocalizationResources } from '../locale.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
@@ -176,10 +175,8 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		{
 			return html``;
 		} else {
-			const uniqueId = getUniqueId();
 			return html`
 				<d2l-button-subtle
-					id="${uniqueId}"
 					text="${moment.duration(Number(this._submissionLateness), 'seconds').humanize()} ${this.localize('late')}"
 					icon="tier1:access-special"
 					@click="${this._openSpecialAccessDialog}"
--- a/components/header/d2l-consistent-evaluation-lcb-file-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-file-context.js
@@ -1,3 +1,4 @@
+/* global moment:false */
 import 'd2l-polymer-siren-behaviors/store/entity-store.js';
 import '@brightspace-ui/core/components/button/button-subtle.js';
 import { attachmentListRel, submissions } from '../controllers/constants';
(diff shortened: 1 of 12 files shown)


components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -127,6 +159,67 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
+			return html`
+				<d2l-button-subtle
+					text="${moment.duration(Number(this._submissionLateness), 'seconds').humanize()} ${this.localize('late')}"
+					icon="tier1:access-special"
+					@click="${this._openSpecialAccessDialog}"
+				></d2l-button-subtle>`;
+		}
+	}
+
+	_openSpecialAccessDialog() {
```
JamieTran, line 175: I didn't know how to get access to this method from [activities](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-special-access-editor.js#L65) so I copied the method into here. I'm not sure if this is the right way to do it.

vix633: This is probably better anyway. The alternative would have been making the method public and using the query selector to call it.

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -64,6 +73,12 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 
 	constructor() {
 		super();
+		/* global moment:false */
+		moment.relativeTimeThreshold('s', 60);
```
AlexBedley, line 77: if i'm understanding this correctly, this `moment` stuff is already set up in submissions page. it seems like the correct solution would be to set this stuff at a top level, rather than setting it multiple times across the components that will use it.

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/4837a33aba72e9456039df4d391bff2321559083/components/left-panel/consistent-evaluation-submissions-page.js#L47-L51

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -121,6 +143,12 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		} else {
 			this._dispatchRenderEvidenceTextEvent(submission);
 		}
+
+		if (submission.latenessTimespan !== undefined) {
```
AlexBedley, line 147: I think we can simplify this to just `this._submissionLateness = submission.latenessTimespan;`

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -152,6 +180,67 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
+		];
+
+		const delayedResult = D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
+			/*               opener: */ document.body,
+			/*             location: */ location,
+			/*          srcCallback: */ 'SrcCallback',
+			/*       resizeCallback: */ '',
+			/*      responseDataKey: */ 'result',
+			/*                width: */ 1920,
+			/*               height: */ 1080,
```
AlexBedley, line 230: do we want these to be static? what if this is bigger than the viewport?

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -152,6 +180,67 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
+			/*            closeText: */ this.localize('closeBtn'),
+			/*              buttons: */ buttons,
+			/* forceTriggerOnCancel: */ false
+		);
+
+		// "X" abort handler
+		// refetch special access in case the user count has changed
+		delayedResult.AddReleaseListener(() => specialAccess);
```
AlexBedley, line 238: I don't understand this, could you please elaborate?

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -99,6 +116,11 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 			if (submission.entity.getSubEntityByClass(Classes.assignments.submissionComment)) {
 				sf.properties.comment = submission.entity.getSubEntityByClass(Classes.assignments.submissionComment).properties.html;
 			}
+
+			if (submission.entity.getSubEntityByClass(Classes.assignments.submissionDate)) {
+				sf.properties.latenessTimespan = submission.entity.properties.lateTimeSpan;
```
AlexBedley, line 121: I know this is just copying the current pattern but I find it a bit strange that we're editing `sf.properties` here. This will cause this component's cached version of the entity (contained in _`files` I believe) to differ from the actual API or EntityStore cache. This could cause huge headaches down the road. I suggest just creating the object we want from a new object. This also prevents any surprises if we change the API.

kinda unrelated: what is going on with `displayNum` in this function? isn't it always 0?

AlexBedley: oh wow reading further it looks like we are relying on this behaviour? I previously warned us against caching things like this. it hasn't burned us yet I don't think but I do worry about files like this one that do their own fetching and data modification. everything is way easier to reason about if we just work off of the API. I wouldn't tackle this problem in this PR, but if we get some time after M1 I would love to revisit here.

components/header/d2l-consistent-evaluation-lcb-file-context.js
AllanKerr: Ideally everywhere that needs to use special access uses a single component so it'd be awesome to reuse the `d2l-activity-special-access-editor` component here and just pass in the `href`. What are your thoughts on updating `d2l-activity-special-access-editor` to take a `lateness` attribute that determines what text to render and whether to render an icon?

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: To my understanding, this is the ratio to which the dialog will open, not a static width and height. The dialog cannot extend past the viewport.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: Agreed, I will move it to consistent-evaluation-page.js

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -152,6 +180,67 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
+			{
+				Text: this.localize('cancelBtn'),
+				ResponseType: 2, // D2L.Dialog.ResponseType.Negative
+				IsPrimary: false,
+				IsEnabled: true
+			}
+		];
+
+		const delayedResult = D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
+			/*               opener: */ document.body,
```
dbatiste, line 224: Are you sure you want `docunent.body` as the opener?  This is where focus will be placed after the dialog closes?

components/header/d2l-consistent-evaluation-lcb-file-context.js
dbatiste: I don't recall this being a ratio.  Is it necessary to specify height and width - i.e. if no dimensions are specified, will it try to auto-size itself?

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: That is a good question, I'm not sure what it does so, I reached out to the developer who implemented it in [activities](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-special-access-editor.js#L103). The behavior doesn't change when I remove it so it might be a precautionary thing, but I will wait for a response.

dbatiste: I think the handler provided to the `AddReleaseListener` method is invoked when the user aborts the dialog (i.e. closes via `[x]` button in top right corner, or in theory hitting `Escape` key.  If the caller is returning another LMS MVC `Event` or a `Promise`, they would typically need to resolve them by either calling `Release` (on the `Event` object), or `reject`/`resolve` for the `Promise`.  Otherwise, that consumer of that result could be waiting indefinitely.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: Oh, you're right, it is not a ratio. I thought it was because changing it to `height & width = 1` gave me this 
image
If no dimensions are specified, the component fails on trying to install the buttons with a `length` of `0`. It does look like it auto-sizes to make sure it does not go off the screen.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: Ahh, so that's what the opener is. It would make sense for the focus to be on the `d2l-button-subtle` after the dialog is closed.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: I totally agree as I was trying initially to use the component but couldn't get it working. I am not opposed to the idea but don't think it should be encapsulated in this PR.

components/header/d2l-consistent-evaluation-lcb-file-context.js
dbatiste: Cool yeah.  The logic for opener is fairly flexible.  In MVC this is usually an `IHtmlId`, but the JS method can accept an element as well.  If you have or know to get a reference to the opener, that should work well.  In cases where the opener is not necessarily known, it would typically be appropriate to return the focus to the element that had focus upon opening (usually the opener), which could be determined using the `getComposedActiveElement` from `core`.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: the reason for having both `delayedResult.AddReleaseListener` and `delayedResult.AddListener` is because when the dialog is closed by the save or cancel button, it triggers the listener while closing the dialog via the dialog top right 'x' only triggers the release listener, so we had to handle both cases

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -64,7 +73,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 
 	constructor() {
 		super();
-
+		/* global moment:false */
```
AlexBedley, line 76: forgot to delete this?

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -152,6 +170,68 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
 	}
 
+	_renderLateButton() {
+		if (this.selectedItemName === submissions || !this._submissionLateness)
+		{
+			return html``;
+		} else {
+			return html`
+				<d2l-button-subtle
+					id="lateButton"
```
AlexBedley, line 180: we normally try and make id's more unique. TBH I forget exactly why, something to do with inadequate browser support for shadowdom or something. but at the very least we prefix them with the component name. There's also a core library which will help you generate actually unique IDs.

https://github.com/BrightspaceUI/core/blob/e0d1123612c4929f037bf49fe00b492293b79889/helpers/uniqueId.js

components/header/d2l-consistent-evaluation-lcb-file-context.js
AlexBedley: In this case then, `specialAccess` is not a promise or anything so I don't think we need it

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -44,11 +44,6 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 		this._token = undefined;
 		this._submissionEntities = [];
 		/* global moment:false */
```
AlexBedley, line 46: forgot to delete?

components/header/d2l-consistent-evaluation-lcb-file-context.js
AllanKerr: IDs need to be unique because Legacy Edge still relies on ShadyDOM to polyfill web component support. Because of this, we don't get true shadow roots and need to namespace IDs. If its a component that will be used multiple times on a page you'll want to use a unique ID. If its only used once you should be able to get away with namespacing it using the component name.

components/header/d2l-consistent-evaluation-lcb-file-context.js
JamieTran: It is here for linting, without it when `moment.duration()` is called it says `'moment' is not defined`. Is there another way of doing this?

AlexBedley: ah maybe just moving it to where `moment` is used would clear it up. This is an eslint directive right? there are ways to disable it for a specific line 

https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments

AlexBedley (approved):

PR #110, opened 2020-10-14
US121085 - Evidence - Iterating when evidence is viewed the submissions are not changed

amauryperez:


components/consistent-evaluation.js
```diff
@@ -68,6 +68,7 @@ export class ConsistentEvaluation extends MobxLitElement {
 			this._userName = await controller.getUserName();
 			this._iteratorTotal = await controller.getIteratorInfo('total');
 			this._iteratorIndex = await controller.getIteratorInfo('index');
+			this.shadowRoot.querySelector('d2l-consistent-evaluation-page')._resetEvidence();
```
AlexBedley, line 71: I'm worried that setting `this.submissionInfo = undefined;` here will override the value set on line 64 in this file. but we might be ok if lit waits for `updated` to finish before passing props to its children

AlexBedley (approved):

components/consistent-evaluation.js
amauryperez: will moving the function call to before line 64 help?

PR #111, opened 2020-10-14
US120390 - Submissions - Remove menu item for Edit a Copy

amauryperez:


components/left-panel/consistent-evaluation-evidence-text.js
```diff
@@ -49,7 +49,6 @@ export class ConsistentEvaluationEvidenceText extends LocalizeMixin(LitElement)
 					<d2l-dropdown-menu id="dropdown" boundary="{&quot;right&quot;:10}">
 						<d2l-menu>
 							<d2l-menu-item-link text="${this.localize('download')}" href="${this.downloadUrl}"></d2l-menu-item-link>
-							<d2l-menu-item-link text="${this.localize('editCopy')}" href="#"></d2l-menu-item-link>
```
AlexBedley, line 52: I wonder if we should delete the langterm. hopefully whoever puts it back notices the langterms are there

AlexBedley (approved):

components/left-panel/consistent-evaluation-evidence-text.js
amauryperez: I though about it, but was not sure if we had sent them for translations already so to be safe I left them

PR #113, opened 2020-10-15
US119139 - Submissions - Wire up menu items for Mark as Read/Unread and Flag

amauryperez:


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -369,25 +394,26 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+				<d2l-menu label="${this.localize('moreOptions')}">
+					${this.submissionType === textSubmission ? html`
 						<d2l-menu-item-link text="${this.localize('viewFullSubmission')}"
 							href="javascript:void(0);"
-							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null }
+							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null}
 					<d2l-menu-item-link text="${this.localize('download')}" href="${downloadHref}"></d2l-menu-item-link>
-					<d2l-menu-item-link text="${oppositeReadState}" href="#"></d2l-menu-item-link>
-					<d2l-menu-item-link text="${oppositeFlagState}" href="#"></d2l-menu-item-link>
+					<d2l-menu-item text="${oppositeReadState}" href="javascript:void(0);" @click="${toggleIsReadStatusOnClick}"></d2l-menu-item>
```
AlexBedley, line 411: I'm surprised this is necessary, according to the docs we can listen for `d2l-menu-item-select` events

https://github.com/BrightspaceUI/core/blob/e0d1123612c4929f037bf49fe00b492293b79889/components/menu/README.md#d2l-menu-item

Ideally we don't have an `href` attribute which does nothing and we don't have to bind the click listeners in the constructor

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -369,25 +394,26 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 		});
 	}
 
-	_addMenuOptions(read, flagged, downloadHref) {
-		// Placeholder for menu presentational
+	_addMenuOptions(read, flagged, downloadHref, attachmentEntity) {
 		const oppositeReadState = read ? this.localize('markUnread') : this.localize('markRead');
+		const toggleIsReadStatusOnClick = () => this._dispatchToggleFileIsReadStatusEvent(attachmentEntity);
 		const oppositeFlagState = flagged ? this.localize('unflag') : this.localize('flag');
+		const toggleFlagStatusOnClick = () => this._dispatchToggleFileFlagStatusEvent(attachmentEntity);
```
AlexBedley, line 401: I can see why these lambdas exist but it's another design smell.

Ideally we just fire the event with some kind of unique ID (`attachmentId`?) Then the controller (in this case, `consistent-evaluation-submissions-page`), can find the entity inside of the list of `_submissionEntities` and update it accordingly. This would also solve the perf problem we have with reloading all of the attachments whenever you mark something as read or flagged.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -91,7 +93,8 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 	}
 
 	async _getSubmissionEntity(submissionHref) {
-		return await window.D2L.Siren.EntityStore.fetch(submissionHref, this._token, false);
+		const byPassCache = true;
+		return await window.D2L.Siren.EntityStore.fetch(submissionHref, this._token, byPassCache);
```
AlexBedley, line 97: just for awareness's sake, now that we're bypassing the cache each time, we have to load the entire submissions list for every change, which makes it easier to be correct but can be expensive with regards to load time

AlexBedley (approved):

components/left-panel/consistent-evaluation-submission-item.js
amauryperez: I'll remove the href, I put it because I saw it on line 382

amauryperez: I'll see if I can find and example of `d2l-menu-item-select` somewhere

dbatiste: Alex is spot on here.  For `d2l-menu-item`, the "action" is JavaScript code, and the way to hook in is using `d2l-menu-item-select` event on the item or one of the ancestors (since the event will bubble).  The `href` only applies to `d2l-menu-item-link`, which are akin to `<a>` elements and perform some navigation.

amauryperez: the documentation is not very detailed on what to do with d2l-menu-item-select

amauryperez: @dbatiste how do I send then event directly to the parent (consistent-evaluation-submissions-page.js) from d2l-menu-item-select?

dbatiste: @amauryperez : thanks for pointing that out Amaury - yeah, we could add an example that shows this a bit better.

The `d2l-menu-item-select` event will be automatically dispatched by the menu-item component.  The consumer code just needs to wire up a listener for it.  This is how the wire-up is done in Lit-elements (using the `@` directive).
``` html
<d2l-menu label="...">
    <d2l-menu-item @d2l-menu-item-select="${this._onFirstMenuItemSelect}"></d2l-menu-item>
    <d2l-menu-item @d2l-menu-item-select="${this._onSecondMenuItemSelect}"></d2l-menu-item>
    ...
</d2l-menu>
```
or
``` html
<d2l-menu label="..." @d2l-menu-item-select="${this._onMenuItemSelect}">
    <d2l-menu-item data-key="action1"></d2l-menu-item>
    ...
</d2l-menu>
```
The event can be wired up on any ancestor of the item.  Obviously, if using a single event handler for more than one menu-item, the handler would need to check the target (`e.target` where `e` is automatically passed to all event handlers) in order to know what action to perform.  The single handler approach results in a more complex handler but may be useful in cases where the menu-items are not known until run-time.  If the menu-items are known at dev-time, I would likely create a handler for each item to avoid any conditional logic in such a single handler.  Nevertheless, the target would be checked such as...
```
_onMenuItemSelect(e) {
    if (e.target.getAttribute('data-key') === 'action1') {
        // do stuff for menu item with some-key="action1"
    }
}
```


For wiring up in vanilla JS code... one would simply...
``` javascript
element.addEventListener('d2l-menu-item-select', (e) => {
    // do something
});
```
Does that help?

amauryperez: It helps In this case we need to know to which file entity to apply the menu action to, so for `data-key` will need to give it some file identifier during runtime and in the event listener loop through the entities and find it, the current implementation was easier because the whole entity its passed to the event

amauryperez: this means that `_onMenuItemSelect` still needs to send another event to the parent `consistent-evaluation-submissions-page.js` right?

dbatiste: It should be pretty easy to look up the file if necessary.  I'm not that familiar with what's being done at a higher level here, but yes, dispatching a new event that is specific to this component for higher consumers would be perfectly fine, and desirable, as to abstract away the fact that a menu-item was even clicked.  It should be possible to change the implementation (ex. to use something other than a d2l-menu) of this component without breaking consumers.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -369,25 +394,24 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+				<d2l-menu label="${this.localize('moreOptions')}">
+					${this.submissionType === textSubmission ? html`
 						<d2l-menu-item-link text="${this.localize('viewFullSubmission')}"
 							href="javascript:void(0);"
-							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null }
+							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null}
 					<d2l-menu-item-link text="${this.localize('download')}" href="${downloadHref}"></d2l-menu-item-link>
-					<d2l-menu-item-link text="${oppositeReadState}" href="#"></d2l-menu-item-link>
-					<d2l-menu-item-link text="${oppositeFlagState}" href="#"></d2l-menu-item-link>
+					<d2l-menu-item text="${oppositeReadState}" id="${name}" @d2l-menu-item-select="${this._dispatchToggleFileIsReadStatusEvent}"></d2l-menu-item>
```
amauryperez, line 409: Using the `name` property for now to validate the approach. The API will need to send the actual `fileId` to have a unique identifier

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -369,25 +394,24 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
...
+				<d2l-menu label="${this.localize('moreOptions')}">
+					${this.submissionType === textSubmission ? html`
 						<d2l-menu-item-link text="${this.localize('viewFullSubmission')}"
 							href="javascript:void(0);"
-							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null }
+							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null}
 					<d2l-menu-item-link text="${this.localize('download')}" href="${downloadHref}"></d2l-menu-item-link>
-					<d2l-menu-item-link text="${oppositeReadState}" href="#"></d2l-menu-item-link>
-					<d2l-menu-item-link text="${oppositeFlagState}" href="#"></d2l-menu-item-link>
+					<d2l-menu-item text="${oppositeReadState}" data-key="${id}" @d2l-menu-item-select="${this._dispatchToggleFileIsReadStatusEvent}"></d2l-menu-item>
```
amauryperez, line 409: using the fileId as unique identifier

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -109,6 +112,69 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
...
+		if (!attachmentEntity) {
+			throw new Error('Invalid entity provided for attachment');
+		}
+
+		const action = attachmentEntity.getActionByName(toggleIsReadActionName);
+		await performSirenAction(this.token, action, undefined, true);
+
+		attachmentEntity.properties.read = !attachmentEntity.properties.read;
+
+		this._updateAttachmentEntity(attachmentEntity) ;
```
amauryperez, line 157: is this better than calling `await this._initializeSubmissionEntities();`

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -109,6 +111,46 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
...
+			for (let y = 0;y < attachments.length; y++) {
+				if (attachments[y].properties.id === fileId) {
+					return attachments[y];
+				}
+			}
+		}
+		return null;
+	}
+
+	async _updateAttachmentEntity(submissionEntity, submissionSelfLinkHref) {
```
AlexBedley, line 126: nit: this probably should be `_updateSubmissionEntity`

AlexBedley (approved):

AlexBedley (approved):

PR #115, opened 2020-10-16
Cmitchell/us117788 coa page footer cleanup

Callum-Mitchell:
**What this does:**
- Implements demos for the save draft/publish/update/retract footer buttons to cover different states/permissions
- COA override page now has a demo for both published and unpublished states
- Changes coa override page demos + data to use an ActorActivityUsage entity instead of UserActivityUsage

These changes do not add additional component functionality; only local demos are affected.


Callum-Mitchell:
**Context for changes:** was working with @gfrancu on getting the COA eval override page working in the LMS; original goal was to add the page footer but it appears to already be in place and fully functional (correct me if I'm wrong). However, the LMS page is receiving an ActorActivityUsage entity from the API but the local demo was using UserActivityUsage - the local demo change is meant to reflect that fact. Other changes are simply suggestions found during investigation.

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -147,4 +147,4 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeMixin(LitElemen
 	}
 }
 
-customElements.define('d2l-consistent-evaluation-feedback-presentational', ConsistentEvaluationFeedbackPresentational);
+customElements.define('d2l-consistent-evaluation-feedback', ConsistentEvaluationFeedbackPresentational);
```
AlexBedley, line 150: we should keep the element name consistent with the class and file name

AlexBedley (commented):
Thanks for updating the demos.

Curious as to why we removed `presentational` from the names of the components. It's meant to denote that the components are simple - they just render their properties and emit events. In contrast to the non-presentational components that act more like controllers and do API things.

Callum-Mitchell:
> Curious as to why we removed `presentational` from the names of the components. It's meant to denote that the components are simple - they just render their properties and emit events. In contrast to the non-presentational components that act more like controllers and do API things.

Makes sense. In that case, I'll change the names back. I thought presentational might mean they were there for display purposes but not necessarily fully functional.

AlexBedley (approved):

PR #117, opened 2020-10-19
Jmurray/leave without saving dialog

jmurray-d2l:


components/header/consistent-evaluation-nav-bar.js
```diff
@@ -97,20 +93,20 @@ class ConsistentEvaluationNavBar extends LocalizeMixin(LitElement) {
 	}
 
 	_dispatchButtonClickEvent(eventName) {
-		this.dispatchEvent(new CustomEvent(eventName, {
+		console.log(eventName);
```
jmurray-d2l, line 96: ahhh cripes

components/consistent-evaluation-page.js
```diff
@@ -314,10 +310,12 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
-			() => this.dispatchEvent(new CustomEvent('d2l-consistent-evaluation-previous-student-click', {
-				composed: true,
-				bubbles: true
-			}))
+			() => {
+				this.dispatchEvent(new CustomEvent('d2l-consistent-evaluation-previous-student-click', {
+					composed: true,
+					bubbles: true
+				}));
+			}
```
AlexBedley, line 318: nit: unnecessary block

components/consistent-evaluation-page.js
```diff
@@ -460,37 +456,44 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
-		if (!this.confirmUnsavedChanges) return;
-		if (!this._hasUnsavedChanges && value) {
-			window.addEventListener('beforeunload', this._confirmUnsavedChangesBeforeUnload);
-		} else if (this._hasUnsavedChanges && !value) {
-			window.removeEventListener('beforeunload', this._confirmUnsavedChangesBeforeUnload);
+	_navigate() {
+		switch (this.navigationTarget) {
+			case 'back':
+				if (this.evaluationEntity.hasClass('unsaved')) {
+					window.removeEventListener('beforeunload', this.unsavedChangesHandler);
```
AlexBedley, line 479: my understanding is the disconnected callback should handle this for us

components/consistent-evaluation-page.js
```diff
@@ -460,37 +456,44 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
 		}
-		this._hasUnsavedChanges = value;
 	}
 
 	_confirmUnsavedChangesBeforeUnload(e) {
-		e.preventDefault();
-		e.returnValue = 'Unsaved changes';
+		if (this.evaluationEntity.hasClass('unsaved')) {
+			e.preventDefault();
+			e.returnValue = 'Unsaved changes';
```
AlexBedley, line 495: I found this pretty confusing until I read https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

Might be worth a comment, as it's quite opaque without the correct context

components/consistent-evaluation-page.js
```diff
@@ -460,37 +456,44 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
+				if (this.evaluationEntity.hasClass('unsaved')) {
+					window.removeEventListener('beforeunload', this.unsavedChangesHandler);
+				}
+				window.location.assign(this.returnHref);
+				break;
+			case 'next':
+				this._onNextStudentClick();
+				break;
+			case 'previous':
+				this._onPreviousStudentClick();
```
AlexBedley, line 487: nit: making this `async` and `await`ing these function calls might make it easier for the next developer, if they are adding other things to this chain of events (functionally no difference now though)

components/consistent-evaluation-page.js
```diff
@@ -460,37 +456,44 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		this._displayToast = false;
 	}
 
-	_showDialog() {
-		this._dialogOpened = true;
+	_showDialog(e) {
+		this.navigationTarget = e.detail.key;
+		if (this.evaluationEntity.hasClass('unsaved')) {
```
AlexBedley, line 461: note that we're hitting the "cache" here instead of getting the freshest data from the API (or entitystore might be slightly fresher). Not a problem, if we're properly invalidating `evaluationEntity` ;)

components/consistent-evaluation-page.js
```diff
@@ -533,6 +536,17 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		this._hideScrollbars();
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this.unsavedChangesHandler = this._confirmUnsavedChangesBeforeUnload.bind(this);
```
AlexBedley, line 541: nit: this is usually done in the constructor, and you can re-use the name
```js
this._confirmUnsavedChangesBeforeUnload = this._confirmUnsavedChangesBeforeUnload.bind(this);
```

components/consistent-evaluation-page.js
jmurray-d2l: would you suggest that we get this elsewhere? how would I go about doing that?

components/consistent-evaluation-page.js
jmurray-d2l: I found that navigating back would cause the browser default pop-up to also show after our brightspace popup without this line

AlexBedley: ahh right I get it

components/consistent-evaluation-page.js
```diff
@@ -460,37 +446,54 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
+			bubbles: true
+		}));
+
+		await this._mutex.dispatch(
+			async() => {
+				this.navigationTarget = e.detail.key;
+				if (this.evaluationEntity.hasClass('unsaved')) {
+					this._dialogOpened = true;
+				} else {
+					this._navigate();
```
AlexBedley, line 461: you'll want to `await` this, otherwise it will immediately give up the mutex. not a big deal now, since navigation is always the last thing to be queued up

AlexBedley: hmm this is actually only half the battle, since we don't wait for the next or previous students to load (that happens outside of the context of this component, and thus the mutex as well). if we ever have the ability to queue things after navigation that might cause a problem. But for M1 it should be ok

components/consistent-evaluation-page.js
```diff
@@ -460,37 +446,54 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
-
-	_updateHasUnsavedChanges(value) {
-		if (!this.confirmUnsavedChanges) return;
-		if (!this._hasUnsavedChanges && value) {
-			window.addEventListener('beforeunload', this._confirmUnsavedChangesBeforeUnload);
-		} else if (this._hasUnsavedChanges && !value) {
-			window.removeEventListener('beforeunload', this._confirmUnsavedChangesBeforeUnload);
+	async _navigate() {
+		switch (this.navigationTarget) {
+			case 'back':
```
AlexBedley, line 476: nit: these should probably be constants somewhere

AlexBedley (approved):

amauryperez (approved):

PR #123, opened 2020-10-21
- Fetching evaluation entity before checking unsaved saved state sinc…

jmurray-d2l:
…e attachments update entity outside of current scope

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -465,8 +465,9 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 
 		await this._mutex.dispatch(
 			async() => {
+				const entity = await this._controller.fetchEvaluationEntity(false);
 				this.navigationTarget = e.detail.key;
-				if (this.evaluationEntity.hasClass('unsaved')) {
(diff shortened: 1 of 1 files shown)


jmurray-d2l:
Attaching files wasn't triggering the unsaved dialog, I believe its because the attachments are handled outside of the scope of this `consistent-evaluation-page.js` .  Following pattern of fetching evaluation entity prior to checking its class

components/consistent-evaluation-page.js
```diff
@@ -465,8 +465,9 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 
 		await this._mutex.dispatch(
 			async() => {
+				const entity = await this._controller.fetchEvaluationEntity(false);
```
AlexBedley, line 468: Note that this call is hitting the EntityStore cache. The advantage there is everything uses the EntityStore cache so this is probably up-to-date "enough" and on a cache-hit you won't cause an API call. The obvious disadvantage is the EntityStore cache could be out of date. However, for the specific use-case you mentioned, this will detect attachment updates.

AlexBedley (approved):

PR #125, opened 2020-10-21
Vy/files launch from submissions page

vix633:
When clicking on a file in assignment submissions, you are now taken to that file within CE


vix633:
@max-hao-d2l FYI

components/consistent-evaluation.js
```diff
@@ -61,14 +62,25 @@ export class ConsistentEvaluation extends MobxLitElement {
 		if (changedProperties.has('href')) {
 			const controller = new ConsistentEvaluationHrefController(this.href, this.token);
 			this._childHrefs = await controller.getHrefs();
+			this.shadowRoot.querySelector('d2l-consistent-evaluation-page')._resetEvidence();
```
amauryperez, line 65: Verify that the acceptance of US121085 is still met

AlexBedley: I think this change might flash the submissions-page during iteration. I'd check that

vix633: @max-hao-d2l This actually does break US121085 for us.

vix633:
Merging master into this silently broke the functionality

components/consistent-evaluation.js
```diff
@@ -69,6 +70,17 @@ export class ConsistentEvaluation extends MobxLitElement {
 			this._iteratorTotal = await controller.getIteratorInfo('total');
 			this._iteratorIndex = await controller.getIteratorInfo('index');
 			this.shadowRoot.querySelector('d2l-consistent-evaluation-page')._resetEvidence();
```
amauryperez, line 72: @max-hao-d2l maybe `resetEvidence` does not need to modify `userProgressOutcomeHref` and that will solve the issue?

max-hao-d2l: I'll try it out

max-hao-d2l: It seems to work. I'll put it on a separate PR

amauryperez: cool, the idea of that method is for the left panel and that href has nothing to do with it.

max-hao-d2l: So... I forgot to create a new branch and accidentally pushed to master. Huge oversight on my part, I won't do it again. Build has succeeded, though.

vix633:
Huge thanks to @amauryperez for pointing out that we have a file-id attribute being set in the rendered already.

components/consistent-evaluation.js
```diff
@@ -99,6 +112,7 @@ export class ConsistentEvaluation extends MobxLitElement {
 				special-access-href=${ifDefined(this._childHrefs && this._childHrefs.specialAccessHref)}
 				return-href=${ifDefined(this.returnHref)}
 				return-href-text=${ifDefined(this.returnHrefText)}
+				current-file-id=${ifDefined(this.fileId !== null && this.fileId)}
```
AlexBedley, line 115: you can simplify this to just `ifDefined(this.fileId)` I think

components/consistent-evaluation.js
```diff
@@ -69,6 +73,15 @@ export class ConsistentEvaluation extends MobxLitElement {
 			this._iteratorTotal = await controller.getIteratorInfo('total');
 			this._iteratorIndex = await controller.getIteratorInfo('index');
 			this.shadowRoot.querySelector('d2l-consistent-evaluation-page')._resetEvidence();
+			this._setFileIdFromUrl();
+		}
+	}
+
+	_setFileIdFromUrl() {
```
AlexBedley, line 80: Do we still need to do this if we're getting the `fileId` directly from props?

components/consistent-evaluation-page.js
```diff
@@ -282,6 +287,14 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		return this.organizationName;
 	}
 
+	async updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (changedProperties.has('submissionInfo') && this.token) {
```
AlexBedley, line 293: I am a bit worried about the trigger for this, it feels a bit fragile. For example, what if `submissionInfo` is set before `currentFileId`?

Additionally I would expect that changing `currentFileId` would be enough to change the current file to that one.

I think ideally `currentFileId` actually does control the current file. To make this happen, we just need to change right panel to accept fileId instead of `fileEvidenceUrl`. Then something lower down can figure out if that's text or file, and how to link `fileEvidenceUrl` and `fileId` (its just through the siren properties). This would really clean up page (we could delete like 3 properties) and make it much easier to reason about (in that world we just literally need to change `currentFileId` to change the current file).

AlexBedley (changes requested):

components/consistent-evaluation-page.js
vix633: I'd like to do this in a followup PR today or tomorrow. Rather than here.

AlexBedley (approved):
Looks good. Not a fan of us being responsible for nuking the query string. I'd prefer if the LMS did that, since it's the one that's reading from it. Not sure how feasible that would be though.

PR #126, opened 2020-10-21
Reset evidence prior to setting _submissionInfo

max-hao-d2l:
`_resetEvidence` sets userProgressOutcomeHref to `undefined` upon initial load. Resetting evidence prior to setting _submisisonInfo fixes this.
Upon testing, this does not have any effect on the existing evaluation page in Assignments (including the iterator buttons).


vix633:
I've included this in my pr https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/125

max-hao-d2l:
Closed in favour of #125

amauryperez:
As long as US121085 still works , this is OK, if not then we would need to find another way that does not break outcomes

PR #127, opened 2020-10-21
- Fixing issue with iterator navigation and rubric grade sync

jmurray-d2l:
This is the only way I could get it to work without making changes outside of our own component

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -308,7 +308,7 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 	}
 
 	async _onNextStudentClick() {
-		this.shadowRoot.querySelector('consistent-evaluation-right-panel')._hasNavigated();
+		this.shadowRoot.querySelector('consistent-evaluation-right-panel')._closeRubric();
 		this.dispatchEvent(new CustomEvent('d2l-consistent-evaluation-next-student-click', {
 			composed: true,
 			bubbles: true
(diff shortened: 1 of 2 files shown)


components/consistent-evaluation-page.js
```diff
@@ -308,13 +308,15 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
 	async _onNextStudentClick() {
+		this.shadowRoot.querySelector('consistent-evaluation-right-panel')._hasNavigated();
 		this.dispatchEvent(new CustomEvent('d2l-consistent-evaluation-next-student-click', {
 			composed: true,
 			bubbles: true
 		}));
 	}
 
 	async _onPreviousStudentClick() {
+		this.shadowRoot.querySelector('consistent-evaluation-right-panel')._hasNavigated();
```
jmurray-d2l, line 319: I tried to pass this as a parameter, but it did not update correctly

amauryperez (approved):

AlexBedley (approved):
Looks good. Clearly not the solution we want to run with long term, but way less work than refactoring rubrics to only fire the event on user-input (vs entity change).

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -186,13 +187,36 @@ export class ConsistentEvaluationRightPanel extends LocalizeMixin(LitElement) {
...
+		if (e.detail && e.detail.expanded) {
+			this.rubricOpen = e.detail.expanded;
+		}
+	}
+
+	_closeRubric() {
+		if (this.hideRubric) {
+			return;
+		}
+		try {
```
jmurray-d2l, line 200: I found that if this logic fails for some reason the navigation fails so added the try/catch for general safety

PR #129, opened 2020-10-23
DE41070 - Show COA achievements selector on right panel

max-hao-d2l:
The COA achievements selector does not show up if there is no initial evaluated checkpoint demonstration. We can instead get the demonstration from the linked Alignments entity.
https://rally1.rallydev.com/#/detail/defect/449083112472?fdp=true

### DIFF
--- a/components/controllers/ConsistentEvaluationHrefController.js
+++ b/components/controllers/ConsistentEvaluationHrefController.js
@@ -1,5 +1,5 @@
 import 'd2l-polymer-siren-behaviors/store/entity-store.js';
-import { actorRel, alignmentsRel, assessmentRel, checkpointItemType, demonstrationRel, editSpecialAccessApplicationRel, evaluationRel, groupRel, nextRel, previousRel, rubricRel, userProgressOutcomeRel, userRel } from './constants.js';
+import { actorRel, alignmentsRel, assessmentRel, demonstrationRel, editSpecialAccessApplicationRel, evaluationRel, groupRel, nextRel, previousRel, rubricRel, userProgressOutcomeRel, userRel } from './constants.js';
 import { Classes, Rels } from 'd2l-hypermedia-constants';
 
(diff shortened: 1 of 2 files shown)


max-hao-d2l:
Note: We also have an end-to-end for the Mastery View Evaluation Page
LD Flags: `checkpoints-enabled`, `outcomes-overall`
1. Create an achievements scale and set it as default
1. Create a list of outcomes in a course
1. Align outcomes to an assignment/rubric/discussion
1. Evaluate a student with the outcomes
1. Go to Grades and click on "Mastery"
1. Click on any cell on the table
1. The consistent-eval page should show up

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -75,34 +76,21 @@ export class ConsistentEvaluationHrefController {
...
-							if (checkpointActivityEntity) {
-								const demonstrationEntity = checkpointActivityEntity.getSubEntityByClass(Classes.outcomes.demonstration);
-								if (demonstrationEntity) {
-									coaDemonstrationHref = this._getHref(demonstrationEntity, 'self');
-								}
+						const referencedAlignmentEntity = alignmentsEntity.entity.getSubEntityByRel('item');
+						if (referencedAlignmentEntity) {
+							const alignmentEntity = await this._getEntityFromHref(referencedAlignmentEntity.href, bypassCache);
+							if (alignmentEntity && alignmentEntity.entity) {
+								coaDemonstrationHref = alignmentEntity.entity.getLinkByRel(demonstrationRel)?.href;
```
jeroach, line 85: Why do we have the `?` before the `href` ?

Do we need another `if` check before assignment?

max-hao-d2l: Yeah, I'll put it in

jeroach (approved):

PR #132, opened 2020-10-23
Abedley/annotations/file id refactor

AlexBedley:
Problem: I started this refactor because I needed the id of the currently displayed file to properly save annotations. This sparked a whole chain of events which lead to this PR. Since I was already in the area I cleaned up a bunch of stuff relating to files.

Solution:

The new way of doing things is we pass `currentFileId` everywhere. This uniquely determines what file is currently being shown. This also has the advantage of us not having to know whether it's text or a file, and passing large amounts of information up in events and down through props.

Now `left-panel` (and below) are the only things that really care if it's text vs file. I tried to make LCB not care at all, but I was held up by the lateness thing. So LCB can take a `fileId` and get the `lateness`.

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -17,7 +17,6 @@ import { ConsistentEvaluationController } from './controllers/ConsistentEvaluati
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { loadLocalizationResources } from './locale.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
-import { submissions } from './controllers/constants';
 
 const DIALOG_ACTION_LEAVE = 'leave';
 
@@ -144,18 +143,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 			},
 			_scrollbarStatus: {
 				attribute: false
-			},
-			_fileEvidenceUrl: {
-				attribute: false,
-				type: String
-			},
-			_textEvidence: {
-				attribute: false,
-				type: Object
-			},
-			_selectedFile: {
-				attribute: false,
-				type: String
 			},
 			_dialogOpened: {
 				attribute: false
@@ -325,13 +312,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 			composed: true,
 			bubbles: true
 		}));
-	}
-
-	_resetEvidence() {
-		this._selectedFile = submissions;
-		this.submissionInfo = undefined;
-		this._fileEvidenceUrl = undefined;
-		this._textEvidence = undefined;
 	}
 
 	_hideScrollbars() {
@@ -527,7 +507,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 				<d2l-consistent-evaluation-learner-context-bar
 					user-href=${ifDefined(this.userHref)}
 					group-href=${ifDefined(this.groupHref)}
-					selected-item-name=${this._selectedFile}
 					special-access-href=${ifDefined(this.specialAccessHref)}
 					.token=${this.token}
 					.currentFileId=${this.currentFileId}
@@ -537,29 +516,18 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 		}
 	}
 
+	_selectFile(e) {
+		this.currentFileId = e.detail.fileId;
+		this._hideScrollbars();
+	}
+
 	_setSubmissionsView() {
 		this.currentFileId = undefined;
-	}
-	_setFileEvidence(e) {
-		this._fileEvidenceUrl = e.detail.url;
-		this._selectedFile = e.detail.name;
-		this._textEvidence = undefined;
-		this._hideScrollbars();
-	}
-	_setTextEvidence(e) {
-		this._textEvidence = e.detail.textSubmissionEvidence;
-		this._selectedFile = e.detail.textSubmissionEvidence.name;
-		this._fileEvidenceUrl = undefined;
-		this._hideScrollbars();
+		this._showScrollbars();
 	}
 
 	async _handleAnnotationsUpdate() {
 
-	}
-
-	_selectFile(e) {
-		console.log(e);
-		this.currentFileId = e.detail.fileId;
 	}
 
 	connectedCallback() {
@@ -595,8 +563,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 					<d2l-consistent-evaluation-left-panel
 						.submissionInfo=${this.submissionInfo}
 						.token=${this.token}
-						file-evidence-url=${ifDefined(this._fileEvidenceUrl)}
-						.textEvidence=${this._textEvidence}
 						user-progress-outcome-href=${ifDefined(this.userProgressOutcomeHref)}
 						.currentFileId=${this.currentFileId}
 						@d2l-consistent-eval-annotations-update=${this._handleAnnotationsUpdate}
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -530,7 +530,7 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement) 
 					selected-item-name=${this._selectedFile}
 					special-access-href=${ifDefined(this.specialAccessHref)}
 					.token=${this.token}
-					current-file-id=${this.currentFileId}
+					.currentFileId=${this.currentFileId}
 					.submissionInfo=${this.submissionInfo}
 				></d2l-consistent-evaluation-learner-context-bar>
 			`;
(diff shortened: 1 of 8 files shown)


components/consistent-evaluation-page.js
```diff
@@ -146,18 +144,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
 			},
-			_fileEvidenceUrl: {
-				attribute: false,
-				type: String
-			},
-			_textEvidence: {
-				attribute: false,
-				type: Object
-			},
-			_selectedFile: {
```
AlexBedley, line 157: page no longer needs to track things like file name, evidence urls, etc.

components/consistent-evaluation-page.js
```diff
@@ -287,14 +272,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		return this.organizationName;
 	}
 
-	async updated(changedProperties) {
-		super.updated(changedProperties);
-
-		if (changedProperties.has('submissionInfo') && this.token) {
```
AlexBedley, line 293: Instead of having page determine all the file information from `fileId`, it just passes `fileId` down to it's children and lets them do their thing

components/consistent-evaluation-page.js
```diff
@@ -337,13 +314,6 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 		}));
 	}
 
-	_resetEvidence() {
```
AlexBedley, line 340: This has been replaced with `_setSubmissionsView`, which is essentially the same thing

components/consistent-evaluation-page.js
```diff
@@ -537,68 +507,27 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
...
-	_setTextEvidence(e) {
-		this._textEvidence = e.detail.textSubmissionEvidence;
-		this._selectedFile = e.detail.textSubmissionEvidence.name;
-		this._fileEvidenceUrl = undefined;
+	_selectFile(e) {
+		this.currentFileId = e.detail.fileId;
 		this._hideScrollbars();
 	}
 
-	async setSubmissionViewFromUrl() {
```
AlexBedley, line 568: This is now done in left-panel

components/consistent-evaluation.js
```diff
@@ -112,7 +112,7 @@ export class ConsistentEvaluation extends MobxLitElement {
 				special-access-href=${ifDefined(this._childHrefs && this._childHrefs.specialAccessHref)}
 				return-href=${ifDefined(this.returnHref)}
 				return-href-text=${ifDefined(this.returnHrefText)}
-				current-file-id=${ifDefined(this.fileId !== null && this.fileId)}
+				current-file-id=${ifDefined(this.fileId)}
```
AlexBedley, line 115: random cleanup, unrelated to this PR really

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -72,73 +67,45 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
 		if (changedProperties.has('submissionInfo')) {
 			this._files = await getSubmissions(this.submissionInfo, this.token);
-			this._submissionLateness = undefined;
 		}
 
-		this._showFiles = (this._files && this._files.length > 0);
+		if ((changedProperties.has('currentFileId') || changedProperties.has('submissionInfo')) && this._files) {
+			const currentFile = findFile(this.currentFileId, this._files);
+			if (currentFile) {
+				this._submissionLateness = currentFile.latenessTimespan;
```
AlexBedley, line 80: I don't like that LCB knows how to do this. Ideally it's just told what the `submissionLateness` is. But this PR was already getting insanely large, this could be done in a future PR

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -33,9 +29,8 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 				attribute: false,
 				type: Number
 			},
-			_showFiles: {
```
AlexBedley, line 36: I found that this calculation was so simple (does files exists and have elements), I just removed the computed value and shoved it into the if-statement so we didn't have to keep it in-sync

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -72,73 +67,45 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
-			const event = new Event('d2l-consistent-evaluation-evidence-back-to-user-submissions', {
+			this.dispatchEvent(new Event('d2l-consistent-evaluation-evidence-back-to-user-submissions', {
 				composed: true,
 				bubbles: true
-			});
-			this.dispatchEvent(event);
-			return;
-		}
-
-		const submission = JSON.parse(e.target.value);
```
AlexBedley, line 101: Note we don't have to parse JSON here anymore since the only thing we need to know to tell page to select the file is just the `id`

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -72,73 +67,45 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
...
 			composed: true,
 			bubbles: true
-		});
-		this.dispatchEvent(event);
+		}));
 	}
 
 	_renderLateButton() {
-		if (this.selectedItemName === submissions || !this._submissionLateness)
+		if (!this.currentFileId || !this._submissionLateness)
```
AlexBedley, line 108: I got rid of `selectedItemName`. Seemed fragile anyways since names are not unique and a student could easily submit a file called `submissions`. Instead, the lack of a `fileId` means user submissions is currently viewed. I don't like this much either, but this could be addressed in another PR. (I liked @jmurray-d2l 's approach in his PR that he declined)

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -91,8 +86,9 @@ export class ConsistentEvaluationLearnerContextBar extends RtlMixin(LitElement)
 				?is-group-activity="${this.groupHref}"
 			></d2l-consistent-evaluation-lcb-user-context>
 			<d2l-consistent-evaluation-lcb-file-context
-				selected-item-name=${this.selectedItemName}
+				.token="${this.token}"
```
AlexBedley, line 89: We actually weren't passing token down to lcb-file-context. I guess it was still able to make API calls because entity-store falls back to using cookies when token is not present

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -71,11 +70,7 @@ export class ConsistentEvaluationLearnerContextBar extends RtlMixin(LitElement)
 	}
 
 	_getIsExempt() {
-		if (this.submissionInfo && this.submissionInfo.isExempt) {
-			return true;
-		} else {
-			return false;
-		}
+		return this.submissionInfo && this.submissionInfo.isExempt;
```
AlexBedley, line 73: I just simplified this, unrelated to this PR

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -89,6 +94,48 @@ export class ConsistentEvaluationLeftPanel extends LocalizeMixin(LitElement) {
...
+		) {
+			await this.getFileFromId();
+		} else if (changedProperties.has('currentFileId') && !this.currentFileId) {
+			this.textEvidence = undefined;
+			this.fileEvidenceUrl = undefined;
+		}
+
+	}
+
+	async getFileFromId() {
```
AlexBedley, line 115: This is where the magic happens now where we resolve `fileId` to something we can render

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -395,7 +372,9 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeMixin(L
 					${this.submissionType === textSubmission ? html`
 						<d2l-menu-item-link text="${this.localize('viewFullSubmission')}"
 							href="javascript:void(0);"
-							@click="${this._dispatchRenderEvidenceTextEvent}"></d2l-menu-item-link>` : null}
+							@click="${
+	// eslint-disable-next-line lit/no-template-arrow
```
AlexBedley, line 376: I tried to avoid this weird eslint exception, but it was becoming more trouble than it was worth. the `data-key` method wouldn't work because the event target was the entire application for reasons I'm not sure

jmurray-d2l (approved):
Looks good to me! We'll probably want to give Peggy a heads up given this touches so many workflows.  Always nice to see more red than green, nice work!

vix633 (approved):
Noice

PR #134, opened 2020-10-26
Jmurray/fix submission list alternate language

jmurray-d2l:


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -144,7 +144,7 @@ export class ConsistentEvaluationHrefController {
 			if (assignmentHref) {
 				const assignmentEntity = await this._getEntityFromHref(assignmentHref, false);
 				if (assignmentEntity && assignmentEntity.entity) {
-					submissionType = assignmentEntity.entity.properties.submissionType.title;
+					submissionType = assignmentEntity.entity.properties.submissionType.value.toString();
```
jmurray-d2l, line 147: I found that in some usages of this it was already converted to a string, so in order to get it to work everywhere with the constant I just  convert it immediately - seems weird to me, but safer I guess

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -144,7 +144,7 @@ export class ConsistentEvaluationHrefController {
 			if (assignmentHref) {
 				const assignmentEntity = await this._getEntityFromHref(assignmentHref, false);
 				if (assignmentEntity && assignmentEntity.entity) {
-					submissionType = assignmentEntity.entity.properties.submissionType.title;
```
jmurray-d2l, line 147: The title is translated :),  I didn't want to remove that because I believe FACE is using the translation

amauryperez (approved):

components/controllers/ConsistentEvaluationHrefController.js
AlexBedley: yeah I noticed some weirdness around this as well, probably safer to make everything a string, unless it's guaranteed to be numeric

components/controllers/constants.js
```diff
@@ -26,10 +26,10 @@ export const itemRel = 'item';
 export const checkpointItemType = 'checkpoint-item';
 export const publishedState = 'Published';
 export const draftState = 'Draft';
-export const fileSubmission = 'File submission';
```
AlexBedley, line 29: I think this might break some demo pages, but ever since BSI became so much easier to use, I don't think anyone is using demo pages

AlexBedley (approved):

PR #136, opened 2020-10-26
DE41081 - Update Overall Achievement Left Panel upon Updating Evaluation

max-hao-d2l:
Related: https://github.com/Brightspace/d2l-outcomes-overall-achievement/pull/37
Update the left panel component for Overall Achievement when an evaluation is saved/updated/published/retracted


jeroach (commented):
Hmmm, I get the feeling that using the chained `shadowRoot` calls isn't the best practice for something like this, but I'm not sure what would be a good alternative.

Can't use events since we're going down the tree, not up.

Curious as to others thoughts.

components/consistent-evaluation-page.js
```diff
@@ -382,6 +385,9 @@ export default class ConsistentEvaluationPage extends LocalizeMixin(LitElement)
 				this.evaluationEntity = await this._controller.update(entity);
 				if (!(this.evaluationEntity instanceof Error)) {
 					this._showToast(this.localize('updated'));
+					if (this.userProgressOutcomeHref) {
```
rgoessl, line 388: Could make this more generic and invoke a general refresh on the left pane which could be responded to/ignored by the inner component

rgoessl (approved):

rgoessl:
> Hmmm, I get the feeling that using the chained `shadowRoot` calls isn't the best practice for something like this, but I'm not sure what would be a good alternative.
> 
> Can't use events since we're going down the tree, not up.
> 
> Curious as to others thoughts.

Could pass a `refresh` prop through all the children and make a change to it whenever there's new data. Then a child that wants to refresh can do so whenever the passed prop changes

AlexBedley (commented):
I think I'm missing some context here, but I agree with Josh, this approach doesn't seem very idiomatic in Lit. I'd be interested in learning more around what we need to accomplish (how does saving the evaluation change the underlying entity?)

We've committed similar sins though with our `d2l-flush` feature. In that case we use `window` events. Also a bit of a design smell but at least it limits the hack to 2 small functions. That might work for you, especially if you have other things that need to happen on save/publish.

max-hao-d2l:
Using a window event sounds like a cleaner fix. I'll try it out.
In our case, saving the evaluation should update a component in the left panel which takes in an href and token. Since the href doesn't change when saving the evaluation, the component doesn't update.

rgoessl (approved):

jeroach (approved):

PR #139, opened 2020-10-27
US121725 - Submissions - Evidence - Opening a submission does not mark the submission as read

amauryperez:


components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -77,7 +77,7 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeMixin(L
 		if ((changedProperties.has('currentFileId') || changedProperties.has('submissionInfo')) && this._files) {
 			const currentFile = findFile(this.currentFileId, this._files);
 			if (currentFile) {
-				this._submissionLateness = currentFile.latenessTimespan;
+				this._submissionLateness = currentFile.properties.latenessTimespan;
```
jmurray-d2l, line 80: is it worth checking for `currentFile.properties` ?

components/helpers/submissionsAndFilesHelpers.js
```diff
@@ -5,10 +5,10 @@ import { Classes } from 'd2l-hypermedia-constants';
 export function findFile(fileId, submissions) {
 	for (let i = 0; i < submissions.length; i++) {
 		const submission = submissions[i];
-		const files = getSubmissionFiles(submission);
+		const files = submission.entity.getSubEntityByRel(attachmentListRel).entities;
```
jmurray-d2l, line 8: should we check for `getSubEntityByRel` return before accessing it's entities?

components/header/d2l-consistent-evaluation-lcb-file-context.js
amauryperez: dont we always have properties?

jmurray-d2l (approved):

components/helpers/submissionsAndFilesHelpers.js
```diff
@@ -5,10 +5,17 @@ import { Classes } from 'd2l-hypermedia-constants';
 export function findFile(fileId, submissions) {
 	for (let i = 0; i < submissions.length; i++) {
 		const submission = submissions[i];
-		const files = getSubmissionFiles(submission);
```
AlexBedley, line 8: `getSubmissionFiles` decorated the properties object and renamed a bunch of stuff (for better or worse, but that's another issue). Now that we're no longer using it, do we need to change consumers of this function in tandem?

For example, I don't think the `submissionLateness` property is actually from the file's `properties` it seems to be from the submission's properties

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/dfee293f1e32fefe6f66cd42444c4267c30766dc/components/helpers/submissionsAndFilesHelpers.js#L34

components/helpers/submissionsAndFilesHelpers.js
```diff
@@ -5,10 +5,17 @@ import { Classes } from 'd2l-hypermedia-constants';
...
+		if (!submission.entity) {
+			continue;
+		}
+		const attachmentList = submission.entity.getSubEntityByRel(attachmentListRel);
+		if (!attachmentList || !attachmentList.entities) {
+			continue;
```
AlexBedley, line 13: were we seeing some bad data or is this just for safety?

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -94,48 +96,56 @@ export class ConsistentEvaluationLeftPanel extends LocalizeMixin(LitElement) {
...
+		const submissions = await getSubmissions(this.submissionInfo, this.token);
+		const currentFile = findFile(this.currentFileId, submissions);
+
+		if (!currentFile) {
+			console.error(`Cannot find fileId ${this.currentFileId}`);
+			return;
+		}
+
+		const action = currentFile.getActionByName(toggleIsReadActionName);
+		for (let i = 0; i < action.fields.length; i++) {
```
AlexBedley, line 127: You can simplify this (and make it more clear we're just looking for a single field that fits the criteria) by using `some`

```js
if (action.fields.some(f => f.name === 'isRead' && f.value)) {
  performSirenAction(this.token, action, undefined, true);
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

AlexBedley (changes requested):

components/helpers/submissionsAndFilesHelpers.js
amauryperez: Interesting, I tested it and it works fine. Somehow that property is getting added, that is so weird

components/helpers/submissionsAndFilesHelpers.js
amauryperez: not bad data, I guess is just for safety. Do you think we don't need this check?

components/helpers/submissionsAndFilesHelpers.js
amauryperez: Adding those extra properties seems a bit weird :-(

components/helpers/submissionsAndFilesHelpers.js
AlexBedley: On one hand if the data isn't the way we expect it might be better to know about it by failing loudly. But on the other hand we don't want to crash the entire application for a non-essential feature. We had a bit of a debate over this during the QE implementation and it was never really decided.

components/helpers/submissionsAndFilesHelpers.js
AlexBedley: yeah I wonder if the entity store cache objects themselves are getting updated with these properties (if someone called `getSubmissionFiles` before you called `findFile`)

I do think adding those properties is a design smell. I would approve a PR to fix that after M1

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -94,48 +96,54 @@ export class ConsistentEvaluationLeftPanel extends LocalizeMixin(LitElement) {
...
+
+		if (!currentFile) {
+			console.error(`Cannot find fileId ${this.currentFileId}`);
+			return;
+		}
+
+		const action = currentFile.getActionByName(toggleIsReadActionName);
+		if (action.fields.some(f => f.name === 'isRead' && f.value)) {
+			// If the action value is true it means it can be called to set the IsRead value to true, otherwise it is already read and we dont want to unread it
+			performSirenAction(this.token, action, undefined, true);
```
AlexBedley, line 129: we should `await` this call

not a huge deal right now, but could be a problem if this function was modified.

components/helpers/submissionsAndFilesHelpers.js
amauryperez: yeah, we should clean that for sure

AlexBedley (approved):

PR #141, opened 2020-10-27
Jmurray/us121522 console error

jmurray-d2l:


AlexBedley (approved):
nice! how did you track this issue down? I wouldve never have noticed this

jmurray-d2l:
The js stack trace pointed to the lcb, and then I just poked around in there and found it eventually!

PR #149, opened 2020-10-29
Abedley/annotations/actually save

AlexBedley:
Problem: Annotations aren't being saved in CE
Solution: This is the first PR to cover the majority of cases. Basically, now that we've added an API route for annotations similar to grades/feedback, I just implemented annotations in the same way.

The reason more work is needed is because annotatations are file-specific (vs grades or feedback which are specific to the entire evaluation). Thus we need to prompt users when changing files, because they will lose their transiently saved annotations data. That feature is not included in this PR as I will need API changes to detect if I have unsaved annotations data.

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -16,7 +16,6 @@ import { Awaiter } from './awaiter.js';
 import { ConsistentEvaluationController } from './controllers/ConsistentEvaluationController.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeConsistentEvaluation } from '../lang/localize-consistent-evaluation.js';
-import { performSirenAction } from 'siren-sdk/src/es6/SirenAction.js';
 
 const DIALOG_ACTION_LEAVE = 'leave';
 
@@ -347,6 +346,18 @@ export default class ConsistentEvaluationPage extends LocalizeConsistentEvaluati
 		);
 	}
 
+	async _transientSaveAnnotations(e) {
+		await this._mutex.dispatch(
+			async() => {
+				const entity = await this._controller.fetchEvaluationEntity(false);
+				const annotationsData = e.detail;
+				const fileId = this.currentFileId;
+
+				this.evaluationEntity = await this._controller.transientSaveAnnotations(entity, annotationsData, fileId);
+			}
+		);
+	}
+
 	async _saveEvaluation() {
 		window.dispatchEvent(new CustomEvent('d2l-flush', {
 			composed: true,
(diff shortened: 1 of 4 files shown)


components/consistent-evaluation-page.js
```diff
@@ -346,6 +346,18 @@ export default class ConsistentEvaluationPage extends LocalizeConsistentEvaluati
 		);
 	}
 
+	async _transientSaveAnnotations(e) {
```
AlexBedley, line 349: nothing groundbreaking here. basically the same as `_transientSaveFeedback`

components/consistent-evaluation-page.js
```diff
@@ -574,7 +582,7 @@ export default class ConsistentEvaluationPage extends LocalizeConsistentEvaluati
 						.token=${this.token}
 						user-progress-outcome-href=${ifDefined(this.userProgressOutcomeHref)}
 						.currentFileId=${this.currentFileId}
-						@d2l-consistent-eval-annotations-update=${this._handleAnnotationsUpdate}
+						@d2l-consistent-eval-annotations-update=${this._transientSaveAnnotations}
```
AlexBedley, line 585: renamed it to match grades/feedback

components/controllers/ConsistentEvaluationController.js
```diff
@@ -110,6 +110,23 @@ export class ConsistentEvaluationController {
 		return await this._performAction(targetEntity, saveGradeActionName, saveGradeFieldName, gradeValue);
 	}
 
+	async transientSaveAnnotations(evaluationEntity, annotationsData, fileId) {
+		const annotationsEntity = evaluationEntity.getSubEntityByRel('annotations');
+		const saveAnnotationsAction = annotationsEntity.getActionByName('SaveAnnotations');
```
AlexBedley, line 115: hmm I could probably be safer here and check that these things exist. I guess one test case I didn't cover is annotations off. I'll check what happens

components/controllers/ConsistentEvaluationController.js
```diff
@@ -110,6 +110,23 @@ export class ConsistentEvaluationController {
...
+
+		const encodedAnnotationsData = {
+			FileId: fileId,
+			AnnotationsData: JSON.stringify(annotationsData)
+		};
+
+		const fields = [{
+			name: 'value',
+			value: JSON.stringify(encodedAnnotationsData)
```
AlexBedley, line 124: image

components/left-panel/consistent-evaluation-evidence-file.js
```diff
@@ -74,11 +81,21 @@ export class ConsistentEvaluationEvidenceFile extends LitElement {
...
 
 	_handleAnnotationsUpdate(e) {
-		this.dispatchEvent(new CustomEvent('d2l-consistent-eval-annotations-update', {
-			composed: true,
-			bubbles: true,
-			detail: e.data.value
-		}));
+		this._debounceJobs.annotations = Debouncer.debounce(
+			this._debounceJobs.annotations,
+			timeOut.after(1000),
```
AlexBedley, line 86: I picked a longer debounce time because these requests are much bigger, and every subsequent request contains all the other request data before it combined (ie request 3 will send `[1, 2, 3]` instead of just `[3]`)

components/controllers/ConsistentEvaluationController.js
jmurray-d2l: You might be able to get away with only checking the entity by rel returned depending on how the entity store handles a null action. bit:  could maybe use the local `_performAction`?

jmurray-d2l (approved):

jeroach (approved):

components/controllers/ConsistentEvaluationController.js
AlexBedley: I'll shore this up and use the local `_performAction` in the next PR. Originally I thought I'd have to set 2 fields (fileId and annotationsData) so I didn't think the local one would work for me

PR #150, opened 2020-10-30
US122010 - Submissions - Evidence - Downloading a file does not mark the file as read

amauryperez:


components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -155,17 +174,19 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
 					const submissionDate = submissionEntity.getSubEntityByClass(Classes.assignments.submissionDate).properties.date;
 					const evaluationState = submissionEntity.properties.evaluationStatus;
 					const latenessTimespan = submissionEntity.properties.lateTimeSpan;
+					const submissionNumber = submissionEntity.properties.submissionNumber;
 					itemTemplate.push(html`
 						<d2l-consistent-evaluation-submission-item
 							date-str=${submissionDate}
-							display-number=${this._submissionEntities.length - i}
+							display-number=${submissionNumber}
```
amauryperez, line 181: fixing this to use the value from the api

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -139,8 +139,27 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
...
+		}
+
+		const action = attachmentEntity.getActionByName(actionName);
+		if (action.fields.some(f => f.name === 'isRead' && f.value)) {
+			// If the action value is true it means it can be called to set the IsRead value to true, otherwise it is already read and we dont want to unread it
+			await this._doSirenActionAndRefreshFileStatus(action);
+		}
+	}
+
+	async _doSirenActionAndRefreshFileStatus(action) {
```
amauryperez, line 161: created this new function to avoid code duplication

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -219,6 +219,22 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 		this.dispatchEvent(event);
 	}
 
+	_dispatchDownloadEvent(e) {
+		const fileId = e.target.getAttribute('data-key');
+		const downloadHref = e.target.getAttribute('data-href');
+		const event = new CustomEvent('d2l-consistent-evaluation-evidence-file-download', {
```
amauryperez, line 225: sending event to mark as read

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -139,8 +139,27 @@ export class ConsistentEvaluationSubmissionsPage extends LitElement {
...
+	async _downloadAction(e) {
+		const fileId = e.detail.fileId;
+		const actionName = e.detail.action;
 
+		const attachmentEntity = this._getAttachmentEntity(fileId);
+		if (!attachmentEntity) {
+			throw new Error('Invalid entity provided for attachment');
+		}
+
+		const action = attachmentEntity.getActionByName(actionName);
```
jmurray-d2l, line 154: Are we expecting to use other actions as well?

amauryperez: not really, I guess I can just use the constant directly here instead of passing it

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -219,6 +219,22 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
...
+			detail: {
+				fileId: fileId,
+				action: toggleIsReadActionName
+			},
+			composed: true,
+			bubbles: true
+		});
+		this.dispatchEvent(event);
+
+		window.location = downloadHref;
```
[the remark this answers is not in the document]
AlexBedley: does this change the user's browser to this dl address? or does it just download the file without changing your current screen?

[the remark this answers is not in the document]
amauryperez: it just downloads the file

jmurray-d2l (approved):

components/left-panel/consistent-evaluation-submission-item.js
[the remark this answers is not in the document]
AlexBedley: Normally this logic should be in the event handler (`_downloadAction`). It's a bit weird in this case since the thing we're operating on is a global (`window`). But reading this code I would just expect this function to create the event, and the actual consequences would be in the handler

AlexBedley (approved):

components/left-panel/consistent-evaluation-submission-item.js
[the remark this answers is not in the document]
amauryperez: so you ok with leaving it here? it might be weirder having the `windows.location` thing in the handler

PR #151, opened 2020-11-02
Jmurray/loading skeleton

jmurray-d2l:


components/consistent-evaluation.js
```diff
@@ -57,6 +61,11 @@ export class ConsistentEvaluation extends LitElement {
 		this._gradeItemInfo = undefined;
 		this.returnHref = undefined;
 		this.returnHrefText = undefined;
+		this._loading = true;
+		this._loadingComponents = {
+			'main' : true,
+			'submissions': true
```
AlexBedley, line 67: nit: I'm surprised the linter didn't complain, but in general we avoid using strings as keys, unless you absolutely have to (to escape a character)

AlexBedley (approved):
Looks good. One thing you might consider is refactoring the loading implementation to a class so it can be unit-tested and abstracted in case we ever change the implementation. See `Awaiter.js` for examples

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/awaiter.js

PR #154, opened 2020-11-03
Jmurray/file attachments fix

jmurray-d2l:
the attachementsHref was not being updated correctly.  As suggested by Alex I changed the pattern to grab the href from the entity when needed instead of caching it, I applied this pattern to the other members that were also being cached for consistency and to avoid more state bugs

@vix633 richTextEditorDisabled doesn't look like it is being used?  I removed it, let me know if thats a problem

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -175,8 +175,6 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		this._scrollbarStatus = 'default';
 		this._mutex = new Awaiter();
 		this._dialogOpened = false;
-		this.allowEvaluationWrite = false;
-		this.allowEvaluationDelete = false;
 		this.unsavedChangesHandler = this._confirmUnsavedChangesBeforeUnload.bind(this);
 	}
 
@@ -407,7 +405,6 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 					this._showToast(this.localize('publishError'));
 				}
 				this.submissionInfo.evaluationState = publishedState;
-				this.allowEvaluationDelete = this._controller.userHasDeletePermission(this.evaluationEntity);
 			}
 		);
 	}
@@ -430,7 +427,6 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 				}
 				this.evaluationState = this.evaluationEntity.properties.state;
 				this.submissionInfo.evaluationState = draftState;
-				this.allowEvaluationWrite = this._controller.userHasWritePermission(this.evaluationEntity);
 			}
 		);
 	}
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -543,7 +543,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 
 	_getAttachmentsLink() {
 		if (!this.evaluationEntity || !this.evaluationEntity.hasLinkByRel(attachmentsRel)) {
-			return null;
+			return undefined;
 		}
 
 		return this.evaluationEntity.getLinkByRel(attachmentsRel).href;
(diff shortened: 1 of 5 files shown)


components/consistent-evaluation-page.js
```diff
@@ -48,14 +49,6 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 				attribute: 'special-access-href',
 				type: String
 			},
-			richTextEditorDisabled: {
```
vix633, line 51: Just test to make sure the rich text editor still works.

components/consistent-evaluation-page.js
```diff
@@ -230,7 +220,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 
 	get _feedbackText() {
 		if (this._feedbackEntity && this._feedbackEntity.properties) {
-			return this._feedbackEntity.properties.text || '';
+			return this._feedbackEntity.properties.html || '';
```
jmurray-d2l, line 223: This change is unrelated, thought I would just sneak it in here.  Any styling that was applied in the feedback box was not being shown upon reload.  The html property has the correct styling

components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -112,7 +112,7 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeConsistentEvalu
 	}
 
 	render() {
-		if (this.href && this.token) {
+		if (this.href && this.token && this.richTextEditorConfig) {
```
jmurray-d2l, line 115: This just stops an initial error from appearing in the console, no functional difference

components/consistent-evaluation-page.js
```diff
@@ -274,10 +264,6 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		const bypassCache = true;
 		this.evaluationEntity = await this._controller.fetchEvaluationEntity(bypassCache);
 		this.evaluationState = this.evaluationEntity.properties.state;
-		this.allowEvaluationWrite = this._controller.userHasWritePermission(this.evaluationEntity);
```
AlexBedley, line 277: hmm weird that these weren't properties. oh well

components/consistent-evaluation-page.js
```diff
@@ -557,6 +541,44 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		}));
 	}
 
+	_getAttachmentsLink() {
```
AlexBedley, line 544: might make more sense for these 4 functions to live in a helper class for unit-testability

components/consistent-evaluation-page.js
```diff
@@ -557,6 +541,44 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+		if (!this.evaluationEntity || !this.evaluationEntity.hasLinkByRel(attachmentsRel)) {
+			return null;
```
AlexBedley, line 546: what does `IfDefined` do with `null`?

components/consistent-evaluation-page.js
```diff
@@ -596,19 +618,18 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 						rubric-assessment-href=${ifDefined(this.rubricAssessmentHref)}
 						outcomes-href=${ifDefined(this.outcomesHref)}
 						coa-eval-override-href=${ifDefined(this.coaDemonstrationHref)}
-						attachments-href=${ifDefined(this.attachmentsHref)}
-						.richTextEditorConfig=${this.richtextEditorConfig}
+						attachments-href=${ifDefined(this._getAttachmentsLink())}
```
AlexBedley, line 621: I wonder if lit is smart enough to know to re-render if any dependencies in these functions change. probably doesn't matter for this PR (since we use `evaluationEntity` elsewhere in `render`)

AlexBedley (approved):

components/consistent-evaluation-page.js
dbatiste: The `ifDefined` directive only looks for `undefined`.  It treats `null` as a valid value.

jmurray-d2l: I'll update to return undefined

PR #155, opened 2020-11-04
Bypassing cache when fetching attachments

jmurray-d2l:
Problem:  When you attach a file and then publish or save draft the attached file is removed from the ui (its saved correctly)

Solution: No longer hit the cache when fetching file attachments

I think this is an easy fix, but exposes some issues with our overall design.  Ideally we don't re-fetch the attachment list after publish since we already know what is attached.


AlexBedley (approved):
Makes sense. As you pointed out, ideally we just assume it worked and make the API call in the background. A feature for another day

PR #156, opened 2020-11-04
Right-panel mobile adaptive design

JamieTran:
The right panel is now mobile friendly with the inclusion of the mobile drawer from Primary-Secondary and the implementation of lists in that view. 

- When switching between mobile and web view the state of the component is maintained.
- Dialogs opened from other dialogs are now supported thanks to @jeroach and @dbatiste 

mobile-adaptive-view


components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +38,61 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
 		`];
 	}
 
 	constructor() {
 		super();
+		this._dialogOpened = false;
 		this.noTitle = false;
+		this.supportingInfo = undefined;
+		this._isMobile = (window.innerWidth <= MOBILE_WINDOW_THRESHHOLD);
+		window.addEventListener('resize', this._handleResize.bind(this));
```
JamieTran, line 56: I know we wanted to avoid using a listener for resizing but I couldn't find an elegant solution with only CSS

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -101,6 +101,7 @@ export class ConsistentEvaluationRightPanel extends LocalizeConsistentEvaluation
 		this.allowEvaluationWrite = false;
 		this.rubricOpen = false;
 		this.attachmentsHref = null;
+		window.D2L.DialogMixin.preferNative = false;
```
JamieTran, line 104: Fix for legacy dialogs rendering behind full-screen dialogs

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +38,61 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+				@click=${this._toggleOpenDialog}>
+				<d2l-list-item-content class="no-border">
+					${this._getTitle()}
+					<div slot="supporting-info">${this.supportingInfo}</div>
+				</d2l-list-item-content>
+			</d2l-list-item>
+		`;
+	}
+
 	render() {
```
JamieTran, line 86: State & data are being maintained but not in this component. My best guess is that state is being handled in` right-panel` and the components are just slotted in here.

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -1,6 +1,13 @@
+import '@brightspace-ui/core/components/colors/colors.js';
```
dbatiste, line 1: I don't see any color references here... is this color import needed?

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +38,61 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+				padding-right: 0.75rem;
 			}
 		`];
 	}
 
 	constructor() {
 		super();
+		this._dialogOpened = false;
 		this.noTitle = false;
+		this.supportingInfo = undefined;
```
dbatiste, line 54: Nit: it's not necessary to initialize strings as `undefined`.  Variables are initially `undefined` unless assigned a value.

components/right-panel/consistent-evaluation-right-panel-block.js
dbatiste: This handler can result in a memory leak because when/if the element is removed from the DOM, the handler (which has references to this component) remains, holding onto all the references preventing the element and its references from being garbage collected.  Event handlers that are wired up to elements outside the components local DOM scope should be added in `connectedCallback`, and removed in `disconnectedCallback`.  

Note that `bind` produces a new function, and when calling `removeEventListener`, the same function reference must be passed as was padded to `addEventListener`, which is why the `bind` is done once in the `constructor`.
```javascript
constructor() {
    super();
    this._handleResize = this._handleResize.bind(this);
}

connectedCallback() {
    super.connectedCallback();
    window.addEventListener('resize', this._handleResize);
}

disconnectedCallback() {
    super.disconnectedCallback();
    window.removeEventListener('resize', this._handleResize);
}
```

dbatiste: Note: alternatively, a simple elegant solution can be to use `matchMedia` with the desired media query, and wire up to the change event.  It may be more performant than wiring up to resize for this case, since we seem to just be interested in a crossing a threshold.

Ref:
https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia
https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList

Be careful of the change in spec for the change event.  Be sure to test all browsers.

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +38,61 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+			</d2l-list-item>
+		`;
+	}
+
 	render() {
+		if (this._isMobile) {
+			return html`
+				${this._renderListItems()}
+				<d2l-dialog-fullscreen ?opened=${this._dialogOpened} title-text=${this.title}
+				@d2l-dialog-close=${this._onListItemClick}>
```
dbatiste, line 91: I don't see `_onListItemClick` here... 🤔 .  Is this working as expected?

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +38,61 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
 	}
 
+	_toggleOpenDialog() {
+		this._dialogOpened = !this._dialogOpened;
+	}
+
+	_renderListItems() {
+		return html`
+			<d2l-list-item class="d2l-list-item"
+				@click=${this._toggleOpenDialog}>
```
dbatiste, line 77: This is an FYI that a future change will be needed here - these list item actions (for the click) aren't keyboard accessible, and there's no styling associated.  There is some `core` work required to make a variant list-item that has button-like semantics.  Once created, you'll be able to wire up for an event, and the component will provide a tab stop and correct styling for hover/focus.

components/right-panel/consistent-evaluation-right-panel.js
dbatiste: I am continuing to explore better solutions to this, to avoid consumers having to use this flag.

components/right-panel/consistent-evaluation-right-panel-block.js
JamieTran: Oops, accidentally left this in from testing

components/right-panel/consistent-evaluation-right-panel-block.js
JamieTran: Thanks for the heads up

jeroach:
Just noticed this now lol

> Dialogs opened from other dialogs are now supported thanks to @jeroach and @dbatiste

It was all Dave - I'm merely a messenger who doesn't deserve credit

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +35,72 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+		this._dialogOpened = false;
 		this.noTitle = false;
+		this.mobileMediaQuery = window.matchMedia('(max-width: 767px)');
+		this._handleResize(this.mobileMediaQuery.matches);
+		this._handleResize = this._handleResize.bind(this);
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.mobileMediaQuery.addEventListener('change', this._handleResize);
```
JamieTran, line 58: This method isn't supported by safari, I wonder if i should be using `window.addEventListener('change', this._handleResize);` instead

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -22,21 +35,72 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.mobileMediaQuery.removeEventListener('change', this._handleResize);
```
JamieTran, line 63: This method isn't supported by safari, I wonder if i should be using `window.removeEventListener('change', this._handleResize);` instead

components/right-panel/consistent-evaluation-right-panel-block.js
JamieTran: _Prior to Safari 14, MediaQueryList is based on EventTarget, so you must use `addListener() `and `removeListener() `to observe media query lists._ The issue is they are both deprecated functions

dbatiste: If necessary, it should be possible to check...
```
if (this.mobileMediaQuery.addEventListener) {
    // use preferred API
} else {
    // use deprecated API
}
```

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -24,19 +37,77 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+		super.connectedCallback();
+		if (this.mobileMediaQuery.addEventListener) {
+			this.mobileMediaQuery.addEventListener('change', this._handleResize);
+		} else {
+			this.mobileMediaQuery.addListener(this._handleResize);
+		}
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
```
AlexBedley, line 66: I usually see this at the end of the function, haven't been burned yet either way though

AlexBedley (approved):
this looks awesome, I'm super happy about the size and scope of the change. well done!

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -10,6 +14,15 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 			noTitle: {
 				attribute: 'no-title',
 				type: Boolean
+			},
+			supportingInfo: {
```
jeroach, line 18: nit: it looks like we're just adding `supportingInfo` but not actually implementing it's usage.

I'm fine with leaving it since I assume that's the next piece you're working on @JamieTran and I expect we'll see that in a follow up PR.

jeroach (approved):

components/right-panel/consistent-evaluation-right-panel-block.js
JamieTran: Yup, that is the next step!

PR #157, opened 2020-11-05
Save coa-eval-override component prior to publishing evaluation

max-hao-d2l:
Related: https://github.com/Brightspace/outcomes-level-of-achievement-ui/pull/58
Unlike the achievements selector in the `consistent-evaluation-outcomes` component, the achievements selector in the `coa-eval-override` component does not autosave, and fires an event with the corresponding "select" action instead.
Changes:
* Add a listener when an outcome in `coa-eval-override` is selected which "unsaves" the evaluation
* Prior to saving/updating/publishing an evaluation, publish the COA demonstration (if applicable)

This should not change any functionality in any tool other than in Mastery View.

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -179,6 +179,20 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		this.unsavedChangesHandler = this._confirmUnsavedChangesBeforeUnload.bind(this);
 	}
 
+	get coaDemonstrationHref() {
+		return this._coaDemonstrationHref;
+	}
+
+	set coaDemonstrationHref(val) {
+		const oldVal = this.coaDemonstrationHref;
+		if (oldVal !== val) {
+			this._coaDemonstrationHref = val;
+			if (this._coaDemonstrationHref && this._token) {
+				this._initializeController();
+			}
+		}
+	}
+
 	get evaluationEntity() {
 		return this._evaluationEntity;
 	}
@@ -201,20 +215,6 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 			this._evaluationHref = val;
 			if (this._evaluationHref && this._token) {
 				this._initializeController().then(() => this.requestUpdate());
-			}
-		}
-	}
-
-	get coaDemonstrationHref() {
-		return this._coaDemonstrationHref;
-	}
-
-	set coaDemonstrationHref(val) {
-		const oldVal = this.coaDemonstrationHref;
-		if (oldVal !== val) {
-			this._coaDemonstrationHref = val;
-			if (this._coaDemonstrationHref && this._token) {
-				this._initializeController();
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


components/consistent-evaluation-page.js
```diff
@@ -220,7 +224,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 
 	get _feedbackText() {
 		if (this._feedbackEntity && this._feedbackEntity.properties) {
-			return this._feedbackEntity.properties.html || '';
```
max-hao-d2l, line 223: Unrelated fix: The html field in feedbackEntity in Mastery View is `null`, but we can get text instead.

rgoessl (approved):

components/consistent-evaluation-page.js
jmurray-d2l: We should probably look into fixing this in the back end.  If there is any styling applied to the feedback in the editor then it won't be returned with `text`

jmurray-d2l:
@AlexBedley Did a bunch of work around the timing of transient saving, looks like you're trying to something similar.  I'd suggest getting his thoughts on this PR before merging

AlexBedley:
> Did a bunch of work around the timing of transient saving, looks like you're trying to something similar. I'd suggest getting his thoughts on this PR before merging

Spoke offline with Max. Gonna look into adding transient save in the API. My fear around doing it this way is that we're moving in the complete opposite direction - the front-end will know less and less as we refactor and just get stuff from the API. So the hack in this PR to make the evaluation unsaved might just be the tip of the iceberg of stuff you'll have to do to make it appear like it's transiently saving.

gfrancu (approved):

max-hao-d2l:
Given the time constraint, it is probably not feasible to add a new evaluation entity to the evaluation. I've made it so that the COA component autosaves by itself (see https://github.com/Brightspace/outcomes-level-of-achievement-ui/pull/59), and perform a "publish" action on the corresponding entity upon publish/save/update. This prevents us to cache the save action in the frontend.
In the future, we could investigate on making the transient-save and publish action to be similar to Grades and Feedback, as it is most likely cleaner and easier to work with
(Gonna need to fix these failing tests though)

components/controllers/ConsistentEvaluationController.js
```diff
@@ -131,6 +146,7 @@ export class ConsistentEvaluationController {
 			throw new Error(ConsistentEvaluationControllerErrors.INVALID_EVALUATION_ENTITY);
 		}
 
+		if (this.coaDemonstrationHref) await this.saveCoaDemonstration();
```
AlexBedley, line 149: nit: i know it's probably the linter's responsibility, but I really think all if-blocks should have brackets `{}`

components/consistent-evaluation-page.js
max-hao-d2l: I've reverted this change and made the fix in the backend (https://github.com/Brightspace/lms/pull/3185)

jmurray-d2l: Nice work Max!

jmurray-d2l (approved):
LGTM! Would still suggest we wait for Alex to give the green light on this, but up to you.

AlexBedley (approved):

PR #158, opened 2020-11-05
US122162 - Selecting an unsupported file in the LCB submission list should give user option to download

amauryperez:
https://files.slack.com/files-pri/T025EV233-F01DNP6G0B1/image.png


components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -128,6 +134,10 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 		if (this.submissionInfo.submissionType === fileSubmission) {
 			this.fileEvidenceUrl = currentFile.properties.fileViewer;
 			this.textEvidence = undefined;
+			this.fileNonViewable = {
```
jeroach, line 137: So in the case that a file would not be viewable, `fileEvidenceUrl` would be `undefined` ?

And thus we'd skip over the if block on line 232 - https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/158/files#diff-896d584e29e3574c5af99431a6bcb97df335109651d39dee09230d2f33a3e870R232

And fall into the newly added if block? Am I understanding that correctly?

amauryperez: Exactly. The `fileViewer` property wont be present. Maybe I can add a comment.

jeroach (approved):

components/left-panel/consistent-evaluation-left-panel.js
AlexBedley: Ya I think adding something like

```js
this.textEvidence = undefined;
if (currentFile.properties.fileViewer) {
  // file is viewable
  this.fileEvidenceUrl = currentFile.properties.fileViewer;
} else {
  // file is unviewable
  this.fileNonViewable  = {...};
}
```

would make it slightly more clear

AlexBedley (approved):

PR #163, opened 2020-11-09
- Removing manual scrollbar changes

jmurray-d2l:


AlexBedley:
this is awesome cuz now i can clean up the select file stuff. did we fix some css somewhere?

jmurray-d2l:
unfortunately I don't think will for us as it, I put this up to get thoughts from Josh since I know he implemented this for a reason.  Locally I saw no issues, but there is a double scrollbar issue with the annotations viewer

jmurray-d2l:
Okay so is seems as though there was some margin being applied to the iframe as well as some border stuff?  The good news is that this fixes the double scroll bar issue so we don't have to manage the scroll bars ourselves.  The bad news is that I can't fully explain why these exact numbers solve the problem (other numbers work as well, these are just the smallest values that work for margin-bottom and the height calculation)

AlexBedley:
hmm I don't fully understand the issue so I'll wait until @jeroach chimes in

jeroach:
hey guys, just catching up on this. For context, here is the original issue I found, and the subsequent feature added to `primary-secondary`

https://d2l.slack.com/archives/C0PHG3QB0/p1594925848243600

The gist of the issue was this - in the case that we're putting something with it's own scrollbars into the primary panel (e.g. the Annotations viewer) - and this content was larger than the primary panel view - the primary panel would add it's own scrollbars, thus causing double scrollbars.

So in the case that we're embedding something with scrollbars into primary, we can then set `primary-overflow: hidden` to prevent double scrollbars.

jmurray-d2l:
That makes sense.  The css changes I've made here seem to fix the issue with the annotations viewer content being larger than the primary panel view, which stops the primary panel from adding its own scroll bars

jeroach (approved):

PR #165, opened 2020-11-10
- Additional design css updates

jmurray-d2l:


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -65,7 +65,7 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 			display: flex;
 			flex-direction: column;
 			margin: 0.5rem 0.65rem;
-			padding: 0 1.2rem 0.75rem 1.2rem;
+			padding: 1px 1.2rem 0.75rem 1.2rem;
```
jmurray-d2l, line 68: `d2l-list-item` has a margin-top of -1px which was causing the top border of the submission item to be overlapped, this fixes that

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -31,8 +31,6 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 		return [labelStyles, css`
 			.d2l-block {
 				margin-top: 1.35rem;
-				padding-left: 0.75rem;
```
jmurray-d2l, line 34: Since I added the div around the entire right panel and added margins to that div, we no longer need this padding

jeroach (approved):

PR #166, opened 2020-11-11
Nullcheck COA publish action

max-hao-d2l:
In the COA Manual Override page, when only publishing feedback with an achievement selector, the evaluation fails to publish since publish action is null and there is no nullcheck.

Changes:
* Add a nullcheck to `publishAction` in saveCoaDemonstration

### DIFF
--- a/components/controllers/ConsistentEvaluationController.js
+++ b/components/controllers/ConsistentEvaluationController.js
@@ -188,6 +188,8 @@ export class ConsistentEvaluationController {
 	async saveCoaDemonstration() {
 		const coaDemonstrationEntity = await this.fetchCoaDemonstrationEntity(false);
 		const publishAction = coaDemonstrationEntity.getActionByName('publish');
-		return await this._performSirenAction(publishAction);
+		if (publishAction) {
+			return await this._performSirenAction(publishAction);
+		}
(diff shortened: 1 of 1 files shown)


components/controllers/ConsistentEvaluationController.js
```diff
@@ -188,6 +188,8 @@ export class ConsistentEvaluationController {
 	async saveCoaDemonstration() {
 		const coaDemonstrationEntity = await this.fetchCoaDemonstrationEntity(false);
 		const publishAction = coaDemonstrationEntity.getActionByName('publish');
-		return await this._performSirenAction(publishAction);
+		if (publishAction) {
```
jeroach, line 191: Does anything need to be done/alerted/logged in case of the `publishAction` being null?

rgoessl (approved):

components/controllers/ConsistentEvaluationController.js
max-hao-d2l: No, the demonstration entity containing the action doesn't have a `publish` action when nothing in the achievements selector is selected and we want to only save feedback, so it's not an error condition.

jeroach (approved):

PR #167, opened 2020-11-11
Add mobile drawer summaries to right-panel components

JamieTran:
Added summary information for components in the mobile drawer to match the following specs:
image

The preview:
image


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -103,6 +106,18 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeConsistentEvalu
...
 	}
 
+	_setFeedbackSummaryInfo() {
+		let summary = '';
+		if (this.feedbackText === null || this.feedbackText === '') {
+			summary = this.localize('noFeedbackSummary');
+		} else {
+			const tmpDiv = document.createElement('div');
+			tmpDiv.innerHTML = this.feedbackText;
+			summary = tmpDiv.textContent || tmpDiv.innerText || '';
```
JamieTran, line 116: This looks funny but `this.feedbackText` contains formatted text so tags like `<p>` and `<strong>` would appear in the summary if we didn't get rid of it.

jeroach (approved):
LGTM :)

lang/en.js
```diff
@@ -44,6 +48,7 @@ export default {
 	"retract": "Retract",
 	"retracted":   "Evaluation retracted",
 	"rubrics": "Rubrics",
+	"rubricSummary": "1 rubric attached",
```
jmurray-d2l, line 51: nit: consider parameterizing the number so this won't have to be changed when we add more rubrics

jmurray-d2l (approved):

PR #168, opened 2020-11-11
Abedley/annotations/unsaved confirmation

AlexBedley:
**Problem:** when navigating away from your current file that has annotations, they are seemingly discarded without warning, but they are actually transiently saved which is very confusing and causes some weird edge cases.
**Solution:** add a confirmation dialog to alert the user that they have unsaved annotations. if they decide to discard their changes, then they are removed from the transient state

image

I have 2 fast follows for this PR:

1. unify show submissions and select file. previously they would do scrollbar things, but now they're pretty much the same. gonna add that refactor next
2. handle resetting the file selector if the user decides not to discard their changes


components/consistent-evaluation-page.js
```diff
@@ -528,13 +532,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 	}
 
 	_selectFile(e) {
-		this.currentFileId = e.detail.fileId;
+		window.dispatchEvent(new CustomEvent('d2l-flush', {
```
AlexBedley, line 536: we must flush here since the annotations changes are debounced

components/consistent-evaluation-page.js
```diff
@@ -528,13 +532,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+		}
+		this._changeFile(newFileId);
 	}
 
-	_setSubmissionsView() {
+	_changeFile(newFileId) {
+		this.currentFileId = newFileId;
+	}
+
+	async _setSubmissionsView() {
```
AlexBedley, line 552: this function and `_selectFile`  are basically the same. the first fast follow is to unify them

components/consistent-evaluation-page.js
```diff
@@ -528,13 +532,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+		}
+		this.displaySubmissions();
+	}
+
+	displaySubmissions() {
 		this.currentFileId = undefined;
 	}
 
+	async _checkUnsavedAnnotations(newFileId) {
+		return await this._mutex.dispatch(
```
AlexBedley, line 570: this must be protected by the mutex, otherwise we could still be in the process of saving annotations

components/consistent-evaluation-page.js
```diff
@@ -528,13 +532,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+			async() => {
+				if (this.currentFileId !== undefined) {
+					const annotationsEntity = this.evaluationEntity.getSubEntityByRel('annotations');
+					const unsavedAnnotations = annotationsEntity.hasClass('unsaved');
+					if (unsavedAnnotations) {
+						this.nextFileId = newFileId;
```
AlexBedley, line 576: unfortunately we need to add this "global" variable to pass state between here and the dialog-closed handler

components/consistent-evaluation-page.js
```diff
@@ -528,13 +532,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+		this._unsavedAnnotationsDialogOpened = false;
+		if (e.detail.action === DIALOG_ACTION_DISCARD) {
+			await this._discardAnnotationsChanges();
+			if (this.nextFileId) {
+				this._changeFile(this.nextFileId);
+			} else {
+				this.displaySubmissions();
+			}
+		} else {
+			// need to reset the file selector
```
AlexBedley, line 596: this is the 2nd fast follow. I have a solution for this, but it's kinda hacky, so I want the diff to be clean when we review it

components/consistent-evaluation-page.js
```diff
@@ -528,13 +532,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+		}
+	}
+
+	async _discardAnnotationsChanges() {
+		await this._mutex.dispatch(
```
AlexBedley, line 601: hmm this probably doesn't need to be protected by the mutex since the dialog prevents you from doing anything else (unless you use a screen reader, but that's another bug haha). however, it probably doesn't hurt to be here either and could prevent a future bug. I could take this out. thoughts?

components/controllers/ConsistentEvaluationController.js
```diff
@@ -141,6 +141,13 @@ export class ConsistentEvaluationController {
 		return await performSirenAction(this.token, saveAnnotationsAction, fields, true);
 	}
 
+	async transientDiscardAnnotations(evaluationEntity) {
+		const annotationsEntity = evaluationEntity.getSubEntityByRel('annotations');
+
+		// explicitly send `value` as empty string to avoid 400 missing param
+		return await this._performAction(annotationsEntity, 'RemoveAnnotations', 'value', '');
```
AlexBedley, line 148: I think the siren action stuff has logic to delete empty fields. Well the LMS doesn't like that at all, it will give 400 if the empty field is missing. so this code explicitly sets it

lang/en.js
```diff
@@ -55,6 +55,9 @@ export default {
 	"tryDonwloadFile": "Try downloading the file instead",
 	"unevaluated": "Unevaluated",
 	"unflag": "Unflag",
+	"unsavedAnnotationsBody": "Are you sure you want to leave without saving your annotation changes?",
+	"unsavedAnnotationsDiscardButton": "Discard annotation changes",
+	"unsavedAnnotationsTitle": "Unsaved annotation changes",
```
AlexBedley, line 60: I vetted these langterms with Dan

components/consistent-evaluation-page.js
```diff
@@ -528,13 +533,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+		const newFileId = e.detail.fileId;
+		const shouldSelectFile = this._checkUnsavedAnnotations(newFileId);
+		if (!shouldSelectFile) {
+			return;
+		}
+		this._changeFile(newFileId);
 	}
 
-	_setSubmissionsView() {
+	_changeFile(newFileId) {
```
jeroach, line 548: Does this still need to be a separate method?

components/consistent-evaluation-page.js
```diff
@@ -528,13 +533,79 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+	displaySubmissions() {
 		this.currentFileId = undefined;
 	}
 
+	async _checkUnsavedAnnotations(newFileId) {
+		return await this._mutex.dispatch(
+			async() => {
+				if (this.currentFileId !== undefined) {
+					const annotationsEntity = this.evaluationEntity.getSubEntityByRel('annotations');
+					const unsavedAnnotations = annotationsEntity.hasClass('unsaved');
```
jeroach, line 574: Do we need a null check on `annotationsEntity` ?

components/consistent-evaluation-page.js
jeroach: `nextFileId` implies iteration - is that strictly the case? The name sounds a bit confusing to me, but I might not be understanding correctly :)

components/consistent-evaluation-page.js
jeroach: I'm good with the extra safety - I assume it doesn't cost us much

jeroach (commented):
Overall looks good, just a few questions

components/consistent-evaluation-page.js
jeroach: I think I did see it used elsewhere, so I think this is needed.

components/consistent-evaluation-page.js
AlexBedley: no iteration here, `nextFileId` is the file that will be shown if the user decides to discard their annotations changes (which is not the same as `currentFileId`). happy to change the name though

components/consistent-evaluation-page.js
AlexBedley: yes i needed a function to change the file and bypass the confirmation message. this will be merged with set submissions view in the next PR

components/consistent-evaluation-page.js
AlexBedley: I'll add one

components/controllers/ConsistentEvaluationController.js
jmurray-d2l: We chatted about creating and explicit clear action on the API, I guess we never implemented that?  Would probably make this easier

jmurray-d2l (approved):

components/controllers/ConsistentEvaluationController.js
AlexBedley: This is the explicit clear action, it's just a bit picky about which query params need to be present. This is inline with my experience writing controllers in the LMS, they are very specific about the qps

components/consistent-evaluation-page.js
jeroach: Ah it's fine

jeroach (approved):
Approved assuming the null check is coming :)

AlexBedley:
added the null checks

PR #173, opened 2020-11-12
Jroach/us122359

jeroach:
[US122359](https://rally1.rallydev.com/#/detail/userstory/456046385436?fdp=true): Unable to keyboard navigate into the file names to select them in the Evidence panel

Keyboard focus:
image

### DIFF
--- a/components/left-panel/consistent-evaluation-submission-item.js
+++ b/components/left-panel/consistent-evaluation-submission-item.js
@@ -100,6 +100,9 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 			position: relative;
 			top: 0;
 		}
+		.d2l-submission-attachment-list-item-content:hover {
+			cursor: pointer;
+		}
 		.d2l-attachment-read-status {
 			color: var(--d2l-color-carnelian);
 			position: absolute;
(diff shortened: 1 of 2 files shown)


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -71,6 +71,11 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 		d2l-list-item, d2l-list-item:hover {
 			--d2l-list-item-content-text-color: var(--d2l-color-ferrite);
 		}
+		d2l-list-item-content:focus {
```
jeroach, line 74: This is the same styling used for other items with focus (drop down list, etc.)

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -208,6 +213,20 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 		return Math.max(fileSizeBytes, 0.1).toFixed(1) + unit;
 	}
 
+	_dispatchFileSelectedKeyboardEvent(e) {
+		const fileId = e.target.getAttribute('file-id');
+
+		if (e.key === 'Enter' || e.key === ' ') {
```
jeroach, line 219: I confirmed with Peggy that we want the navigation to trigger on either `Enter` or `Space`

dbatiste:
I think what we really want here is the component that I have in my backlog to create - a list item whose primary action is a button, which dispatches an event.  Then there won't be any of this monkey business with manually adding tabindex or providing custom styles.  I think I've seen a couple places where list-items are being used in a button-like way, that are not keyboard accessible.

JamieTran (approved):
LGTM

AlexBedley (approved):

PR #174, opened 2020-11-13
design updates to mobile adaptive

jeroach:
[US122407](https://rally1.rallydev.com/#/detail/userstory/456565213984?fdp=true): Alignment work in Mobile view

Updated:
image

### DIFF
--- a/components/right-panel/consistent-evaluation-right-panel-block.js
+++ b/components/right-panel/consistent-evaluation-right-panel-block.js
@@ -36,8 +36,14 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 				margin-bottom: 0.4rem;
 			}
 			.d2l-list-item {
-				padding-left: 0.75rem;
-				padding-right: 0.75rem;
+				overflow: hidden;
+			}
+			.d2l-list-item-content {
+				padding-left: 1.25rem;
+			}
+			.d2l-list-item-content([dir="rtl"]) {
+				padding-left: 0;
+				padding-right: 1.25rem;
 			}
 			.d2l-truncate {
 				overflow: hidden;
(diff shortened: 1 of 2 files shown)


components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -36,8 +36,14 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 				margin-bottom: 0.4rem;
 			}
 			.d2l-list-item {
-				padding-left: 0.75rem;
-				padding-right: 0.75rem;
+				overflow: hidden;
```
jeroach, line 39: This prevents scrollbars from showing up sometimes in the mobile drawer

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -36,8 +36,14 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
...
+			}
+			.d2l-list-item-content {
+				padding-left: 1.25rem;
```
jeroach, line 42: Same padding as the profile image

components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -84,7 +90,7 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 	}
 
 	_getTitle() {
-		return this.noTitle ? html`` : html`<div class="d2l-label-text">${this.title}</div>`;
+		return this.noTitle ? html`` : html`<h2 class="d2l-label-text">${this.title}</h2>`;
```
jeroach, line 93: This was missed in the headers story - just adding here since it's one line in a file we're already changing.

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -93,7 +93,7 @@ export class ConsistentEvaluationRightPanel extends LocalizeConsistentEvaluation
 
 			@media (max-width: 767px) {
 				.d2l-consistent-evaluation-right-panel {
-					margin: 1.5rem 0.9rem 0 0.9rem;
+					margin: 0 0 0 0;
```
JamieTran, line 96: nit: I think this can be changed to just `margin: 0;`

JamieTran (approved):

components/right-panel/consistent-evaluation-right-panel.js
jeroach: Fixed

AlexBedley (approved):

PR #175, opened 2020-11-13
Add in not great way to refresh the file selector after a declined discard of annotations

AlexBedley:
As promised, here's follow up 2 from the previous PR. This adds a super hacky way of keeping the file selector in-sync with the currently selected file. The problem is the UI change of selecting a different file changes the UI state, but since the `currentFileId` does not change, Lit does not re-render. Thus we need to manually tell it "hey actually we need you to display the file that corresponds to your `currentFileId`"

### DIFF
--- a/components/consistent-evaluation-page.js
+++ b/components/consistent-evaluation-page.js
@@ -594,6 +594,9 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 			}
 		} else {
 			// need to reset the file selector
+			await this.shadowRoot.querySelector('d2l-consistent-evaluation-learner-context-bar')
+				.shadowRoot.querySelector('d2l-consistent-evaluation-lcb-file-context')
+				.refreshSelect();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


components/consistent-evaluation-page.js
```diff
@@ -594,6 +594,9 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 			}
 		} else {
 			// need to reset the file selector
+			await this.shadowRoot.querySelector('d2l-consistent-evaluation-learner-context-bar')
+				.shadowRoot.querySelector('d2l-consistent-evaluation-lcb-file-context')
```
AlexBedley, line 598: this is the hacky part, not really a fan of going 2 shadowroots deep. If there's a way we can tell a lit component "i know your properties didn't change but I need you to re-render anyways" then I would prefer that. unfortunately all the permutations of `requestUpdate` didn't do it for me

jmurray-d2l (approved):

components/consistent-evaluation-page.js
AlexBedley: hopefully we can address this tech debt in M2

PR #178, opened 2020-11-13
- Fix issue with double scroll bar

jmurray-d2l:
This is clearly a bit too fragile, might be worth revisiting later.


jeroach (approved):
Agreed on fragility

PR #184, opened 2020-11-24
Add required logic so that tests can run

jeroach:
As per this thread: https://d2l.slack.com/archives/C0PHG3QB0/p1606252808231200


JamieTran (approved):
It's interesting as to why this is needed on certain machines. Tested this on my machine and everything still runs properly 👍

AlexBedley (approved):

PR #185, opened 2020-11-27
Vy/add telemetry client

vix633:
I'd like to get some input on this. Is the API clear? Would you know how to measure timing of something if you saw this pr?


jmurray-d2l:
Cool! I think I understand the API, looks easy to use - One question, what would happen if you tried to log an event with out setting the markEventStart first?

vix633:
> Cool! I think I understand the API, looks easy to use - One question, what would happen if you tried to log an event with out setting the markEventStart first?

If there is no start mark, it behaves as if the start mark was the page load start. In the example I included, the start mark not actually needed, its more a documentation piece; a lie. I will remove it as I add other monitoring events that organically demonstrate the complete startMark/log behaviour.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -209,6 +213,7 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 	}
 
 	_finishedLoading() {
+		this._telemetry.markEventEndAndLog(undefined, this._perfRenderEventName);
```
jmurray-d2l, line 216: What is undefined here?

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -133,6 +134,9 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 		this._submissionList = [];
 		this._token = undefined;
 		this._submissionEntities = [];
+		this._perfRenderEventName = 'submissionsComponentRender';
+		this._telemetry = new ConsistentEvalTelemetry();
```
jmurray-d2l, line 138: Does this need the endpoint?

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -99,6 +104,11 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 		`];
 	}
 
+	constructor() {
```
jmurray-d2l, line 107: Probably don't need this if we aren't logging anything in this file

components/helpers/consistent-eval-telemetry.js
```diff
@@ -0,0 +1,56 @@
...
+		performance.clearMarks(eventStartMarkName);
+		performance.mark(eventStartMarkName);
+	}
+
+	_getEventStartMarkName(type) {
+		return `d2l-consistent-eval-event-${type}`;
+	}
+
+	async _logUserEvent(href, action, type, performanceMeasureName) {
+		if (!href || !action || !type || !performanceMeasureName) return;
```
jmurray-d2l, line 41: I think we use brackets even on one line if statements.

components/consistent-evaluation.js
```diff
@@ -65,6 +70,8 @@ export class ConsistentEvaluation extends LitElement {
 			main : true,
 			submissions: true
 		};
+		this._telemetry = new ConsistentEvalTelemetry(this._dataTelemetryEndpoint);
+		this._perfRenderEventName = 'consistentEvalMain';
```
jmurray-d2l, line 74: nit: I think we could just have this as a constant near its only usage, or even inline as a param since its only used once

components/helpers/consistent-eval-telemetry.js
```diff
@@ -0,0 +1,56 @@
...
+
+	constructor(dataTelemetryEndpoint) {
+		this._dataTelemetryEndpoint = dataTelemetryEndpoint;
+	}
+
+	//Mark that a page has been loaded
+	logLoadEvent(type) {
+		if (!type) return;
+
+		const measureName = `d2l-consistent-eval-${type}.page.rendered`;
```
jmurray-d2l, line 13: just out of curiosity - is `.page.rendered` a key phrase?  is it marked by default somehow?

components/helpers/consistent-eval-telemetry.js
```diff
@@ -0,0 +1,56 @@
...
+		performance.measure(measureName);
+		this._logUserEvent(window.location.hostname, 'LoadView', type, measureName);
+	}
+
+	//Submit an event measure
+	markEventEndAndLog(type) {
+		if (!type) return;
+
+		const measureName = `d2l-consistent-eval-event-${type}`;
```
jmurray-d2l, line 22: nit: this is the same value as `eventStartMarkName`, not sure if thats intentional

components/left-panel/consistent-evaluation-submissions-page.js
vix633: Nothing, missed in cleanup!

components/helpers/consistent-eval-telemetry.js
vix633: No, its a pattern i got from Steven Ao

components/helpers/consistent-eval-telemetry.js
vix633: I think them being the same is okay? I'm going add `mark` to the name for clarity

components/consistent-evaluation.js
vix633: That's fair

components/left-panel/consistent-evaluation-submissions-page.js
vix633: I've added it. Odd that it still fired events to the qa endpoint without that being there.

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -56,10 +55,6 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 			currentFileId: {
 				attribute: 'current-file-id',
 				type: String
-			},
-			_dataTelemetryEndpoint: {
```
jmurray-d2l: We'll still need this if we want to pass it to the submissons page

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -34,6 +34,10 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 						return retVal;
 					}
 				}
+			},
+			_dataTelemetryEndpoint: {
```
jmurray-d2l, line 38: We aren't passing this in here, I'm confused how its working, maybe it doesn't need the endpoint?  We could also consider just passing around the telemetry object, I'd like to hear what @AlexBedley  thinks

vix633: I think because its on the page already it gets bound here. Otherwise... i have no idea how its properly attempting to send the events. I've tested changing the endpoint to something distinct to ensure it does actually get picked up by the client, and it did.

components/consistent-evaluation.js
```diff
@@ -136,6 +142,7 @@ export class ConsistentEvaluation extends LitElement {
 			}
 		}
 		this._loading = false;
+		this._telemetry.logLoadEvent('consistentEvalMain');
```
AlexBedley, line 145: Note that this will fire on every iteration. Once the caching is in place, subsequent loads will be much faster and that might skew the data. we might want to differentiate between "first load" and loading from cache

components/left-panel/consistent-evaluation-submissions-page.js
AlexBedley: We are passing it in? I think I'm missing something.

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/603aca98432223063e4c8c74c5715b9b5b20d5ab/components/consistent-evaluation.js#L170

components/consistent-evaluation.js
AlexBedley, line 145: note that this will fire on every iteration. After caching is in-place, subsequent loads will be much faster than initial loads. We might want to differentiate between initial load and loading from cache. Unfortunately, I don't think EntityStore tell us when there's a cache hit, so we'd have to keep track here

AlexBedley: ah reading more, it looks like here we're just counting the number of times the page is loaded, not the actual load time (correct me if i'm wrong). In that case ignore my comment (but it might apply to submissions-page where it looks like we are timing things)

AlexBedley (approved):
Looks good. I'd consider unit testing `ConsistentEvalTelemetry`

components/consistent-evaluation.js
vix633: This actually implicitly measures from navigationStart to loaded time.

vix633: I think cache loads are still loads. Do you have any suggestions how to differentiate between a first load and a cache hit?

components/left-panel/consistent-evaluation-submissions-page.js
jmurray-d2l: I was referring to the `consistent-evaluation-submissions-page` does it not need to be passed down through?

AlexBedley: oh duhhh that's what I was missing haha. I saw a file ending in `page` and just assumed it was the other one

AlexBedley: my guess is `D2L.Telemetry` contains a default, which just happens to match up with what we want. Let's not rely on that and just pass it down normally

components/helpers/consistent-eval-telemetry.js
```diff
@@ -0,0 +1,56 @@
...
+		const eventBody = new Events.PerformanceEventBody()
+			.setAction(action)
+			.setObject(href, type)
+			.addUserTiming(performance.getEntriesByName(performanceMeasureName));
+		const event = new Events.TelemetryEvent()
+			.setType('PerformanceEvent')
+			.setDate(new Date())
+			.setSourceId('consistent-eval')
+			.setBody(eventBody);
+		const client = await D2L.Telemetry.CreateClient({endpoint: this._dataTelemetryEndpoint});
```
AlexBedley, line 52: It seems a bit fragile to rely on a global here. Let's just use `new Events.Client`. This is probably also the source of the confusion on why it works without passing in an endpoint (like why is this function `async`? what is it doing?)

components/consistent-evaluation.js
AlexBedley: off the top of my head, retain a `Set` of hrefs that have been loaded, and check that to determine if it's first load or not.

Note that you'll have to switch to your other load time tracking mechanism (where you explicitly start the load)

AlexBedley: maybe `EntityStory.get` returns something to signify it was a cache miss, but that seems onerous

components/left-panel/consistent-evaluation-submissions-page.js
AlexBedley: ah yes, we're using some weird BSI global that tries to guess your endpoint. As I mentioned in the other comment, let's avoid doing that and just do it normally.

https://github.com/Brightspace/brightspace-integration/blob/99925835df171dc0656cd53e2e6129b34973d956/web-components/bsi-unbundled.js#L41

vix633: I'm sort of leaning towards just using the global and releasing our control over the endpoint to BSI. That's there for a reason and we don't need to override it.

vix633: Clearly keeping in mind that its getting the same attribute internally anyway, i think we should remove the telemetry endpoint references everywhere since its handled at the telemetry client level.

AlexBedley: I'm not a fan of using the global. Since the file is named "bsi-unbundled" I worry that it would not be present in the production bundle and is only in the unbundled bsi to support some weird edge case. Additionally, this just further ties our application to BSI (all the demo pages/vdiff tests will throw errors since they do not know about BSI). And lastly, the usage of the global will be annoying to stub in unit tests.

vix633: These are good points. I think my main leaning here is that its already implemented here and I like the idea of handing over control of utility stuff to the overarching UI framework we have. I asked Dave B to have a quick look over this discussion and maybe contribute some input on whether or not globals like this are a good idea to use or not. In the mean time, I've touched on the implementation of passing through the attribute, and not using the global client and stepped through it, it looks like our attribute is not getting bound for some reason :/

components/consistent-evaluation.js
```diff
@@ -92,6 +97,7 @@ export class ConsistentEvaluation extends LitElement {
 			}
 
 			this._loadingComponents.main = false;
+			this._telemetry = new ConsistentEvalTelemetry(this.dataTelemetryEndpoint);
```
AlexBedley, line 100: this shouldn't be tied to `href` changes, it only needs to change if `this.dataTelemetryEndpoint` is `updated`

components/consistent-evaluation.js
```diff
@@ -136,6 +142,7 @@ export class ConsistentEvaluation extends LitElement {
 			}
 		}
 		this._loading = false;
+		this._telemetry.logLoadEvent('consistentEvalMain', this._submissionInfo.submissionList.length);
```
AlexBedley, line 145: there's a few tricks you can do to avoid null pointer exceptions here, might be worth refactoring to not crash the app if telemetry is missing

components/helpers/consistent-eval-telemetry.js
```diff
@@ -0,0 +1,59 @@
...
+			.addUserTiming(performance.getEntriesByName(performanceMeasureName));
+		if (submissionCount) {
+			eventBody.addCustom('SubmissionCount', `${submissionCount}`);
+		}
+		const event = new Events.TelemetryEvent()
+			.setType('PerformanceEvent')
+			.setDate(new Date())
+			.setSourceId('consistent-eval')
+			.setBody(eventBody);
+		const client = new Events.Client({endpoint: this._dataTelemetryEndpoint});
```
AlexBedley, line 55: might be safer if you do this in the constructor, instead of making a client per log message (not sure if creating a client is expensive or not)

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -163,6 +169,9 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 	}
 
 	async _initializeSubmissionEntities() {
+		this._telemetry = new ConsistentEvalTelemetry(this.dataTelemetryEndpoint);
```
AlexBedley, line 172: same as the other comment, this belongs in `updated` and should change when `dataTelemetryEndpoint` changes

components/helpers/consistent-eval-telemetry.js
vix633: This seems a lot like a 'might' rather than concrete change. I'll leave this to the GC.

components/consistent-evaluation.js
```diff
@@ -163,6 +173,8 @@ export class ConsistentEvaluation extends LitElement {
 				return-href=${ifDefined(this.returnHref)}
 				return-href-text=${ifDefined(this.returnHrefText)}
 				current-file-id=${ifDefined(this.currentFileId)}
+				data-telemetry-endpoint=${ifDefined(this.dataTelemetryEndpoint)}
+				.rubricHrefs=${this._childHrefs && this._childHrefs.rubricHrefs}
```
AlexBedley, line 177: bad merge? I think this was deleted in the rubrics PR

AlexBedley (approved):

components/consistent-evaluation.js
jmurray-d2l: this line should be removed, probably from bad merge conflict resolution

jmurray-d2l: whoops, way behind - sorry about that

jmurray-d2l (approved):

PR #186, opened 2020-11-30
- Add support for multiple rubrics in consistent evaluation

jmurray-d2l:


JamieTran (approved):
lgtm

components/consistent-evaluation-page.js
```diff
@@ -51,13 +51,9 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
-			rubricAssessmentHref: {
-				attribute: 'rubric-assessment-href',
-				type: String
-			},
-			rubricHref: {
-				attribute: 'rubric-href',
-				type: String
+			rubricHrefs: {
+				attribute: false,
+				type: Object
```
AlexBedley, line 56: should this be `Array`?

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -58,19 +65,24 @@ export class ConsistentEvaluationHrefController {
...
 
-			if (rubricAssessmentHref) {
-				const assessmentEntity = await this._getEntityFromHref(rubricAssessmentHref, bypassCache);
-				if (assessmentEntity && assessmentEntity.entity) {
-					rubricHref = this._getHref(assessmentEntity.entity, rubricRel);
-				}
+			if (rubricHrefs) {
+				rubricHrefs = await Promise.all(rubricHrefs.map(async rubricAssessmentHref => {
+					const assessmentEntity = await this._getEntityFromHref(rubricAssessmentHref, bypassCache);
+					if (assessmentEntity && assessmentEntity.entity) {
```
AlexBedley, line 78: I forget if `fetch` throws on failures or not. But if it does throw, it will cause the entire promise chain to reject.

If it doesn't throw, then you'll end up with `undefined` in your `rubricsHrefs` array (which you probably want to filter out, or display an error)

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -58,19 +65,24 @@ export class ConsistentEvaluationHrefController {
...
+						const rubricHref = this._getHref(assessmentEntity.entity, rubricRel);
+						return {
```
AlexBedley, line 80: nit: shorthand for this is just:

```js
return {
  rubricHref,
  rubricAssessmentHref
};
```

no need to make strings

components/consistent-evaluation-page.js
```diff
@@ -51,13 +51,9 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
 			},
-			rubricAssessmentHref: {
-				attribute: 'rubric-assessment-href',
-				type: String
-			},
-			rubricHref: {
-				attribute: 'rubric-href',
-				type: String
+			rubricHrefs: {
+				attribute: false,
```
AlexBedley, line 55: any particular reason we're doing `attribute: false`?

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -26,19 +23,39 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
...
+		return  css`
+			.d2l-consistent-evaluation-rubric:nth-child(n + 2) {
+				margin-top: 0.7rem;
+			}
+		`;
+	}
+
+	_getRubrics() {
+		const rubrics = [];
+		this.rubricHrefs.forEach(rubric => {
```
AlexBedley, line 36: nit: it's more idiomatic to use `map` instead of `forEach` since you don't care about side-effects

```js
const rubrics = this.rubricHrefs.map(rubric => ...);
```

components/consistent-evaluation-page.js
jmurray-d2l: We were having issues getting this to work without it, but it may have been something else causing it and we didn't revisit after we got it working. I'll try to update this again

components/consistent-evaluation-page.js
jmurray-d2l: Yes.  I didn't know that was a valid lit type, whoops!

components/controllers/ConsistentEvaluationHrefController.js
jmurray-d2l: Tested this, it returns undefined.  I updated the render part to handle undefined

components/consistent-evaluation-page.js
AlexBedley: my guess is because you had the type set to `Object` the parsing would fail. either way, not a deal breaker

AlexBedley (approved):
my code concerns have been addressed, although it looks like some vdiff tests need to be fixed/updated. I heard there's a problem right now with the bot

PR #187, opened 2020-12-01
Using attachments entity instead of link

amauryperez:


components/right-panel/consistent-evaluation-attachments-editor.js
```diff
@@ -2,21 +2,32 @@ import 'd2l-activities/components/d2l-activity-editor/d2l-activity-attachments/d
...
-import { AttachmentCollectionEntity } from 'siren-sdk/src/activities/AttachmentCollectionEntity.js';
-import { AttachmentEntity } from 'siren-sdk/src/activities/AttachmentEntity.js';
-import { Rels } from 'd2l-hypermedia-constants';
 
 class ConsistentEvaluationAttachmentsEditor extends LitElement {
 	static get properties() {
 		return {
 			href: { type: String },
 			destinationHref: { type: String },
 			token: { type: String },
```
AlexBedley, line 11: I think `href`, `token`, and `destinationHref` can be deleted

components/consistent-evaluation-page.js
```diff
@@ -266,6 +283,11 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		this._controller = new ConsistentEvaluationController(this._evaluationHref, this._token);
 		const bypassCache = true;
 		this.evaluationEntity = await this._controller.fetchEvaluationEntity(bypassCache);
+		const attachmentsInfo = await this._controller.fetchAttachments(this.evaluationEntity);
+		this._canAddFeedbackFile = attachmentsInfo.canAddFeedbackFile;
+		this._canRecordFeedbackVideo = attachmentsInfo.canRecordFeedbackVideo;
+		this._canRecordFeedbackAudio = attachmentsInfo.canRecordFeedbackAudio;
+		this._attachments = attachmentsInfo.attachments;
```
AlexBedley, line 290: Having all these derived variables is dangerous. I would prefer we only keep `attachmentsInfo` as a property here and divvy up it's members in `render`. That way we don't need to keep these 4 variables in-sync.

For example, if you notice in `_transientAddAttachment` and `_transientRemoveAttachment` we update `_attachments` but the other 3 properties are now stale. I guess in this case, it's not a huge deal because permissions don't change often. But this pattern has caused major bugs in the past when we cache things in lit components. I would prefer if we eliminated this pattern so developers don't see it and copy it in more dangerous domains.

components/controllers/ConsistentEvaluationController.js
```diff
@@ -44,6 +45,50 @@ export class ConsistentEvaluationController {
...
+		if (!attachmentsEntity) {
+			throw new Error(ConsistentEvaluationControllerErrors.ERROR_FETCHING_ATTACHMENTS_ENTITY);
+		}
+
+		const canAddFeedbackFile = attachmentsEntity.hasActionByName('add-file');
+		const canRecordFeedbackVideo = attachmentsEntity.hasActionByName('add-video-note');
+		const canRecordFeedbackAudio = attachmentsEntity.hasActionByName('add-audio-note');
+		const attachments = [];
+		if (attachmentsEntity.entities) {
+			attachmentsEntity.entities.map(a => {
```
AlexBedley, line 68: nit: `map` returns an array which isn't used in this case so `forEach` would be more idiomatic

AlexBedley (changes requested):

AlexBedley (approved):

PR #188, opened 2020-12-01
Switch test & release to GitHub Actions

dlockhart:
As discussed in Slack, this converts the Test & Release workflows from Travis CI -> GitHub Actions. @svanherk is still working on a visual-diff action, but I this can run in hybrid mode (most things in GHAs, Visual Diff in Travis) temporarily.

Look over the instructions in the README to make sure everything is clear. I've switched the default version increment from patch -> minor so that if you need to do a hotfix there are patch versions available. If that's not what you want, I can put it back.

For those hotfix releases, how do they normally work branch-wise? I'd recommend naming the branches something like `release/20.21.1` according to the Brightspace release, or `1.15.x` according to the consistent-eval version. Let me know so I can add that branch name regex into the release workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: CI
+on: pull_request
+jobs:
+  test:
+    name: Test
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+      - uses: Brightspace/third-party-actions@actions/setup-node
+      - name: Install dependencies
+        run: npm ci
+      - name: Lint
+        run: npm run lint
+      - name: Unit Tests (SauceLabs)
+        run: npm run test:sauce
+        env:
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY_DESIRE2LEARN }}
+          SAUCE_USERNAME: Desire2Learn
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - master
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+.x'
 jobs:
   release:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
+jobs:
+  release:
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
...
+on: pull_request
+jobs:
+  test:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+      - uses: Brightspace/third-party-actions@actions/setup-node
+      - name: Install dependencies
+        run: npm ci
```
dlockhart, line 11: This runs `npm ci` instead of `npm install` since you're using a `package-lock.json` file. If you ever switch that this will need to change.

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
```
dlockhart, line 5: TODO: add in release/hotifx branch regex in here so that you can trigger releases from those branches

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+        with:
+          persist-credentials: false
+      - name: Setup Node
+        uses: Brightspace/third-party-actions@actions/setup-node
+      - name: Install dependencies
+        run: npm ci
+      - name: Unit Tests (Headless)
+        run: npm run test:headless
```
dlockhart, line 22: I kept this the same as before, where it'll run the headless tests in a non-pull request scenario. I think this is overkill since technically nothing should be merging to master without tests passing -- let me know if you want it removed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Incremental Release
+        uses: BrightspaceUI/actions/incremental-release@master
+        with:
+          DEFAULT_INCREMENT: minor
```
dlockhart, line 26: Switched this to minor as I mentioned in the PR description to give you wiggle room to do patch releases on top of a hotfix.

.github/workflows/release.yml
jeroach: Meh, it does seem overkill - perhaps we can lose it. Thoughts @jmurray-d2l

jeroach (approved):

.github/workflows/release.yml
dlockhart: I've removed it for now so you'll have super speedy releases. You can add it back later if it becomes problematic.

PR #189, opened 2020-12-02
fix demo page and add vdiff tests for submission list page

jeroach:
[US122584](https://rally1.rallydev.com/#/detail/userstory/458214617340?fdp=true): [Visual Diff Tests] Left Panel - Submission List

Skeleton:
image

File:
image

Text:
image


components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -22,7 +22,17 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 				type: String
 			},
 			token: {
-				type: Object
+				type: Object,
+				converter: {
```
jeroach, line 26: This converter is now required due to the token being an `Object` and us using a dummy `string` value in both the demo and the vdiff tests

demo/consistent-evaluation-submissions-page.html
```diff
@@ -32,15 +32,15 @@ <h2>File Submissions</h2>
 			<d2l-demo-snippet>
 				<d2l-consistent-evaluation-submissions-page id="fileSubmissions"
 				token="token"
-				submission-type="File submission"
+				submission-type="0"
```
jeroach, line 35: These need to be the proper values in order for the demo to render properly

demo/data/submission1.json
```diff
@@ -1,10 +1,11 @@
 {
-  "properties": {
-    "evaluationStatus": "Evaluated"
-  },
   "class": [
     "assignment-submission"
   ],
+  "properties": {
```
jeroach, line 5: So I updated all of the test data files such that:
- they have the minimum required info
- no references to sites that no longer exist

jeroach:
Any ideas how to regen goldens? That's the test failure.

I don't see the button on the failed build. Might be something do to with the recent switch to GitHub actions, but I thought the visual diff tests were still on Travis CI.

dbatiste:
> Any ideas how to regen goldens? That's the test failure.
> 
> I don't see the button on the failed build. Might be something do to with the recent switch to GitHub actions, but I thought the visual diff tests were still on Travis CI.

I think it might have to do with the fact that the bot looks for this message, and we recently removed the first stage from the travis config, making this stage the first stage. I _think_ this is why the bot is not engaging.

https://github.com/BrightspaceUI/visual-difference-bot/blob/master/index.js#L3

Once we get the VisualDiff GitHub action up and running, we won't need the bot any more, and this fragile logic won't be an issue.  In the interim, we can either put a no-op stage in here to fake it out, or make some changes to the bot to make it better.  @dlockhart thoughts?

dlockhart:
Bah, yep that's the issue. Sorry fighting another fire related to GHA with all the Serge translations failing -- can you just disable visual-diffs here temporarily and we can resolve this super soon with @svanherk converts all the visual-diff stuff to a GHA anyway?

jeroach:
> Bah, yep that's the issue. Sorry fighting another fire related to GHA with all the Serge translations failing -- can you just disable visual-diffs here temporarily and we can resolve this super soon when @svanherk converts all the visual-diff stuff to a GHA anyway?

Yeah that's fine - I'll let me team know as well, since several of us are working on vdiff items at the moment

dlockhart:
Ack sorry now I know what Dave B meant by a no-op. Like modify the `.travis.yml` file so that there's an empty stage 1, allowing vdiff to still be stage 2.

.travis.yml
```diff
@@ -1,23 +1,27 @@
...
+ - 14
+addons:
+  chrome: stable
+
+jobs:
+  include:
+  - stage: Empty First Stage
+    script:
+    - |
+      echo "Empty first stage so visual-diff is 2nd."
```
AlexBedley, line 12: looool

components/left-panel/consistent-evaluation-submissions-page.js
AlexBedley: I know @JamieTran had the exact same issue. It might be worth putting this somewhere more easily accessible so all `token`-using components can use it

AlexBedley (approved):
looks good! thanks for fixing the demo page

components/left-panel/consistent-evaluation-submissions-page.js
JamieTran: Yeah, we discussed this earlier in the sprint and decided to do it this way for now as to not block the 3 stories on the existence of the helper

JamieTran (approved):
Expanding the viewport for lcb vdiff tests also fixed things in my branch, thanks!

PR #191, opened 2020-12-03
US122972 - CE - evaluation entity - removing Attachments siren link -…

amauryperez:
… using new ID property on attachment sub-entity


components/controllers/ConsistentEvaluationController.js
```diff
@@ -66,13 +66,13 @@ export class ConsistentEvaluationController {
 		const attachments = [];
 		if (attachmentsEntity.entities) {
 			attachmentsEntity.entities.forEach(a => {
-				const self = a.hasLinkByRel('self') && a.getLinkByRel('self').href;
+				const id = a.properties && a.properties.id;
```
amauryperez, line 69: the API now provides an id, instead of the self link route

jmurray-d2l (approved):

PR #192, opened 2020-12-03
tidy up vdiff test setup and naming

jeroach:
As per thread in Slack:
https://d2l.slack.com/archives/CSL64JP9U/p1607006985030500

Also updated test names since I was editing each file anyway.


dbatiste (approved):
Thanks for cleaning these up Josh.  😃

PR #195, opened 2020-12-07
Scy/feedback demo

Scy-D2L:
Updated / fixed feedback demo page


demo/consistent-evaluation-feedback.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en-us">
```
Scy-D2L, line 2: Apparently the HTML editor in the feedback field checks for the specific language in use and doesn't like 'en' without a region. It threw an error unrelated to our stuff so I put this in.

jmurray-d2l (approved):

PR #196, opened 2020-12-08
Jmurray/active scoring rubric consistent eval

jmurray-d2l:
Story: [US115220](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F379942689136)

- Added a dropdown in rubrics so the evaluator is able to select which rubric is actually used for the overall assignment score. 
- Ensure rubric and grade sync works correctly with the selected scoring rubrics. 

grading-rubric


jeroach:
Hey guys, would it be possible to get some in context explanation of what's going on? 

It's a large PR, and there are a few areas I wasn't exactly clear on.

Thanks :)

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -118,18 +118,23 @@ export class ConsistentEvaluationRightPanel extends LocalizeConsistentEvaluation
...
 		this.canRecordFeedbackVideo = false;
 		this.canRecordFeedbackAudio = false;
-		this.rubricOpen = false;
+		this.rubricsOpen = 0;
 	}
 
 	_renderRubric() {
-		if (!this.hideRubric) {
+		if (this.rubricInfo && this.rubricInfo.length > 0) {
+			const hasOutOf = this.grade.getScoreOutOf();
```
jeroach, line 126: Do we need a null check on `this.grade`?

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -65,26 +64,12 @@ export class ConsistentEvaluationHrefController {
...
 			nextHref = this._getHref(root, nextRel);
 			previousHref = this._getHref(root, previousRel);
-			rubricHrefs = this._getHrefs(root, assessmentRel);
 			actorHref = this._getHref(root, actorRel);
 			userHref = this._getHref(root, userRel);
 			alignmentsHref = this._getHref(root, alignmentsRel);
 			groupHref = this._getHref(root, groupRel);
 			userProgressOutcomeHref = this._getHref(root, userProgressOutcomeRel);
 
-			if (rubricHrefs) {
```
JamieTran, line 75: We extracted this functionality into `getRubricInfo()` because we needed more information than just the `hrefs`.

components/consistent-evaluation-page.js
```diff
@@ -141,7 +141,11 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 			_unsavedAnnotationsDialogOpened: {
 				type: Boolean,
 				attribute: false
-			}
+			},
+			_activeScoringRubric: {
+				attribute: 'active-scoring-rubric',
+				type: Number
```
AlexBedley, line 147: nit: I do think that this should be `String` all the way down and we can avoid all the number parsing/worrying about types. Since this is an `id` we should only be checking for equality anyways (not trying like `_activeScoringRubric++`). Is there a place where we receive a `Number`?

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -54,10 +119,30 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
...
+	_getActiveScoringRubricSelectDropdown() {
+		if (!this.showActiveScoringRubricOptions) {
+			return html``;
+		}
+		const scoringRubrics = this.rubricInfo.filter(rubric => rubric.rubricScoringMethod !== 0);
+		if (scoringRubrics.length <= 0) {
+			return html``;
+		}
+		return html`
+			<h3 class="d2l-label-text">${this.localize('gradingRubric')}</h3>
```
AlexBedley, line 131: from an accessibility perspective, are we sure this should be an `h3` semantically and not just in style? (aka is this preceded by an `h1` and `h2`)

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -54,10 +119,30 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
 		return html`${rubrics}`;
 	}
 
+	_getActiveScoringRubricSelectDropdown() {
+		if (!this.showActiveScoringRubricOptions) {
+			return html``;
+		}
+		const scoringRubrics = this.rubricInfo.filter(rubric => rubric.rubricScoringMethod !== 0);
+		if (scoringRubrics.length <= 0) {
```
AlexBedley, line 127: I like that we're covering the negative length case ;)

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -24,28 +34,83 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
...
-		const rubrics =	this.rubricHrefs.map(rubric => {
+		const rubrics =	this.rubricInfo.map(rubric => {
 			if (!rubric) {
 				return html``;
 			}
+
 			return html`
 				<div class="d2l-consistent-evaluation-rubric">
 					<d2l-rubric
+						rubric-id=${rubric.rubricId}
```
AlexBedley, line 105: nit: when attaching attributes like this, it's customary to prefix it with `data-`

https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -24,28 +34,83 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
...
+			}
+		`];
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('activeScoringRubric')) {
+			const activeRubricDropdown = this.shadowRoot.querySelector('.d2l-consistent-evaluation-active-scoring-rubric');
+			if (activeRubricDropdown) {
+				activeRubricDropdown.value = this.activeScoringRubric ?
```
AlexBedley, line 53: I checked and I solved this the same way you did

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/f6f9d152b8b722056e2b68cd8389f6108a75427e/components/header/d2l-consistent-evaluation-lcb-file-context.js#L167

components/consistent-evaluation-page.js
JamieTran: We compare the `_activeScoringRubric` to the `rubricId` we get from the API (which is a `Number`)

components/right-panel/consistent-evaluation-rubric.js
JamieTran: I talked to Dan, this should be h2 like our other headers in the `right-panel`

components/consistent-evaluation-page.js
AlexBedley: maybe I missed it in the PR, but I only found 2 places we compare `activeScoringRubric` and in all cases we have called `parseInt` on both sides of the equation (at some point)

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/2768b32dcfab53d13a092b669c9e1245f2568a98/components/right-panel/consistent-evaluation-rubric.js#L68

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/2768b32dcfab53d13a092b669c9e1245f2568a98/components/right-panel/consistent-evaluation-rubric.js#L135

components/right-panel/consistent-evaluation-rubric.js
AlexBedley: ok and just to confirm, Dan meant semantically an `h2`? Not just text formatted to look like an `h2`?

components/right-panel/consistent-evaluation-right-panel.js
JamieTran: No, `this.grade` will always be [set ](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/a99148b8560f5ad74303a077854022cb2f6374d0/components/consistent-evaluation-page.js#L236) one level up in `consistent-evaluation-page`

components/right-panel/consistent-evaluation-rubric.js
JamieTran: yeah, exactly from the accessibility perspective.

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -54,10 +119,30 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
...
+		const scoringRubrics = this.rubricInfos.filter(rubric => rubric.rubricScoringMethod !== 0);
+		if (scoringRubrics.length <= 0) {
+			return html``;
+		}
+		return html`
+			<h2 class="d2l-label-text">${this.localize('gradingRubric')}</h2>
+			<select class="d2l-input-select d2l-truncate d2l-consistent-evaluation-active-scoring-rubric" aria-label=${this.localize('activeGradingRubric')} @change=${this._onActiveScoringRubricChange}>
+				<option label=${this.localize('noActiveGradingRubric')} ?selected=${!this.activeScoringRubric} value=null></option>
+				${scoringRubrics.map(rubric => html`
+						<option value="${rubric.rubricId}" label=${rubric.rubricTitle} class="select-option"></option>
```
JamieTran, line 135: I removed `?selected`, it seems that the updated() method is handling this for us

AlexBedley (approved):

PR #200, opened 2020-12-10
Jroach/us122195

jeroach:
[US122195](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fuserstory%2F452637966616&fdp=true?fdp=true): Loading skeleton is not required for Observational assignments

File Assignment with skeleton:
file_has_skeleton

Observational Assignment (skeleton only for user name and profile image):
observational_no_skeleton

### DIFF
--- a/components/header/d2l-consistent-evaluation-learner-context-bar.js
+++ b/components/header/d2l-consistent-evaluation-learner-context-bar.js
@@ -73,9 +73,6 @@ export class ConsistentEvaluationLearnerContextBar extends SkeletonMixin(RtlMixi
 				align-items: center;
 				display: flex;
 			}
-			.d2l-skeleton-learner-context-bar-hidden {
-				display: none;
-			}
 			:host([skeleton]) .d2l-consistent-evaluation-learner-context-bar {
 				display: none;
 			}
@@ -129,23 +126,22 @@ export class ConsistentEvaluationLearnerContextBar extends SkeletonMixin(RtlMixi
 		return this.userHref ? this.userHref : this.groupHref;
 	}
 
-	_getSkeletonClass() {
-		let className = 'd2l-skeleton-learner-context-bar';
-
+	_renderSkeleton() {
 		if (submissionTypesWithNoEvidence.includes(this.submissionInfo.submissionType)) {
-			className = 'd2l-skeleton-learner-context-bar-hidden';
+			return html ``;
 		}
 
-		return className;
+		return html`
+			<div class="d2l-skeleton-learner-context-bar" aria-hidden="${!this.skeleton}" aria-busy="${this.skeleton}">
+				<div class="d2l-skeleton-user-profile-image d2l-skeletize"></div>
+				<div class="d2l-skeleton-user-display-name d2l-skeletize"></div>
+				<div class="d2l-skeleton-submission-select d2l-skeletize"></div>
+			</div>`;
 	}
 
 	render() {
 		return html`
-			<div class="${this._getSkeletonClass()}" aria-hidden="${!this.skeleton}" aria-busy="${this.skeleton}">
-				<div class="d2l-skeleton-user-profile-image d2l-skeletize"></div>
-				<div class="d2l-skeleton-user-display-name d2l-skeletize"></div>
-				<div class="d2l-skeleton-submission-select d2l-skeletize"></div>
-			</div>
+			${this._renderSkeleton()}
 			<div class="d2l-consistent-evaluation-learner-context-bar" aria-hidden="${this.skeleton}">
 				<d2l-consistent-evaluation-lcb-user-context
 					.href=${this._getActorHref()}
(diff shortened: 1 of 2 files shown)


components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -125,13 +126,22 @@ export class ConsistentEvaluationLearnerContextBar extends SkeletonMixin(RtlMixi
 		return this.userHref ? this.userHref : this.groupHref;
 	}
 
-	render() {
+	_renderSkeleton() {
+		if (submissionTypesWithNoEvidence.includes(this.submissionInfo.submissionType)) {
```
jeroach, line 130: This also covers `On Paper` assignments

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -234,6 +234,8 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 			if (this.submissionInfo.submissionList === undefined) {
 				return this._renderNoSubmissions();
 			}
+
+			return this._renderSubmissionList();
```
jeroach, line 238: Moving this render call inside this `if` block ensures that the file list skeleton isn't loaded in the default case.

This was showing for all types since at first `this.submissionInfo` was `undefined`

components/header/d2l-consistent-evaluation-learner-context-bar.js
```diff
@@ -125,13 +126,22 @@ export class ConsistentEvaluationLearnerContextBar extends SkeletonMixin(RtlMixi
...
 				<div class="d2l-skeleton-user-profile-image d2l-skeletize"></div>
 				<div class="d2l-skeleton-user-display-name d2l-skeletize"></div>
 				<div class="d2l-skeleton-submission-select d2l-skeletize"></div>
-			</div>
+			</div>`;
+	}
+
+	render() {
+		return html`
+			${this._renderSkeleton()}
```
jmurray-d2l, line 144: Should the learner context bar skeleton be affected by the submission type?

components/left-panel/consistent-evaluation-left-panel.js
jmurray-d2l: Does this cause the skeleton to take a second to show?  I originally put it there so the skeleton would show even before we knew what was being loaded

components/header/d2l-consistent-evaluation-learner-context-bar.js
jeroach: That was the ask based on that it contains the drop down list of file/text submissions.

Though you make a good point in that the profile image and user name skeleton likely should be there regardless.

I'll look into breaking out the skeleton rendering logic. Thanks :)

jmurray-d2l: Okay cool! Dan might have an opinion on this, he could probably let us know

jmurray-d2l:
The gif looks good to me, I would suggest trying it with a student with a large number of submissions, maybe 10 or so? It flickers so fast with only 1 to load

components/left-panel/consistent-evaluation-left-panel.js
jeroach: Yes it does delay the loading ever so slightly. But I don't think we can get around that since we don't want the skeleton to show for Observational and In Person.

In the case where `this.submissionInfo` is `undefined` we currently don't know what type of submission it is.

Given the delay is really short, I think this approach is fine. But open to other ideas if you think otherwise.

jeroach:
> The gif looks good to me, I would suggest trying it with a student with a large number of submissions, maybe 10 or so? It flickers so fast with only 1 to load

Replaced the file gif with one with more submissions - hopefully this helps

components/header/d2l-consistent-evaluation-learner-context-bar.js
jeroach: Dan confirmed that we want to have a skeleton for the user aspects. I have updated the PR accordingly.

Good catch!

components/left-panel/consistent-evaluation-left-panel.js
jmurray-d2l: Yeah I'm okay with it as well, I was just curious as to what you were seeing

jmurray-d2l (approved):

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -248,7 +248,9 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 			return this._renderNonViewableFile();
 		}
 
-		return this._renderSubmissionList();
+		if (this.submissionInfo) {
```
AlexBedley, line 251: I've never considered what would happen if `render` returns `undefined`. Does it just not render anything?

dbatiste: For the most part yes, although it's being rendered into a slot that has default fallback content, the `nothing` should be used... since it really does result in nothing being rendered.

https://lit-html.polymer-project.org/guide/writing-templates#rendering-nothing

AlexBedley (approved):

PR #201, opened 2020-12-11
reduce padding to ensure one line for all languages

jeroach:
[DE41679](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fdefect%2F463461237476&fdp=true?fdp=true): Translations - The text for User x of y in the navbar is wrapping with Japanese and Arabic languages

`2.5rem` fixes this for sufficiently large class sizes (665 of 665):
japanese


Scy-D2L:
It seems a bit odd to be assigning specific padding to the text instead of using a div width with centered text and/or margins on the `d2l-iterator-space` - the functional difference is minor and I'm guessing we don't care much, but doesn't this change the position of the next/previous buttons based on language?

Scy-D2L (approved):

JamieTran (approved):

PR #216, opened 2020-12-17
- Setting no scoring rubric to the default value of -1

jmurray-d2l:


JamieTran (approved):
It was inevitable - Thanos

PR #219, opened 2020-12-17
Hit cache everytime for submissions-page. Assume read/flag worked

AlexBedley:
This PR contains 2 changes:

1. Use cache for submission entities
    * This consists of the single line boolean change on line 181
2. Assume flag/read worked so we can update the UI instantly
    * The rest of the diff is this change

The code change is small but the ramifications are complex. First let's talk about using the cache for submission entities. At first glance it looks wrong, since we're always fetching from cache but never invalidating the cache. It turns out we are implicitly updating the cache here:

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/left-panel/consistent-evaluation-submissions-page.js#L226

What's happening there is `this._submissionEntities[i].entity` is actually the exact same reference that is being used by `EntityStore`. Thus that line is actually updating the cache with our new values (which is pretty convenient). The downside here is it's very brittle, we MUST update the cache every time the submission entity changes, otherwise when we fetch from cache it will be stale. Right now there are only 2 ways to do that, which is toggleRead and toggleFlagged.

So how can we make that better? Well we could be more explicit about evicting the cache (via `EntityStore.update`) to give hints to maintainers. But that still sticks us with the problem of manually updating the cache if the entity changes. I think the real fix is to use `Link` headers on our `toggleRead` and `toggleFlagged` option. Then when we get the action response, it automatically updates the cache for us, and we can check that versus our internal `this._submissionEntities[i].entity` to make sure our prediction was correct.

The 2nd feature is to make toggling read and flagged appear instantaneous, by assuming it worked and then comparing it to the actual response when we get it (very similar to my suggestion above, except much simpler since we don't need `Link` headers here.) The feature includes the ability to detect when our assumption was incorrect, but it does not include handling that in a user-friendly manner. Originally I had added in metrics gathering for that, before that PR was reverted.

### DIFF
--- a/components/left-panel/consistent-evaluation-submissions-page.js
+++ b/components/left-panel/consistent-evaluation-submissions-page.js
@@ -4,11 +4,12 @@ import '@brightspace-ui/core/components/list/list.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import './consistent-evaluation-submission-item.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { toggleFlagActionName, toggleIsReadActionName } from '../controllers/constants.js';
 import { Classes } from 'd2l-hypermedia-constants';
+import { findFile } from '../helpers/submissionsAndFilesHelpers.js';
 import { performSirenAction } from 'siren-sdk/src/es6/SirenAction.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
-import { toggleIsReadActionName } from '../controllers/constants.js';
 
 export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(LitElement)) {
 	static get properties() {
@@ -177,7 +178,7 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 	}
 
 	async _getSubmissionEntity(submissionHref) {
-		const byPassCache = true;
+		const byPassCache = false;
 		return await window.D2L.Siren.EntityStore.fetch(submissionHref, this._token, byPassCache);
 	}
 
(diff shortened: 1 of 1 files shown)


components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -177,7 +178,7 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 	}
 
 	async _getSubmissionEntity(submissionHref) {
-		const byPassCache = true;
+		const byPassCache = false;
```
AlexBedley, line 181: The change that launched a thousand test cases

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -238,8 +239,31 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 			throw new Error('Invalid entity provided for attachment');
 		}
 
+		// Assume toggle worked so we can update UI instantly
+		if (actionName === toggleIsReadActionName) {
```
AlexBedley, line 243: I'm really not a fan of duplicating server logic on the front-end, as that was our main design goal (avoiding this). However, I can't think of another way to pretend it was instantaneous.

As I called out when we started this, the cost of these load time improvements will be violations of our design.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -238,8 +239,31 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
...
+			await this.requestUpdate();
+		} else if (actionName === toggleFlagActionName) {
+			attachmentEntity.properties.flagged = !attachmentEntity.properties.flagged;
+			await this.requestUpdate();
+		}
+
 		const action = attachmentEntity.getActionByName(actionName);
-		await this._doSirenActionAndRefreshFileStatus(action);
+		const newSubmissionEntity = await this._doSirenActionAndRefreshFileStatus(action);
+		const newAttachmentEntity = findFile(fileId, [{ entity: newSubmissionEntity }]);
```
AlexBedley, line 253: The second parameter is weird.

I did it this way because I needed to coerce the `submissionEntity` to the format that `findFile` expects. I decided not to make another function because I figured that would be more confusing (the functions would essentially be identical). I tried to modify `findFile` to automatically wrap the parameter in an array if it wasn't an array already, but then you still need to add the `entity` envelope and it felt wrong. 🤷

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -238,8 +239,31 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
...
+
+		// Detect if our assumption was correct
+		if (
+			actionName === toggleIsReadActionName
+			&& attachmentEntity.properties.read !== newAttachmentEntity.properties.read
+		) {
+			console.warn('Toggle read failed');
```
AlexBedley, line 260: I originally had us send metrics here, since I would consider this case to be rarely hit. It might make sense to toast the user (something like "flagging failed"). Ill check with Dan.

components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -262,6 +286,7 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 		const submissionSelfLink = newSubmissionEntity.getLinkByRel('self');
 		await this._updateSubmissionEntity(newSubmissionEntity, submissionSelfLink.href) ;
 		await this.requestUpdate();
+		return newSubmissionEntity;
```
AlexBedley, line 289: Returning the new entity so I can compare it to my guess

components/left-panel/consistent-evaluation-submissions-page.js
AlexBedley: I pulled up a blank here for a unit test that makes sense. If I stub out EntityStore, that might give us a false sense of security (what if EntityStore started returning object clones for example). And we don't have an integration environment set up to run a full test.

jmurray-d2l (approved):
Looks good to me!

components/left-panel/consistent-evaluation-submissions-page.js
jmurray-d2l: I think the cost of this going wrong is very small compared to the value we get out of it going right

jeroach (approved):

PR #224, opened 2020-12-23
Add vdiff tests for rubric popout

JamieTran:
Added vdiff tests and demo page for `popup-rubric` component


components/right-panel/consistent-evaluation-popup-rubric.js
```diff
@@ -32,6 +44,7 @@ export class ConsistentEvaluationPopupRubric extends LocalizeConsistentEvaluatio
 		if (!this._rubricInfos) {
 			return html ``;
 		}
+		console.log(this._rubricInfos);
```
jmurray-d2l, line 47: debug left over?

jmurray-d2l (approved):

demo/consistent-evaluation-popup-rubric.html
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
```
dbatiste, line 1: FYI, it's not necessary to explicitly include `typography.js`, set the `d2l-typography` class on the body, or specify a `title` element in the head.  The `d2l-demo-page` component will do all these things automatically.  We've seen this come up a couple times recently - I am guessing it's being copied from somewhere?

This is basically what it should look like:
```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
		<script type="module">
			import '@brightspace-ui/core/components/demo/demo-page.js';
			import '../some-component.js';
		</script>
		<meta name="viewport" content="width=device-width, initial-scale=1.0">
		<meta http-equiv="X-UA-Compatible" content="ie=edge">
		<meta charset="UTF-8">
	</head>
	<body unresolved>
		<d2l-demo-page page-title="d2l-some-component demo">

			<h2>Some Heading</h2>

			<d2l-demo-snippet>
				<template>
					<d2l-some-component></d2l-some-component>
					<script>
						// any script that should be run when template is stamped
					</script>
				</template>
			</d2l-demo-snippet>

		</d2l-demo-page>

		<script>
			// any script that is safe to run before demo templates stamped
		</script>
	</body>
</html>
```
The demo-page component will remove the `unresolved` attribute when everything is ready.

test/perceptual/consistent-evaluation-popup-rubric.visual-diff.js
```diff
@@ -0,0 +1,31 @@
...
+describe('d2l-consistent-evaluation-popup-rubric', () => {
+
+	const visualDiff = new VisualDiff('consistent-evaluation-popup-rubric', __dirname);
+
+	let browser, page;
+
+	before(async() => {
+		browser = await puppeteer.launch({
+			headless: true,
+			args: ['--no-sandbox', '--disable-setuid-sandbox', '--lang=en-GB']
```
dbatiste, line 13: Are these needed?  The reason I ask is that we're looking making some changes to visual-diff to eliminate much of the boiler plate code that's currently necessary, and if necessary, we'll need to look at how to provide this configuration in those changes.

Note: `headless` is `true` by default.  Also for our lang stuff to work I think we'd expect the `lang` attribute to be specified on the `html` element (which is currently is).

PR #231, opened 2021-01-05
Vy/add telemetry client

vix633:
Let's give this another whirl. Telemetry client could get called before it was initted :(


components/consistent-evaluation.js
```diff
@@ -139,6 +146,9 @@ export class ConsistentEvaluation extends LitElement {
 			}
 		}
 		this._loading = false;
+		if (this._telemetry) {
+			this._telemetry.logLoadEvent('consistentEvalMain', this._submissionInfo.submissionList.length);
```
jmurray-d2l, line 150: What happens if there are no submission?

vix633: Can we even load this page without submissions?

jmurray-d2l: Yep, we support no submissions - observational assignments are a thing we support

vix633: Ah yeah we can, i'll add a null check!

jeroach (approved):
LGTM

jmurray-d2l (approved):

PR #233, opened 2021-01-06
Jmurray/rubric popout styling

jmurray-d2l:


components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -245,10 +249,17 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(LitElement
 			html`<h2 aria-label=${this.localize('rubricTitle')} class='d2l-consistent-evaluation-rubric-title'>${rubricTitle}</h2>`;
 	}
 
+	_getHeader() {
```
JamieTran, line 252: This can be simplified to `return this.isPopout ? '': this.header` or remove this function entirely and put it in the render below.

JamieTran (approved):

PR #235, opened 2021-01-07
add users name to rubrics

vix633:
[US123683](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F478450711396&fdp=true?fdp=true): [Right Panel] Rubric pop out - title should be Assess Rubric - [student name]

### DIFF
--- a/components/right-panel/consistent-evaluation-popup-rubric.js
+++ b/components/right-panel/consistent-evaluation-popup-rubric.js
@@ -21,6 +21,7 @@ export class ConsistentEvaluationPopupRubric extends LocalizeConsistentEvaluatio
 					}
 				}
 			},
+			_userName: {type: String},
 			_rubricInfos: { type: Array },
 		};
 	}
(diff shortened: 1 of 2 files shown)


lang/en.js
```diff
@@ -54,6 +54,7 @@ export default {
 	"rubricSummary": "1 rubric attached",
 	"rubricsSummary": "{num} rubrics attached",
 	"rubricTitle": "Rubric Title",
+	"rubricsAssess": "Assess Rubrics",
```
jeroach, line 57: Should this lang term incorporate the username? e.g. `Assess Rubrics for {userName}`

Concern being that it might not read correctly as is for other languages

components/right-panel/consistent-evaluation-popup-rubric.js
```diff
@@ -37,6 +38,17 @@ export class ConsistentEvaluationPopupRubric extends LocalizeConsistentEvaluatio
...
+			this.setTitle();
+		}
+
+	}
+
+	setTitle() {
+		if (this._userName) {
+			const title = document.createElement('title');
+			title.textContent = `${this.localize('rubricsAssess')} - ${this._userName}`;
+			document.head.insertBefore(title, document.head.firstChild);
```
jeroach, line 51: This way of setting the title seems fishy. Can we not build out the logic for the title in a web component, and then just set an attribute here?

vix633: This is the example d2l uses in the [lit demo pages](https://github.com/BrightspaceUI/core/blob/master/components/demo/demo-page.js#L51). 

Could use the query selector as well, would you prefer to see that?

lang/en.js
vix633: I think I'll do that! Just let me check with Dan.

components/right-panel/consistent-evaluation-popup-rubric.js
jeroach: Fair enough, if there is a precedent for it then who am I to object :)

jeroach (approved):

lang/en.js
vix633: Good with Dan!

PR #243, opened 2021-01-14
add getGroupName; use in rubric popup

jeroach:


components/right-panel/consistent-evaluation-popup-rubric.js
```diff
@@ -38,7 +38,11 @@ export class ConsistentEvaluationPopupRubric extends LocalizeConsistentEvaluatio
 		if (changedProperties.has('href')) {
 			const controller = new ConsistentEvaluationHrefController(this.href, this.token);
 			this._rubricInfos = await controller.getRubricInfos();
+
 			this._userName = await controller.getUserName();
+			if(!this._userName) {
```
jeroach, line 43: Thoughts on this logic?

- If it's a user assignment, then there will be a user rel (and no group rel), and this call will return a valid name
- If it's a group assignment, then there will be a group rel (and no user rel), and this call will return a valid group name

vix633: Yeah, just rename the variable since it doesn't only store the username now :)

vix633 (approved):

PR #246, opened 2021-01-14
Add Turnitin grade-mark component to CE

JamieTran:
Add TII grade mark
image

- Edit button opens up TII in a new window
- Refresh button pulls new data from the API (displays `Refreshing...`)
- Use this Grade button syncs the TII grade to the right-panel-grade component


components/left-panel/consistent-evaluation-submissions-page.js
```diff
@@ -297,6 +301,34 @@ export class ConsistentEvaluationSubmissionsPage extends SkeletonMixin(RtlMixin(
 		}
 	}
 
+	async _refreshGradeMarkTiiAction(e) {
```
JamieTran, line 304: This function is almost identical to `_submitFileTiiAction()` maybe there's a better way to reuse code?

components/left-panel/consistent-evaluation-tii-grade-mark.js
```diff
@@ -0,0 +1,149 @@
...
+
+	_onEditButtonClick() {
+		window.open(this.gradeMarkHref);
+	}
+
+	_dispatchRefreshButtonClick() {
+		this._overallScore = this.localize('turnitinRefreshing');
+
+		// This timeout is here for when the user refreshes and there is no change in the API(thus no re-render)
+		setTimeout(() => { this._setOverallScore(); }, 3500);
```
JamieTran, line 72: Kinda sus but this component will not rerender if the API returns the same values so it would keep displaying `Refreshing...` indefinetly

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -135,6 +141,9 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 			margin-right: 0.5rem;
 			text-transform: none;
 		}
+		d2l-consistent-evaluation-tii-similarity {
+			margin-right: 2.5rem;
```
jeroach, line 145: We'll need `rtl` support for this - similar to how it's done for `d2l-status-indicator`

components/left-panel/consistent-evaluation-submissions-page.js
jeroach: Could extract the shared parts to a common helper - I can go either way

components/left-panel/consistent-evaluation-tii-grade-mark.js
```diff
@@ -0,0 +1,149 @@
...
+					icon="tier1:grade"
+					text="${this.localize('turnitinUseGrade')}"
+					@click=${this._dispatchUseGradeEvent}
+				></d2l-button-subtle>
+			` :
+			html ``;
+	}
+
+	_renderEditButton() {
+		// @click should make another api call and re-render to update the grade value
```
jeroach, line 108: Is this a todo?

JamieTran: Oh no I missed removing this comment 🙈

components/left-panel/consistent-evaluation-submissions-page.js
JamieTran: I think I'll leave it as is, we can't really tell which action is being called unless we specify it in the event - which I don't think is a great idea.

components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -135,6 +141,12 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
...
+		}
+		:host([dir="rtl"]) d2l-consistent-evaluation-tii-similarity {
+			margin-left: 2.5rem;
```
jeroach, line 148: You also need to zero out the right side:  `margin-right: 0;`

JamieTran: Would I also have to do `margin-left: 0rem;` on the regular style? Or does it only apply to the `rtl` view

jeroach (approved):

components/left-panel/consistent-evaluation-submission-item.js
jeroach: Yeah you're good.

PR #252, opened 2021-01-19
US123981/mbosilj/ display assessor name on rubric.

mbosiljevac:
Displays either Graded By {assessor} or (not graded yet)
image
image


lang/en.js
```diff
@@ -37,6 +38,7 @@ export default {
 	"noFeedbackSummary": "No feedback entered",
 	"noGradeSummary" : "No grade entered",
 	"noSubmissions": "No submissions made for this assignment",
+	"notGradedSummary" : "(not graded yet)",
```
jmurray-d2l, line 41: Not sure if we should include the parentheses in the lang term

mbosiljevac: mm that's fair, I'll move them out!

jmurray-d2l (approved):
Well done!

PR #258, opened 2021-01-21
Extract href converter logic to helper

JamieTran:
Reduce repetitive converter code

### DIFF
--- a/components/helpers/converterHelpers.js
+++ b/components/helpers/converterHelpers.js
@@ -1,0 +1,4 @@
+export function convertHref(value) {
+	const retVal = String(value);
+	return retVal;
+}
--- a/components/helpers/converterHelpers.js
+++ b/components/helpers/converterHelpers.js
@@ -1,4 +1,4 @@
-export function convertHref(value) {
+export function convertToken(value) {
 	const retVal = String(value);
 	return retVal;
 }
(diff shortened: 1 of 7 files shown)


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -42,16 +43,7 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeConsistentEvalu
...
-					fromAttribute(value) {
-						const retVal = String(value);
-						return retVal;
-					},
-					toAttribute(value) {
-						const retVal = Object(value);
-						return retVal;
-					}
-				}
+				converter: (value, type) => convertHref(value, type),
```
jeroach, line 46: Why is `type` being passed in here?

components/helpers/converterHelpers.js
```diff
@@ -0,0 +1,4 @@
+export function convertHref(value) {
+	const retVal = String(value);
```
jeroach, line 2: Does this implicitly handle the conversion `toAttribute` ?

I ask because the code looks to cover only the `fromAttribute` case

jeroach (commented):
A couple of questions.

Thanks for cleaning this up ;)

components/helpers/converterHelpers.js
JamieTran: You got it, although I do have to change them to reflected attributes.
This is from the docs: _If no `toAttribute` function is supplied for a reflected attribute, the attribute is set to the property value without conversion._

jeroach (approved):

PR #261, opened 2021-01-25
US122595/mbosilj show dialog if unscored criteria.

mbosiljevac:
Show a dialog if the user tries to publish/update a rubric with unscored criteria.
Notes:
* must be merged after https://github.com/Brightspace/lms/pull/5802
* used same copy as before except "Publish Anyways" I changed to "Publish" because "Publish Anyways" forced the buttons to be on different lines.
* The only problem with putting the dialog in a separate file is there is ~1second delay after the user hits the publish button and the dialog shows up. This is probably because the consistent-evaluation-page has to rerender but I'm not 100% sure on the cause
image
image


jmurray-d2l (approved):
Very well done! Nice work

components/consistent-evaluation-dialogs.js
```diff
@@ -0,0 +1,150 @@
...
+	async _fireDialogClosedEvent() {
+		this.dispatchEvent(new CustomEvent('d2l-unscored-criteria-dialog-closed', {
+			composed: true,
+			bubbles: true
+		}));
+	}
+
+	render() {
+		return html`
+			<d2l-dialog-confirm
```
dbatiste, line 130: Tip: the `d2l-dialog-confirm` does have some base cost that would be good to avoid if possible.  I think it would be better usage ton conditionally render whichever dialog is needed, but in this case, they look almost identical and the on or two properties within can be conditionally rendered.

mbosiljevac: Thanks for the tip, I didn't know about the base cost when working on this! I'm planning to add some more dialogs to this class so I'll be sure to fix this up to conditionally render properties/dialogs while I'm still working in this file!

PR #262, opened 2021-01-25
Jmurray/new html editor

jmurray-d2l:


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -135,6 +149,32 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeConsistentEvalu
 		this._feedbackSummaryInfo = summary;
 	}
 
+	_getHtmlEditor() {
+		if (this.useNewHtmlEditor) {
+			import('@brightspace-ui/htmleditor/htmleditor.js');
```
jmurray-d2l, line 154: @dbatiste We ran into an issue with the old editor an importing the new one, any concerns about conditionally importing?

dbatiste: Nope - I think FACE will be doing a dynamic import as well for the same reason.

amauryperez (approved):

test/perceptual/consistent-evaluation-right-panel-new-editor.visual-diff.html
```diff
@@ -0,0 +1,170 @@
+<!doctype html>
+<html lang="en-us">
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<title>d2l-consistent-evaluation-right-panel visual diff tests</title>
+		<script type="module">
+			import '@brightspace-ui/core/components/typography/styles.js';
```
dbatiste: Not sure why `@brightspace-ui/core/components/typography/styles.js` is included here.  It just exports a bunch of styles that can be included in the context of a web component, not really usable in a web page.

amauryperez: yeah, I just copied whatever was in `test/perceptual/consistent-evaluation-right-panel.visual-diff.html` over to this new test that will cover the new html editor

amauryperez: But couldn't figure out why the visual tests are failing so removed them from this PR to handle visual tests in a separate one

PR #277, opened 2021-02-01
US122594/mbosilj Update dialog copy.

mbosiljevac:
Talked to Dan, ok with using default d2l-dialog ('x' button has same behaviour as "Continue Grading"). Confirmation dialog was too narrow for the "Continue Grading" and "Publish Anyway" buttons, so using d2l-dialog insteacd. Dialog looks like: 
image


jmurray-d2l (approved):

JamieTran (approved):
👍

PR #278, opened 2021-02-02
Add user-profile-card to CE

JamieTran:
This is the initial implementation of the `user-profile-card`'s UI/responsiveness as we are not getting any data from the backend yet

When hovering over/clicking the `user-context` component, the `user-profile-card` is displayed
The `user-profile-card` disappears when the user navigates their cursor off the `user-profile-card`

Notes: 
There isn't an option to hide the online/offline status so for the current implementation it'll be forever offline

image

### DIFF
--- a/components/header/consistent-evaluation-user-profile-card.js
+++ b/components/header/consistent-evaluation-user-profile-card.js
@@ -1,0 +1,41 @@
+import '@brightspace-ui-labs/user-profile-card/user-profile-card.js';
+import '@brightspace-ui/core/components/icons/icon.js';
+
+import { html, LitElement } from 'lit-element';
+import { LocalizeConsistentEvaluation } from '../../lang/localize-consistent-evaluation.js';
+
+export class ConsistentEvaluationUserProfileCard extends LocalizeConsistentEvaluation(LitElement) {
+
+	static get properties() {
+		return {
+			displayName: {
+				attribute: 'display-name',
+				type: String
+			},
+			tagline: {
+				attribute: 'tag-line',
+				type: String
+			}
+		};
+	}
+
+	dispatchMouseLeaveEvent() {
+		this.dispatchEvent(new CustomEvent('d2l-consistent-eval-profile-card-mouse-leave', {
+			composed: true,
+			bubbles: true,
+		}));
+	}
+
+	render() {
+		return html`
+		<d2l-labs-user-profile-card
+			@mouseleave=${this.dispatchMouseLeaveEvent}
+			tagline=${this.tagline}>
+			<img slot="illustration" src="" width="116px" height="116px">
+			${this.displayName}
+		</d2l-labs-user-profile-card>
+		`;
+	}
+}
+
+customElements.define('d2l-consistent-evaluation-user-profile-card', ConsistentEvaluationUserProfileCard);
--- a/components/header/consistent-evaluation-user-profile-card.js
+++ b/components/header/consistent-evaluation-user-profile-card.js
@@ -13,7 +13,6 @@ export class ConsistentEvaluationUserProfileCard extends LocalizeConsistentEvalu
 				type: String
 			},
 			tagline: {
-				attribute: 'tag-line',
 				type: String
 			}
 		};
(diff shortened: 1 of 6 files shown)


components/header/consistent-evaluation-user-profile-card.js
```diff
@@ -0,0 +1,41 @@
...
+			bubbles: true,
+		}));
+	}
+
+	render() {
+		return html`
+		<d2l-labs-user-profile-card
+			@mouseleave=${this.dispatchMouseLeaveEvent}
+			tagline=${this.tagline}>
+			<img slot="illustration" src="" width="116px" height="116px">
```
JamieTran, line 34: profile pic src kept empty for when we do get it from the API

components/header/consistent-evaluation-user-profile-card.js
```diff
@@ -0,0 +1,41 @@
...
+export class ConsistentEvaluationUserProfileCard extends LocalizeConsistentEvaluation(LitElement) {
+
+	static get properties() {
+		return {
+			displayName: {
+				attribute: 'display-name',
+				type: String
+			},
+			tagline: {
+				attribute: 'tag-line',
```
jeroach, line 16: Should this just be `tagline`? Usually we only hyphen if there is camel case in the var name

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -100,11 +115,40 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
...
+	}
+
+	_toggleOffProfileCard() {
+		this._showProfileCard = false;
+	}
+
 	render() {
 		return html`
+		<div class="d2l-user-context-container"
+			@click=${this._toggleOnProfileCard}
```
jeroach, line 141: Doesn't `mouseover` supersede `click`? One would assume that you'd need to have the mouse over in order for a click to register.

Unless this covers a case I'm not thinking of.

On that topic, what happens if the user navigates to the element via keyboard? I wonder if we need a `focus` check

components/header/consistent-evaluation-user-profile-card.js
JamieTran: hmm you're right, the `slot="illustration"` in the `user-profile-card` already is already set up with a grid that sets `width` and `height` to `116px`. Putting it here is redundant.

components/header/d2l-consistent-evaluation-lcb-user-context.js
JamieTran: > Doesn't mouseover supersede click? One would assume that you'd need to have the mouse over in order for a click to register.

Yeah... now that you put it that way, this is kinda redundant.

> On that topic, what happens if the user navigates to the element via keyboard? I wonder if we need a focus check

This wouldn't support keyboard navigation... I'll investigate how we can implement a `focus` check to when the `user-context` is selected

jeroach (approved):

components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -100,11 +125,41 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
...
+				display-name=${ifDefined(this._displayName)}
+				tagline="This is a tag-line that will come from the API?"
+				@d2l-consistent-eval-profile-card-mouse-leave=${this._toggleOffProfileCard}>
+			</d2l-consistent-evaluation-user-profile-card>
+			` :
+			html``;
+	}
+
+	_toggleOnProfileCard() {
+		this._showProfileCard = true;
```
Scy-D2L, line 141: Jamie and I had a slack conversation about this, but while this setup will show the profile card it doesn't make it accessible to keyboard navigation (visible, not focusable) so that will need to be addressed with design for accessibility.

vix633:
👀  Is this blocked at the moment, or waiting vdiff?

Scy-D2L (approved):

PR #279, opened 2021-02-03
US124041/mbosilj display tii feedback icon.

mbosiljevac:
Note: text displays on hover and wording was taken from old experience
feedback_tii


components/left-panel/consistent-evaluation-tii-grade-mark.js
```diff
@@ -135,12 +158,13 @@ export class ConsistentEvaluationTiiGradeMark extends LocalizeConsistentEvaluati
 	render() {
 		return html`
 			<div class="d2l-label-text">${this.localize('turnitinGradeMark')}</div>
-			<div class="d2l-body-compact">
+			${this._renderFeedbackIcon()}
+			<div class="d2l-body-compact" id=overall-score-container>
```
jeroach, line 162: Classes are preferred to using ids. For this case, the div can simply have two classes:

`<div class="d2l-body-compact overall-score-container">`

jeroach: Also, we try to prefix class names with the component

E.g. `consistent-evaluation-tii-grade-mark-overall-score-container`

mbosiljevac: Will do! Thanks for the info!

jeroach (approved):

JamieTran (approved):

PR #283, opened 2021-02-09
US122418/mbosilj add email action to profile card.

mbosiljevac:
Note: I can also add the message window closing when the user navigates away from the user submission if we'd like that.

This adds user progress, email and pager(IM) actions.


components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -125,12 +130,26 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
 		}
 	}
 
+	_getInstantMessageHref() {
+		if (this.isGroupActivity) {
+			if (super._entity && super._entity._entity.getSubEntityByRel(pagerRel)) {
```
JamieTran, line 135: Could this be extracted into the `controller`? If my understanding this correct, `enrolledUser` isn't defined when it's a group assignment?

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -250,6 +250,50 @@ export class ConsistentEvaluationHrefController {
...
 
+	async getEnrolledUser() {
+		const root = await this._getRootEntity(false);
+		if (root && root.entity) {
+			const enrolledUserHref = this._getHref(root.entity, enrolledUserRel);
+			const groupHref = this._getHref(root.entity, groupRel);
+			if (enrolledUserHref) {
+				const enrolledUserEntity = await this._getEntityFromHref(enrolledUserHref, false);
+				const pagerEntity = enrolledUserEntity.entity.getSubEntityByRel(pagerRel);
+				const userProgressEntity = enrolledUserEntity.entity.getSubEntityByRel('https://api.brightspace.com/rels/user-progress');
```
JamieTran, line 261: nit: we should probably stick to the Constants pattern

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -250,6 +250,50 @@ export class ConsistentEvaluationHrefController {
...
+
+				const emailEntity = enrolledUserEntity.entity.getSubEntityByRel(emailRel, false);
+				const emailPath = emailEntity.properties.path;
```
JamieTran, line 264: I think this breaks if `emailEntity = undefined`

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -250,6 +250,50 @@ export class ConsistentEvaluationHrefController {
...
+				};
+			} else if (groupHref) {
+				const groupEntity = await this._getEntityFromHref(groupHref, false);
+				const pagerEntity = groupEntity.entity.getSubEntityByRel(pagerRel);
+				let pagerPath = undefined;
+				if (pagerEntity) {
+					pagerPath = pagerEntity.properties.path;
+				}
+				return {
+					enrolledUserHref,
```
JamieTran, line 288: I think `enrolledUserHref` would always be `undefined` here

components/controllers/ConsistentEvaluationHrefController.js
vix633: https://github.com/Brightspace/d2l-hypermedia-constants/pull/136#pullrequestreview-586693945

PR out for this, we can have the string for now

components/controllers/ConsistentEvaluationHrefController.js
mbosiljevac: yep whoops should be `groupHref` here, good catch!

components/header/consistent-evaluation-user-profile-card.js
```diff
@@ -12,24 +12,91 @@ export class ConsistentEvaluationUserProfileCard extends LocalizeConsistentEvalu
...
+
+		this.emailPopout = window.open(
+			this.emailHref,
+			'messagePopout',
+			'width=1000,height=1000,scrollbars=no,toolbar=no,screenx=0,screeny=0,location=no,titlebar=no,directories=no,status=no,menubar=no'
+		);
+	}
+
 	render() {
+		console.log(this.emailHref);
```
JamieTran, line 88: debug code?

JamieTran (approved):

PR #285, opened 2021-02-09
US122418/update constant and user-profile-card

mbosiljevac:
Update d2l-hypermedia-constants and use Rels.userProgress
Update user-profile-card and use show-email, show-im and show-progress


jeroach (approved):
LGTM

PR #288, opened 2021-02-10
Add profile pic and display name to user-profile-card

JamieTran:
Removed tagline from user-profile-card
Added profile pic and display name

image

### DIFF
--- a/components/controllers/ConsistentEvaluationHrefController.js
+++ b/components/controllers/ConsistentEvaluationHrefController.js
@@ -261,11 +261,17 @@ export class ConsistentEvaluationHrefController {
 				const userProgressEntity = enrolledUserEntity.entity.getSubEntityByRel(Rels.userProgress);
 				const emailEntity = enrolledUserEntity.entity.getSubEntityByRel(emailRel, false);
 				const userProfileEntity = enrolledUserEntity.entity.getSubEntityByRel(Rels.userProfile);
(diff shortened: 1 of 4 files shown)


vix633:
> Removed tagline from user-profile-card
> Added profile pic and display name
> 
> image

Where's the email button at?

JamieTran:
> > Removed tagline from user-profile-card
> > Added profile pic and display name
> > image
> 
> Where's the email button at?

This user doesn't have an associated email

vix633 (approved):

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -279,6 +285,7 @@ export class ConsistentEvaluationHrefController {
 					userProfilePath = userProfileEntity.properties.path;
 				}
 				return {
+					displayName,
```
mbosiljevac, line 288: This might be out of the scope of this ticket, but this might break if you have a group assignment (the display name would probably be shown as undefined) You might want to also return a `display-name` in the block below this but I'm not positive if the api would let you

JamieTran: I guess we should have asked this sooner but the user-context-bar will be disabled on group assignments D: 
[convo with Dan](https://d2l.slack.com/archives/CSL64JP9U/p1612974422093300)

mbosiljevac (approved):
LGMT!

PR #289, opened 2021-02-10
US124654/mbosilj make tii % similar a link.

mbosiljevac:


components/left-panel/consistent-evaluation-tii-similarity.js
```diff
@@ -104,14 +104,16 @@ export class ConsistentEvaluationTiiSimilarity extends LocalizeConsistentEvaluat
 		const renderScore = Math.round(this.score * 100);
 
 		return html`
-			<div
+			<a
```
mbosiljevac, line 107: Note if we want to keep this as a div I can use `role=link` instead

components/left-panel/consistent-evaluation-tii-similarity.js
```diff
@@ -104,14 +104,16 @@ export class ConsistentEvaluationTiiSimilarity extends LocalizeConsistentEvaluat
...
 				class="d2l-consistent-evaluation-tii-similarity-bar"
 				@click=${this._onSimilarityBarClick}
 				title="${this.localize('turnitinViewReport', 'file', this.fileName)}"
+				aria-label="${renderScore}% Similar"
```
jeroach, line 111: We'll need a lang term for this

components/left-panel/consistent-evaluation-tii-similarity.js
jeroach: if it looks the same, an `<a>` tag is fine

jeroach (approved):

JamieTran (approved):

PR #290, opened 2021-02-11
Fix profile card not disappearing when mousing away

Scy-D2L:
This fixes the problem where the profile card wouldn't disappear when mousing away from it and adds an invisible area between the user context and the profile card that will keep it open. The area is as small as it could be made, and dismisses the card on click so it won't block access to other elements.


vix633:
Let's just ask Dan if he's down with occluding/partially occluding the name

JamieTran (approved):

PR #299, opened 2021-02-18
Group Options menu keyboard accessibility

Scy-D2L:
Use item select to trigger group menu actions instead of onClick to capture both keyboard and mouse.


jeroach (approved):
LGTM

PR #300, opened 2021-02-22
showing instant message button only if href is present

vix633:
We want the updated lifecycle method here because the first few updates, groupInfo is still null as we wait for a response from the api.

### DIFF
--- a/components/header/d2l-consistent-evaluation-lcb-user-context.js
+++ b/components/header/d2l-consistent-evaluation-lcb-user-context.js
@@ -120,6 +120,12 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
 		this._groupMembersItemID = getUniqueId();
 	}
 
+	updated() {
+		if (this.groupInfo) {
+			this._hasPagerPath = this.groupInfo.pagerPath ? true : false;
+		}
+	}
+
 	set _entity(entity) {
 		if (this._entityHasChanged(entity)) {
 			this._onActorEntityChanged(entity);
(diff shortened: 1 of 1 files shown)


components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -285,7 +291,7 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
 					<d2l-menu @d2l-menu-item-select=${this._onGroupOptionSelect} label=${this.localize('groupOptions')}>
 						<d2l-menu-item id=${this._groupEmailItemID} text=${this.localize('emailGroup')}></d2l-menu-item>
 						<d2l-menu-item id=${this._groupMembersItemID} text=${this.localize('seeAllGroupMembers')}></d2l-menu-item>
-						<d2l-menu-item id=${this._groupIMItemID} text=${this.localize('instantMessage')}></d2l-menu-item>
+						<d2l-menu-item id=${this._groupIMItemID} text=${this.localize('instantMessage')} ?hidden=${!this._hasPagerPath}></d2l-menu-item>
```
jmurray-d2l, line 294: can we not just use `this.groupInfo.pagerPath` here instead of creating another variable for it?

vix633: No, for the same reason as noted in the pr description. If we try to evaluate the .pagerPath prop on groupInfo before the component has it, we'll get a `Cannot read property 'pagerPath' of undefined` for every update before we get groupInfo returned from the backend

jmurray-d2l: how about `this.groupInfo && this.groupInfo.pagerPath` then?

vix633: Sure, that should behave the same way

mbosiljevac (approved):
LGTM 🎉

jmurray-d2l (approved):

PR #302, opened 2021-02-22
use new user progress path

vix633:
This now goes to assignments portion of user progress as expected

### DIFF
--- a/components/controllers/ConsistentEvaluationHrefController.js
+++ b/components/controllers/ConsistentEvaluationHrefController.js
@@ -2,7 +2,7 @@ import 'd2l-polymer-siren-behaviors/store/entity-store.js';
 import { actorRel, alignmentsRel, assessmentRel, assessmentRubricApplicationRel,
 	assessorUserRel, assignmentSubmissionListRel, demonstrationRel, editSpecialAccessApplicationRel,
 	emailRel, enrolledUserRel, evaluationRel, groupRel, nextRel, pagerRel, previousRel, rubricRel,
-	userProgressOutcomeRel, userRel, viewMembersRel } from './constants.js';
+	userProgressAssessmentsRel, userProgressOutcomeRel, userRel,  viewMembersRel } from './constants.js';
 import { Classes, Rels } from 'd2l-hypermedia-constants';
 
 export const ConsistentEvaluationHrefControllerErrors = {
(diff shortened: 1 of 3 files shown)


jeroach (approved):

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -286,10 +286,10 @@ export class ConsistentEvaluationHrefController {
 			if (enrolledUserHref) {
 				const enrolledUserEntity = await this._getEntityFromHref(enrolledUserHref, false);
 				const pagerEntity = enrolledUserEntity.entity.getSubEntityByRel(pagerRel);
-				const userProgressEntity = enrolledUserEntity.entity.getSubEntityByRel(Rels.userProgress);
 				const emailEntity = enrolledUserEntity.entity.getSubEntityByRel(emailRel, false);
 				const userProfileEntity = enrolledUserEntity.entity.getSubEntityByRel(Rels.userProfile);
 				const displayNameEntity = enrolledUserEntity.entity.getSubEntityByRel(Rels.displayName);
+				const userProgressEntity = root.entity.getSubEntityByRel(userProgressAssessmentsRel, false);
```
amauryperez, line 292: any particular reason this is under `getEnrolledUser` instead of `getHrefs`? Looking at the pattern we were using, all other hrefs that are sub-entities of AAU are returned in `getHrefs` (i.e. submissions list, rubric popup, special access, etc)

vix633: Yeah so I kind of thought about it as well. I think it belongs here because it pertains specifically to this user's progress and the object that this method returns has all the props that pertain to the user. 

I definitely feel weird about it being in this method as well, however only because of the method's name, not because of anything functional.

I'm open to simply moving it to `getHrefs` if you feel its cleaner. I agree that functionally it could live there; contextually I think it belongs in the function that's already returning an object that describes the current user being evaluated.

amauryperez: @jeroach do you have any preference?

amauryperez (approved):

components/controllers/ConsistentEvaluationHrefController.js
jeroach: I can see both sides.

Logically, what Amaury has said makes sense since we don't need to check `if (enrolledUserHref) {` for that call, but organizationally what Viktor's saying makes sense.

@amauryperez is your comment a blocker? If so, let's change it - if not, let's merge as is.

vix633: I'll merge, if we want this changed I'm happy to PR the change

PR #303, opened 2021-02-23
US116472/disable publish button if anonymous and no published submissions exist.

mbosiljevac:
Note: the button is disabled when the page loads (if it's enable on load then its possible for users to publish before the button is disabled). But this means that non-anonymous assignments see a disabled publish button for a bit until the page fully loads (this is probably fine but I just wanted to call it out)
image


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -352,6 +353,30 @@ export class ConsistentEvaluationHrefController {
...
+	async getAnonymousInfo() {
+		const root = await this._getRootEntity(false);
+		if (root && root.entity) {
+			const assignmentHref = this._getHref(root.entity, assignmentRel);
+			if (assignmentHref) {
+				const assignmentEntity = await this._getEntityFromHref(assignmentHref);
+				if (assignmentEntity.entity.hasSubEntityByRel(anonymousMarkingRel)) {
+					const anonymousAssignmentEntity = assignmentEntity.entity.getSubEntityByRel(anonymousMarkingRel);
+					const isAnonymous = anonymousAssignmentEntity.hasClass(checkedClassName);
+					const assignHasPublishedSub = anonymousAssignmentEntity.hasClass(publishedClassName);
```
vix633, line 365: we can probably name this `assignmentHasPublishedSubmission`

jeroach:
> But this means that non-anonymous assignments see a disabled publish button for a bit until the page fully loads (this is probably fine but I just wanted to call it out)

What kind of delay are we talking about? Regardless, I'm leaning towards the other option (publish button is not disabled at first) as non-anonymous assignments are by in large the most common use case

mbosiljevac:
> What kind of delay are we talking about? Regardless, I'm leaning towards the other

it's pretty short. It's possible to click the publish button but unlikely that users will. I can change it to default to enabled!

jeroach (approved):
LGTM

PR #312, opened 2021-02-25
US124842/add mobile overflow menu button.

mbosiljevac:
Note: the more options button is shifted over to keep the margins consistent.
image


components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -144,6 +149,31 @@ export class ConsistentEvaluationRightPanel extends LocalizeConsistentEvaluation
 		this.canRecordFeedbackAudio = false;
 		this.useNewHtmlEditor = false;
 		this.rubricsOpen = 0;
+		this.mobileMediaQuery = window.matchMedia('(max-width: 767px)');
```
vix633, line 152: I don't actually know enough about media queries to review this. @AlexBedley could you peek at this?

jeroach (commented):
I have a feeling that we should be handling the mobile menu via css media queries alone.

e.g.
- different class names on the default menu and mobile menu
- on standard screen size, default is visible, mobile is hidden
- once the screen width is small enough, hide default show mobile

Thoughts?

components/right-panel/consistent-evaluation-right-panel.js
jeroach: Yeah I don't think this is the approach we want to take. See my larger comment.

mbosiljevac:
> I have a feeling that we should be handling the mobile menu via css media queries alone.
> 
> e.g.
> 
> * different class names on the default menu and mobile menu
> * on standard screen size, default is visible, mobile is hidden
> * once the screen width is small enough, hide default show mobile
> 
> Thoughts?

makes sense! I'll play around with that idea and get something working :)

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -105,6 +105,23 @@ export class ConsistentEvaluationRightPanel extends LocalizeConsistentEvaluation
 
 	static get styles() {
 		return  css`
+			.d2l-consistent-evaluation-right-panel-overflow-menu-mobile {
+				display: none;
+			}
+
+			@media (max-width: 767px) {
```
JamieTran, line 112: nit: We can group classes together that have the same `@media` queries
eg:

```
@media (max-width: 767px) {
		.d2l-consistent-evaluation-right-panel-overflow-menu-mobile {
			display: inline-block;
			margin: 0.5rem 1.25rem 0.5rem 1.25rem;
		}
		.d2l-consistent-evaluation-right-panel-clearfix {
			display: none;
		}
	}
```

JamieTran (approved):

PR #322, opened 2021-03-01
US124816/DE42623 Limit submission items menus to submission list context

Scy-D2L:
Fix for US124816/DE42623, stop button from appearing on profile card without impacting `More Options` menu

image

Giving the `Submissions List` it's own [Stacking Context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) stops the popups from appearing over top of the profile card.


JamieTran (approved):
Looks good!

PR #324, opened 2021-03-02
US125337/navigate to file if one file/submission.

mbosiljevac:


components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -115,10 +115,11 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 	async updated(changedProperties) {
 		super.updated();
 
-		if ((changedProperties.has('currentFileId') || changedProperties.has('token') || changedProperties.has('submissionInfo'))
+		if ((changedProperties.has('currentFileId') || changedProperties.has('token'))
```
mbosiljevac, line 118: I'm a little hesitant to remove this check but it looks like if we leave it in when you change users (with the arrows) after viewing a users submission this gets called after the submissionInfo has been updated but the currentFileId hasn't updated (so it's the file associated with the previous user). This just gives a console error that the file couldn't be fetched but it felt best to remove it and just have `currentFileId` reflect what file is being currently viewed and when this changes.

components/consistent-evaluation.js
```diff
@@ -216,12 +234,12 @@ export class ConsistentEvaluation extends LitElement {
 				special-access-href=${ifDefined(this._childHrefs && this._childHrefs.specialAccessHref)}
 				return-href=${ifDefined(this.returnHref)}
 				return-href-text=${ifDefined(this.returnHrefText)}
-				current-file-id=${ifDefined(this.currentFileId)}
 				data-telemetry-endpoint=${ifDefined(this.dataTelemetryEndpoint)}
 				logging-endpoint=${ifDefined(this.loggingEndpoint)}
 				rubric-popout-location=${ifDefined(this._childHrefs && this._childHrefs.rubricPopoutLocation)}
 				download-all-submissions-location=${ifDefined(this._childHrefs && this._childHrefs.downloadAllSubmissionLink)}
 				edit-activity-path=${ifDefined(this._editActivityPath)}
+				.currentFileId=${this.currentFileId}
```
jeroach, line 242: If we're changing this from an attribute to a property, I would have expected that we'd need to change `d2l-consistent-evaluation-page` to match, no?

components/left-panel/consistent-evaluation-left-panel.js
jeroach: So removing this prevents a console error?

components/consistent-evaluation.js
mbosiljevac: I think it doesn't matter, we pass this variable the same way between `consistent-evaluation-page` and `consistent-evaluation-left-panel`  but I can remove it so its more clear!

components/left-panel/consistent-evaluation-left-panel.js
mbosiljevac: yep!

jeroach (approved):

PR #326, opened 2021-03-02
US124891/add file conversion warning.

mbosiljevac:
image


components/left-panel/consistent-evaluation-evidence-file.js
```diff
@@ -119,6 +128,21 @@ export class ConsistentEvaluationEvidenceFile extends LitElement {
...
+
+	_renderToast() {
+		const toastMessage = this.localize('fileConversionWarning');
+		return html`<d2l-alert-toast
+			?open=${this._shouldDisplayToast()}
+			@d2l-alert-toast-close=${this._onToastClose}>${toastMessage}</d2l-alert-toast>`;
+	}
+
+	_shouldDisplayToast() {
+		return this.fileExtension !== pdfExtension && this.displayConversionWarning && this._displayToast;
```
jeroach, line 143: Two questions:
- where is `this.displayConversionWarning` originally set?
- regarding `this.fileExtension !== pdfExtension` - converted files are .pdf so is this check covering the case where the file was a .pdf to begin with?

mbosiljevac: * the displayConversionWarning will be a web component property given by the backend in the `Evaluationview.Renderer.cs` (similar to `use-new-html-editor`)
* yep! It checks if the file was originally a pdf and only displays the warning if the original file was **not** a pdf!

jeroach (approved):
LGTM

JamieTran (approved):

PR #332, opened 2021-03-04
US122137/add skeleton to right panel.

mbosiljevac:
image


components/right-panel/consistent-evaluation-feedback-presentational.js
```diff
@@ -195,7 +238,9 @@ class ConsistentEvaluationFeedbackPresentational extends LocalizeConsistentEvalu
 				: null;
 			this._setFeedbackSummaryInfo();
 			return html`
+				${this._renderSkeleton()}
 				<d2l-consistent-evaluation-right-panel-block
+					class="d2l-consistent-evaluation-feedback-block"
```
jeroach, line 243: I think we want `aria-hidden="${this.skeleton}"` here as well

components/right-panel/consistent-evaluation-outcomes.js
```diff
@@ -26,12 +27,65 @@ class ConsistentEvaluationOutcomes extends LocalizeConsistentEvaluation(LitEleme
...
+				<div class="d2l-skeleton-outcomes-footer d2l-skeletize"></div>
+			</div>
+		</div>`;
 	}
 
 	render() {
 		return html`
+			${this._renderSkeleton()}
 			<d2l-consistent-evaluation-right-panel-block
+				class="d2l-consistent-evaluation-outcomes-block"
```
jeroach, line 88: I think we want `aria-hidden="${this.skeleton}"` here as well

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -329,7 +385,9 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 
 	render() {
 		return html`
+			${!this.isPopout ? this._renderSkeleton() : ''}
 			<d2l-consistent-evaluation-right-panel-block
+				class="d2l-consistent-evaluation-rubric-block"
```
jeroach, line 390: I think we want `aria-hidden="${this.skeleton}"` here as well

jeroach (commented):
vdiff tests look good - so does the full render

Just a few missing attributes, and this is good to go :)

jeroach (approved):

JamieTran (approved):
Nice work! I wonder if we want loading skeletons for the mobile view as well... That shouldn't be encompassed in this PR though :)

mbosiljevac:
> Nice work! I wonder if we want loading skeletons for the mobile view as well... That shouldn't be encompassed in this PR though :)

Ye for sure! That's something we should probably do

PR #334, opened 2021-03-04
Jroach/us125532

jeroach:
Okay this is ready for review :) Here is a summary of what's going on:

- Add error handling logic to `performSirenAction`
- Refactor common code to helpers for various `save, publish, update, retract` methods
- Add logic for error state toast messages (including lang terms)
- Remove `_logError` so that we use the lms level error handler


components/consistent-evaluation-page.js
```diff
@@ -501,13 +509,15 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		await this._mutex.dispatch(
 			async() => {
 				const entity = await this._controller.fetchEvaluationEntity(false);
-				this.evaluationEntity = await this._controller.save(entity);
+				const newEvaluationEntity = await this._controller.save(entity);
```
jeroach, line 512: Assign to a temporary variable to avoid overwriting the valid `evaluationEntity`

components/consistent-evaluation-page.js
```diff
@@ -501,13 +509,15 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
 				this._currentlySaving = false;
-				if (!(this.evaluationEntity instanceof Error)) {
-					this._showToast(this.localize('saved'));
-					this._fireSaveEvaluationEvent();
+
+				if (!newEvaluationEntity) {
+					this._showToast(this.localize('saveError'), true);
 				} else {
-					this._showToast(this.localize('saveError'));
+					this.evaluationEntity = newEvaluationEntity;
```
jeroach, line 518: Only update if the save was successful

components/controllers/ConsistentEvaluationController.js
```diff
@@ -100,7 +100,14 @@ export class ConsistentEvaluationController {
 	}
 
 	async _performSirenAction(action, field = null) {
-		return await performSirenAction(this.token, action, field, true);
+		const response = await performSirenAction(this.token, action, field, true)
+			.catch((error) => {
```
jeroach, line 104: Catch API errors, and (eventually) log them.

components/controllers/ConsistentEvaluationController.js
```diff
@@ -272,7 +279,12 @@ export class ConsistentEvaluationController {
 			this._logError(ConsistentEvaluationControllerErrors.INVALID_EVALUATION_ENTITY);
 		}
 
-		return await this._performAction(evaluationEntity, saveActionName);
+		const newEvaluationEntity = await this._performAction(evaluationEntity, saveActionName);
```
jeroach, line 282: Refactor the `save()` method to be more of a `trySave()`

components/consistent-evaluation-page.js
```diff
@@ -657,6 +676,8 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		return html`<d2l-alert-toast
 			?open=${this._displayToast}
 			button-text=""
+			?no-auto-close=${this._toastNoAutoClose}
```
amauryperez, line 679: what happens with `no-auto-close`, does it stay there until you reload the page or something?

components/consistent-evaluation-page.js
amauryperez: This makes sense

amauryperez: I was looking at the code and it seems that the code for `_saveEvaluation`, `_updateEvaluation`, `publish` and `retract` are all pretty much identical, would it make sense to extract the common code to a single function?

components/consistent-evaluation-page.js
jeroach: Yeah it will stay until the user dismisses it. This is what Dan wanted for the error states.

components/consistent-evaluation-page.js
jeroach: Yeah that's a good call.

I just wanted to do a POC to gain consensus before going full refactor :)

components/controllers/ConsistentEvaluationController.js
vix633: What do api errors look like now? I haven't seen the PR for them but the backend guys did change something there to make them well formed json. Do we need to parse that how? What fields are there?

components/controllers/ConsistentEvaluationController.js
```diff
@@ -100,7 +100,14 @@ export class ConsistentEvaluationController {
 	}
 
 	async _performSirenAction(action, field = null) {
-		return await performSirenAction(this.token, action, field, true);
+		const response = await performSirenAction(this.token, action, field, true)
+			.catch((error) => {
+				console.log(error);
+				//this._logError(error.message);
```
vix633, line 106: The log error stuff can get pulled out entirely, you can just replace `this._logError` with throws. The new global error handler will catch them now hopefully without hosing the app.

lang/en.js
```diff
@@ -78,6 +78,7 @@ export default {
 	"rubricsAssess": "Assess Rubrics for {username}",
 	"saveBtn": "Save",
 	"saveDraft": "Save Draft",
+	"saveError": "OMG ITS ON FIRE",
```
vix633, line 81: Nice

components/controllers/ConsistentEvaluationController.js
vix633: Found the pr: https://github.com/Brightspace/lms/pull/7059/files

lang/en.js
jeroach: If error messages were honest

components/controllers/ConsistentEvaluationController.js
jeroach: Across the entire app? e.g. remove the current wrapper we have?

vix633: so far `logError` is just used here, unless I merged the app-global handler. But basically I want it gone entirely, so we just let errors be errors and the built in handler that gets injected on app load handles them.

components/controllers/ConsistentEvaluationController.js
jeroach: Jamie M and I talked about this offline - yeah they're json and we want to extract the message and statusCode

components/controllers/ConsistentEvaluationController.js
```diff
@@ -100,15 +93,24 @@ export class ConsistentEvaluationController {
 	}
 
 	async _performSirenAction(action, field = null) {
-		return await performSirenAction(this.token, action, field, true);
+		const response = await performSirenAction(this.token, action, field, true)
+			.catch((error) => {
+				const message = error.message;
+				const statusCode = error.json.properties.status;
+
+				//throw new Error(statusCode + message);
```
jeroach, line 101: `throw new Error` is still blowing up the page :(

I've pulled in Dave B's fix so not sure why it's still freezing.

components/controllers/ConsistentEvaluationController.js
jeroach: Removed it

components/consistent-evaluation-page.js
```diff
@@ -517,90 +516,93 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
-					this._showToast(this.localize('retractError'));
-				}
+
+				this.checkEvaluationEntityAndDisplayToast(newEvaluationEntity, 'retractError', 'retracted');
 				this.submissionInfo.evaluationState = draftState;
 			}
 		);
 	}
 
+	async _flushAndWait() {
```
jmurray-d2l, line 569: Nice, this is a great refactor

components/controllers/ConsistentEvaluationController.js
```diff
@@ -268,33 +268,31 @@ export class ConsistentEvaluationController {
...
-		}
-		return await this._performAction(evaluationEntity, publishActionName);
+		return this._tryPerformAction(evaluationEntity, publishActionName);
 	}
 
 	async retract(evaluationEntity) {
+		return this._tryPerformAction(evaluationEntity, retractActionName);
+	}
+
+	async _tryPerformAction(evaluationEntity, actionName) {
```
amauryperez, line 286: I like the changes. This method is the one that is a bit weird because it can either return the entity object or a boolean. I know JavaScript allows for whatever but I don't know, is it common that we do that?, since it is called `try` someone would expect that it always returns a boolean

components/consistent-evaluation-page.js
amauryperez: some much clean now

components/controllers/ConsistentEvaluationController.js
jeroach: I've renamed this anyway with moving other stuff around.

components/controllers/ConsistentEvaluationController.js
```diff
@@ -100,15 +93,21 @@ export class ConsistentEvaluationController {
 	}
 
 	async _performSirenAction(action, field = null) {
-		return await performSirenAction(this.token, action, field, true);
+		let newEvaluationEntity;
+		try {
```
jeroach, line 97: I moved this `try/catch` block up here as there are other error paths that `_performAction` can go down (e.g. missing action name).

This way, we know that the error will have the properties we want to log, since it's coming directly from the API.

components/controllers/ConsistentEvaluationController.js
```diff
@@ -100,15 +93,21 @@ export class ConsistentEvaluationController {
...
+			newEvaluationEntity = await performSirenAction(this.token, action, field, true);
+		} catch (e) {
+			console.error(`HTTP Status Code: "${e.json.properties.code} ${e.json.properties.status}" Message: "${e.message}"`);
```
jmurray-d2l, line 100: Do these console.errors get logged? Ideally we are able to see these messages in the logs

vix633: Josh and I talked about this at length and we settled on these needing to be logged in the back end. Since the failure already happens on the back end, it knows the status code its returning and can log there. Logging in here can be done, but we felt it would be better if it was logged on the back end, since thats closer to where the error actually occurs.

jmurray-d2l: That makes sense, we do get browser information if we log in the front end, that we won't get in the back end.  I would say though, there isn't any reason to log this in the console.

jmurray-d2l (approved):

amauryperez (approved):

JamieTran (approved):

PR #335, opened 2021-03-05
DE42730 - Adding alt text to all status indicators for submission items

jmurray-d2l:


amauryperez (approved):

mbosiljevac (approved):

dbatiste:
What's the goal here?  `alt-text` is not an attribute that `d2l-status-indicator` supports.

jmurray-d2l:
> What's the goal here? `alt-text` is not an attribute that `d2l-status-indicator` supports.

Our accessibility review brought these status indicators up as something that wasn't read out.  I wasn't aware of the lack of support, is there a better way to do this?

dbatiste:
> Our accessibility review brought these status indicators up as something that wasn't read out. I wasn't aware of the lack of support, is there a better way to do this?

Interesting... the `d2l-status-indicator` just renders the text to the page with styling.  It should get announced like other text on the page, discoverable when the page is normally read by the screenreader, or when using the virtual cursor.  I did a quick test and is seems to announce correctly in the component demo page.  It is not focusable though so it won't be announced as the user tabs around the page - perhaps that's why it was thought to be an a11y issue?

That aside, `alt-text` isn't a global attribute supported by browsers either.  Perhaps the idea was to utilize `alt`, however that is only an attribute for `img` elements.  If there is an issue with announcing the status indicator text, we should address it in `d2l-status-indicator` for all consumers.

jmurray-d2l:
Okay cool - I will look into it more, but won't fix it this way.  I really appreciate the review, thank you :)

dbatiste:
> Okay cool - I will look into it more, but won't fix it this way. I really appreciate the review, thank you :)

Sounds good.  No problem!  If you need help or think something needs to change in the status-indicator, feel free to reach out.  :)

PR #337, opened 2021-03-05
DE42728 - Differentiate between the two '...' buttons

amauryperez:


lang/en.js
```diff
@@ -43,6 +43,8 @@ export default {
 	"markRead": "Mark as Read",
 	"markUnread": "Mark as Unread",
 	"moreOptions": "More Options",
```
amauryperez, line 45: I'm not deleting `moreOptions` because we already have the translations and I'm sure we will use it for Discussions/Quizzing

JamieTran (approved):
LGTM

mbosiljevac (approved):

PR #342, opened 2021-03-09
US125717 blur immediately after update/publish.

mbosiljevac:


components/consistent-evaluation-page.js
```diff
@@ -252,6 +252,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		this._isPublishClicked = false;
 		this._shouldWaitForAnnotations = false;
 		this._currentlySaving = false;
+		this._hasClickedButton = false;
```
amauryperez, line 255: I cannot find where this one is used?

mbosiljevac: whoops yep this is leftover from something else I was trying. I'll take it out thanks!

amauryperez (approved):

PR #347, opened 2021-03-09
- Adding off screen elements for screen reader when files are marked …

jmurray-d2l:
…as read / unread or flagged / unflagged


components/left-panel/consistent-evaluation-submission-item.js
```diff
@@ -380,10 +386,15 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
...
+	_renderFlaggedStatus(flag, fileName) {
+		let flagIndicator = html``;
+		let langTermText = 'unflaggedFile';
 		if (flag) {
-			return html`<d2l-status-indicator state="alert" text="${this.localize('flagged')}"></d2l-status-indicator>`;
+			langTermText = 'flaggedFile';
+			flagIndicator =  html`<d2l-status-indicator state="alert" text="${this.localize('flagged')}"></d2l-status-indicator>`;
 		}
+
+		return html`${flagIndicator} <d2l-offscreen aria-live="assertive">${this.localize(langTermText, 'fileName', fileName)}</d2l-offscreen>`;
```
Scy-D2L, line 397: Ah, ok - I get how this solution came up from the suggestion I made in Slack, but seeing it implemented I do see some flaws. The way this is set up, every single item will say 'unflagged' when you select it and read out its filename again (sample item reading from NVDA: _clickable testing document, flagged testing document, flagged_ ) - which might not be ideal. Somehow I'd been imagining that the _flagged_ pill wasn't being read out, which is completely my bad. 🤦🏼 

Leveraging `aria-live` to make it clear the item was updated with the offscreen item works, but it feels a bit messy to have this extra info always present when the screenreader already reads the 'flagged' pill on flagged items. I think using a [toast alert](https://github.com/BrightspaceUI/core/tree/master/components/alert) to announce flagged/unflagged, read/unread actions instead would be cleaner audibly, but that's something we ought to check with Dan about since it has a visual impact too (not necessarily a bad one).

PR #348, opened 2021-03-09
await all calls to _flushAndWait

jeroach:
Megan caught this - we need to await for the entire `_flushAndWait` block.

I confirmed the JS control flow logic with Alex. The code calling `_flushAndWait` will not wait for the inner `await` calls to finish before proceeding.


vix633 (approved):
Really good catch @mbosiljevac

PR #349, opened 2021-03-10
US125716/wrap update info in dispatch.

mbosiljevac:
It looks like the code where we assign all the new information also needs to be in a dispatch to the mutex to get the behaviour we want. 
This shouldn't have any big impact on performance (since it should only wait if there is something in the mutex and the mutex will only ever have this or a change in href from onNextStudentClick or onPreviousStudentClick)
Let me know if I'm missing anything!


jeroach (approved):
Makes sense to me.

jmurray-d2l (approved):

PR #352, opened 2021-03-10
US125660/move right panel skeleton to its own file.

mbosiljevac:
Chatted with Dan and decided to show the entire skeleton on page load, even if the user won't end up seeing rubrics/grades etc.

right_panel_skeleton


jeroach (approved):
LGTM

JamieTran (approved):

PR #358, opened 2021-03-15
US125735/create file to show discussions/assignments.

mbosiljevac:
Currently just shows the assignments page, no logic to switch between assignments and discussions.


components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,78 @@
...
+import { html, LitElement } from 'lit-element';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { LocalizeConsistentEvaluation } from '../../lang/localize-consistent-evaluation.js';
+import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+
+export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsistentEvaluation(LitElement)) {
+
+	static get properties() {
+		return {
+			submissionInfo: {
```
amauryperez, line 12: There are a few properties here that are for assignments only. should we find a way to not have them at this level?

mbosiljevac: hmm since a lot of this information is calculated and modified at the consistent-evaluation-page level (like submissionInfo or currentFileId) I don't think can avoid passing it at this level. We could wrap this information into a object (`assignmentInformation`) though which might be a bit cleaner?

amauryperez: What is the new workflow when we load a discussions AAU? we probably don't want to look for fileId or submissions right? Is there a way to avoid processing that information on `consistent-evaluation-page`?

mbosiljevac: I haven't fully fleshed that out but it looks like most of the calls to the controller to initially set the data is in the `consistent-evaluation.js` file and most calls in the `consistent-evaluation-page.js` are to modify the data. I was thinking we could do a check in `consistent-evaluation.js` to see if its discussion/assignments to choose what information to fetch. 

In terms of `consistent-evaluation-page.js` before we modify the data we _could_ check if we're dealing with an assignment/discussion but that feels like a lot of checks so I'm not sure what the best way to deal with this is :/

Also I'm not sure what AAU stands for :(

components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -0,0 +1,77 @@
...
+				attribute: false,
+				type: Object
+			},
+			token: { type: Object },
+			fileNonViewable: {
```
jeroach, line 17: `fileNonViewable` doesn't seem to be used

components/left-panel/consistent-evaluation-left-panel.js
jeroach: Yeah I'm not sure either what they best way of handling this is.

I think what Megan's done is a good first step, and we can consider optimizing in future.

Thoughts?

jeroach: Also I think AAU is `actor activity usage`

components/left-panel/consistent-evaluation-left-panel.js
mbosiljevac: good catch! ty Josh I'll take it out :)

components/left-panel/consistent-evaluation-left-panel.js
mbosiljevac: chatted with @amauryperez and he agreed to wait to optimize until a later time!

amauryperez (approved):

jeroach (approved):

PR #359, opened 2021-03-15
US125729/update discussion display logic

mbosiljevac:
image


amauryperez:
This looks very good. Do we usually commit the `package-lock.json` file?

mbosiljevac:
> This looks very good. Do we usually commit the `package-lock.json` file?

Good point! Looking at the changes it seems like this is from a npm update and unrelated to the sort by component so I'll take out the changes to the `package-lock.json`!

amauryperez (approved):

JamieTran (approved):

components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
```diff
@@ -0,0 +1,129 @@
...
+	constructor() {
+		super();
+		this.isReply = false;
+	}
+
+	_renderRepliedTo() {
+		return html `<div class="d2l-consistent-evaluation-reply-to-container">
+				<d2l-icon class="d2l-consistent-evaluation-evidence-body-reply-icon" icon="d2l-tier1:reply"></d2l-icon>
+				<span class="d2l-body-compact d2l-consistent-evaluation-evidence-body-reply-to-text">${this.localize('repliedTo')}</span>
+				<span class="d2l-body-compact d2l-link d2l-truncate">Hannah Beiley's post</span>
```
jeroach, line 86: I wonder if we should extract all of the sample values out, and use attributes instead. We're going to have to do that eventually anyway.

Then we could provide all the sample values as attributes in a demo page.

Thoughts? Perhaps I'm missing some context.

mbosiljevac: sure! I can extract them out! I think I only hesitated because I didn't know how the backend would send the information but I don't think that matters too much!

PR #360, opened 2021-03-15
us125728 - refactor file structure discussions

JamieTran:
Looks messy but to summarize:

- Create assignments and discussions folders under left-panel
- Move all previous left-panel files into assignments
- Fix all references to the above changes (demo & vdiff & component files)


mbosiljevac (approved):
LGTM!

components/left-panel/assignments/consistent-evaluation-left-panel.js
```diff
@@ -6,11 +6,11 @@ import './consistent-evaluation-outcomes-overall-achievement.js';
 import { appId, toggleIsReadActionName } from '../controllers/constants.js';
```
mbosiljevac, line 6: This should probably be changed too!

JamieTran: Nice catch, missed it in the merge

mbosiljevac (approved):

PR #363, opened 2021-03-16
US125727/add discussion-evidence-body component.

mbosiljevac:
I wanted to get some eyes on what I've done so far before this gets any bigger. 
This PR:
* adds v-diff tests and a demo page
* add the visual styling for discussion posts/replies

Things that still need to be done:
* allow star ratings (will most likely pull this out into its own component)
* Un-hardcode all the strings and hook up front end to get information from back end
* use data provided by the backend to determine if post is a reply or not 
* format the date similar to how we do in `consistent-evaluation-submission-item.js`
discussion_body_screenshot
image

[design spec](https://xd.adobe.com/view/1f5fb770-bb62-4dfd-8e74-b500a4eafb54-4a0c/screen/eade7c62-9479-435f-97d1-669c06848a82)


lang/en.js
```diff
@@ -17,6 +17,7 @@ export default {
 	"closeBtn": "Close",
 	"comments": "Comments",
 	"continueGrading": "Continue Grading",
+	"downVotes": "Down Votes: ",
```
amauryperez, line 20: I don't think we include the `space` at the end of lang terms

amauryperez (approved):

lang/en.js
mbosiljevac: Thats a good point, I'll update it to use replacement strings since its not too much work!

components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
```diff
@@ -0,0 +1,129 @@
...
+
+	_renderDate() {
+		return html `<div class="d2l-body-small">Jan 28, 2020 1:24PM</div>`;
+	}
+
+	_renderBody() {
+		return html `<div class="d2l-body-compact d2l-consistent-evaluation-discussion-evidence-body-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis non tincidunt nibh. Integer eu rhoncus enim. Morbi vitae rhoncus est, vel ornare nulla. Quisque scelerisque, sapien vel pellentesque porttitor, tortor odio pulvinar nisi, a consectetur augue nulla a tellus. In purus eros, convallis a lacinia vitae, interdum molestie nunc. Donec in magna accumsan, rhoncus massa a, blandit dui. Phasellus accumsan scelerisque ipsum, quis pulvinar ante rhoncus pharetra. Fusce nec magna nisi. In ut lacus pharetra nibh mattis aliquam. Aenean viverra consequat nibh et malesuada.</div>`;
+	}
+
+	_renderRating() {
```
JamieTran, line 112: > allow star ratings (will most likely pull this out into its own component)
100% agree with this, what are your thoughts on it also handling the up/down vote count as well?

mbosiljevac: yep! I was thinking about pulling all the rating stuff into its own component (so upvotes and star ratings)!

JamieTran:
Overall I think this is great work and sets a good foundation for us to build on 😄

JamieTran (approved):

PR #368, opened 2021-03-17
use getComposedActiveElement

jeroach:
[DE42734](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fdefect%2F505877924612&fdp=true?fdp=true): Assignments > 20.21.03 > Dialog box A11y issues

As per suggestion by Dave B here: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/109#discussion_r504909814

Please note this PR only fixed this issue from the defect: `when the dialog closes, the focus is at the top of the page not on the icon that was clicked, this makes it hard for a visually impaired user to know where they are and it take a lot of keystroke to get back to where they were.`

The point around dialogs closing with `esc` is to be done as part of a follow up story, as the logic is the same as in the old experience.


mbosiljevac (approved):
lgtm!

JamieTran (approved):

PR #372, opened 2021-03-19
Us125727/hookup post info

mbosiljevac:
image


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -0,0 +1,133 @@
...
+	async _formatDiscussionPostObject(discussionPostEntity) {
+		const postTitle = discussionPostEntity.properties.subject;
+		const postBody = discussionPostEntity.properties.message;
+		const ratingInformation = { upVotes: 0, downVotes: 0 };
+
+		let createdDate = undefined;
+		if (discussionPostEntity.getSubEntityByClass(Classes.assignments.date)) {
+			createdDate = discussionPostEntity.getSubEntityByClass(Classes.assignments.date).properties.date;
+		} else {
+			console.warn('Consistent Evaluation discussion post date not found');
```
jeroach, line 73: I'm surprised this didn't trip the linter - I think we have a `no console` flag in there.

Regardless, either we should:
- consider it important enough to log via telemetry, and do so OR
- not log it

mbosiljevac: I'm okay with not logging it, it seems like an edge case that wouldn't be super common. 

The only reason I left it in is I took most of this logic from the `submissions-page` which logs an error when the submission date wasn't found.

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -0,0 +1,133 @@
...
+			threadTitle
+		};
+	}
+	async _getDiscussionPostEntities() {
+		this._discussionPostObjects = [];
+		if (this._discussionPostList !== undefined) {
+			for (const discussionPostLink of this._discussionPostList) {
+				if (discussionPostLink.href) {
+					const discussionPost = await this._getDiscussionPostEntity(discussionPostLink.href);
+					const discussionPostEntity = discussionPost.entity;
```
jeroach, line 100: Do we need a null check here after the `fetch`?

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -0,0 +1,133 @@
...
+			console.warn('Consistent Evaluation discussion post date not found');
+		}
+
+		let threadTitle = undefined;
+		let isReply = false;
+		if (discussionPostEntity.hasLinkByRel(threadRel)) {
+			isReply = true;
+			const threadHref = discussionPostEntity.getLinkByRel(threadRel).href;
+			const threadEntity = await this._getDiscussionPostEntity(threadHref);
+			threadTitle = threadEntity.entity.properties.subject;
```
jeroach, line 82: Same thought here around if we need a null check post `fetch`

jeroach (commented):
Overall this is really good :)

Just a few questions

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
jeroach: Hmmm, you're right - maybe I'm misremembering around the `no console`

If it's there in submissions, might as well leave it here.

mbosiljevac: there is a `no-console` lint rule but I think it ignores `console.warn`!

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
mbosiljevac: it would probably be safer! I'll put one in!

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -86,6 +86,16 @@ export class ConsistentEvaluationHrefController {
 		}
 		return undefined;
 	}
+	async getDiscussionPostInfo() {
+		let root = await this._getRootEntity(false);
+		if (root && root.entity) {
+			root = root.entity;
+			if (root.getSubEntityByRel(topicPostListRel)) {
```
JamieTran, line 93: I think this would throw if the `SubEntity` doesn't exist, we usually use the pattern `root.hasSubEntityByRel()`  before `getSubEntityByRel()`

mbosiljevac: It looks like it returns `undefined` if the subentity doesn't exist (https://github.com/Brightspace/node-siren-parser/blob/master/README.md#entitygetxbyystringregexp-key), but using hasSubEntityByRel is still probably cleaner so I'll change it to that!

JamieTran (approved):

jeroach (approved):

PR #375, opened 2021-03-22
Add discussion support for navbar title and subtitle

JamieTran:
- Navbar now displays the forum name and topic name when grading discussions
- Refactor assignment title and subtitle to be generic

image


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -86,6 +86,33 @@ export class ConsistentEvaluationHrefController {
...
+						}
+					}
+				}
+				return {
+					topicName,
+					forumName
+				};
+			}
+		}
+		return undefined;
```
mbosiljevac, line 114: nit: should you return `{ topicName, forumName }` here instead so that when you call `discussionNavInfo.topicName` in consistent-evaluation.js you wont get an error. Or you can do a check there that `discussionNavInfo` isn't undefined

mbosiljevac (approved):
Just one comment, looks good!

components/consistent-evaluation-page.js
```diff
@@ -617,10 +620,10 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		if (this.userProgressOutcomeHref) {
 			return this.localize('overallAchievement');
 		}
-		return this.organizationName;
+		return this.navTitleInfo.subtitleName;
 	}
 	get _navBarTitleText() {
-		if (this.assignmentName) return this.assignmentName;
+		if (this.navTitleInfo.titleName) return this.navTitleInfo.titleName;
```
jeroach, line 626: For these, do we need to check if `this.navTitleInfo` is valid?

jeroach (approved):

PR #376, opened 2021-03-22
DE42856/hide grademark if not enabled tii.

mbosiljevac:


JamieTran (approved):

components/left-panel/assignments/consistent-evaluation-submission-item.js
```diff
@@ -561,6 +553,22 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 		`;
 	}
 
+	_renderTiiSimilarity(tii, id, name) {
+		if (tii.originalityReportHref || tii.reportStatus !== tiiReportCompleteStatus) {
```
jeroach, line 557: Just want to double check the `if` logic - I was thinking this would be:

`tii.originalityReportHref && tii.reportStatus == tiiReportCompleteStatus`

I must be misunderstanding something :)

mbosiljevac: I wasn't 100% sure what logic we wanted here. Currently I show the similarity report if has information we want or if its still loading, since the similarity component has loading states (showing "Pending..." if its still loading).
I felt it was a better experience to have the similarity component load, then hide when its done loading if its disabled than to have it be hidden and appear if its enabled.
But I also could be misunderstanding when the originalityReportHref, but I think I checked that it was only assigned after the report status was complete.

jeroach: Thanks for clarifying - that makes sense - I had forgot about the rendering logic when the report is loading.

This is fine :)

jeroach (approved):

PR #377, opened 2021-03-22
US125734/conditionally render empty state.

mbosiljevac:
I'm not sure if we want to make a demo page for this? I could add one for the discussions page if we want one!
Note: the visual diffs only changed because of what branch this is based on
image


JamieTran (approved):
> I'm not sure if we want to make a demo page for this? I could add one for the discussions page if we want one!

LGTM, I don't think we need a demo page for the empty state

components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
```diff
@@ -63,10 +63,13 @@ export class ConsistentEvaluationEvidenceDiscussion extends RtlMixin(LocalizeCon
 	}
 
 	render() {
+		if (this.discussionPostList && this.discussionPostList.length === 0) {
+			return html`${this._renderSortDropDownList()}
```
jeroach, line 67: Perhaps something to double check with Fiona - but do we need to render the sort drop down if there are no posts to sort?

mbosiljevac: sure! I can double check! I kept it rendered since that's what the design specs currently show :)

jeroach: Yeah we might want it always there for consistency, but worth asking

mbosiljevac: chatted with Fiona and she agrees to hide the sort by when there are no posts! I will hide it :)

jeroach (approved):

PR #398, opened 2021-03-29
US125736/add discussion post skeleton [1/2]

mbosiljevac:
TODO:
* render discussion post inside table once table component is available
* have skeleton be shown until api calls to get discussion posts are done
image

### DIFF
--- a/components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
+++ b/components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
@@ -1,11 +1,12 @@
 import './consistent-evaluation-discussion-evidence-body';
-import { html, LitElement } from 'lit-element';
+import { css, html, LitElement } from 'lit-element';
 import { Classes } from 'd2l-hypermedia-constants';
 import { LocalizeConsistentEvaluation } from '../../../localize-consistent-evaluation.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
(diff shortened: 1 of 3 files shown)


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -58,9 +108,20 @@ export class ConsistentEvaluationDiscussionPostPage extends RtlMixin(LocalizeCon
 
 	render() {
 		return html`
+			${this._renderDiscussionItemSkeleton()}
+			${this._renderDiscussionItemSkeleton()}
```
JamieTran, line 112: any reason we are calling `_renderDiscussionItemSkeleton` twice here?

mbosiljevac: I wanted to write a `renderDiscussionItemSkeleton` function since the final design has that being rendered in the rows of the table, and I'm calling it twice so it renders twice and it looks similar to the designs which has two rows of the skeleton in the table

JamieTran (approved):
Great work! It's not specified in the spec but I wonder if Fiona wants us to skeletonize the sort and load more buttons

mbosiljevac:
> Great work! It's not specified in the spec but I wonder if Fiona wants us to skeletonize the sort and load more buttons

Good point! I'll send her a message about it! I might not implement the skeleton for the load more button since the button might get removed.

jmurray-d2l (approved):

jeroach (approved):

PR #399, opened 2021-03-29
Fix missing feedback attachment buttons

JamieTran:
Ensure we assign `_activityType` before populating any related infos

- fixes the issue where `_activityType` is set to `undefined` in `consistent-evaluation-page.js` (causing feedback attachments to not render)

### DIFF
--- a/components/consistent-evaluation.js
+++ b/components/consistent-evaluation.js
@@ -152,8 +152,8 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 				async() => {
 					const controller = new ConsistentEvaluationHrefController(this.href, this.token);
 					this._childHrefs = await controller.getHrefs();
-					this._rubricInfos = await controller.getRubricInfos(false);
 					this._activityType = await controller.getActivityType();
+					this._rubricInfos = await controller.getRubricInfos(false);
 					this._enrolledUser = await controller.getEnrolledUser();
 					this._groupInfo = await controller.getGroupInfo();
 					this._anonymousInfo = await controller.getAnonymousInfo();


components/consistent-evaluation.js
```diff
@@ -152,8 +152,8 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 				async() => {
 					const controller = new ConsistentEvaluationHrefController(this.href, this.token);
 					this._childHrefs = await controller.getHrefs();
-					this._rubricInfos = await controller.getRubricInfos(false);
 					this._activityType = await controller.getActivityType();
```
mbosiljevac, line 155: We might want to add a comment to explain why the ordering of these calls is important.
The other option is we might want to instead make `consistent-evaluation-page` more robust, so if the activityType changes then we check if it is an `assignmentActivity` and if so fetch the attachmentsInfo (since right now the attachmentsInfo is only fetched after the token/evaluationEntity is updated) 
ie add to consistent-evaluation-page something like 

```jsx
async updated(changedProperties) {
		super.updated(changedProperties);
		if (changedProperties.has('activityType') && this.activityType === assignmentActivity) {
			this._attachmentsInfo = await this._controller.fetchAttachments(this.evaluationEntity);
		}
	}
```

thoughts?

jeroach: I think a comment would be useful.

I'm not sure if we need the extra logic - if the user was iterating from an assignment to a discussion, would that not be a different evaluationEntity? and thus would trigger the update

I guess what I'm saying is that I wouldn't expect the activityType of a given evaluationEntity to change - unless I'm completely mis-remembering how the API works :P

mbosiljevac: Fair! I agree I'm pretty sure thats how the API works. 

I'd mainly want to add this logic to prevent us from breaking this in the future. Since attachments are only fetched once after the evaluationHref and token are assigned in the consistent-evaluation page I think its possible for us to get into the state again where the evaluationHref and token are set but activity type is unassigned, so we would never fetch any attachments

But if you think this is unlikely to happen I'm fine with just a comment explaining why the ordering of these calls is important

jeroach: @amauryperez @jmurray-d2l Hey guys :) can either of you comment from an API perspective here? Thanks!

amauryperez: Hey josh, if what you need is confirmation that iterating always gives you a new evaluationEntity, that is correct.

amauryperez: On the front-end implementation, I think either the comment and the check are fine. My preference would be to not have any of the attachmentsInfo logic at a higher level than necessary (so not on consistent-evaluation-page, but instead somewhere where only assignment stuff are processed), but I understand that is a bigger re-factoring

JamieTran: > My preference would be to not have any of the attachmentsInfo logic at a higher level than necessary

I agree with this but like you said, `consistent-eval-page` would need a refactor. I think that adding in a comment to provide context would help and we can monitor if the `activity-type` is `unassigned` - off the top of my head I can't think of a scenario where this would happen again with the current ordering of calls.

jeroach (approved):

PR #400, opened 2021-03-29
US125732/use topic entities under evaluation.

mbosiljevac:
Use the topic entities in the evaluation entity so we get the score related actions


components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
```diff
@@ -137,7 +137,7 @@ export class ConsistentEvaluationDiscussionEvidenceBody extends RtlMixin(Localiz
 		if (this.isReply) {
 			return html `<div class="d2l-consistent-evaluation-reply-to-container">
 				<d2l-icon class="d2l-consistent-evaluation-evidence-body-reply-icon" icon="d2l-tier1:reply"></d2l-icon>
-				<span class="d2l-body-compact d2l-consistent-evaluation-evidence-body-reply-to-text d2l-truncate">
+				<span class="d2l-body-small d2l-consistent-evaluation-evidence-body-reply-to-text d2l-truncate">
```
mbosiljevac, line 140: I realized I used the wrong styling so I'm just cleaning it up :p

jeroach (approved):
This is the refactor to support the API changes Amaury mentioned last week right?

LGTM

mbosiljevac:
> This is the refactor to support the API changes Amaury mentioned last week right?
> 
> LGTM

yep! so we can get the save post grade action!

PR #403, opened 2021-03-31
US125732/add save score action.

mbosiljevac:
Still need to double check that syncing works with rubrics and fix it if not
Also will make it look nicer when we have a table to put it in :p
image

### DIFF
--- a/components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
+++ b/components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
@@ -125,7 +125,7 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
 			}
 		}));
 	}
-	async _formatDiscussionPostObject(discussionPostEntity) {
+	async _formatDiscussionPostObject(discussionPostEntity, discussionPostEvaluationEntity) {
 		const postTitle = discussionPostEntity.properties.subject;
 		const postBody = discussionPostEntity.properties.message;
 		const ratingInformation = { upVotes: 0, downVotes: 0 };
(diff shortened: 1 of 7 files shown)


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -154,17 +154,18 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
...
 		};
 	}
 	async _getDiscussionPostEntities() {
 		this._discussionPostObjects = [];
 		if (this._discussionPostList !== undefined) {
-			for (const discussionPostEntity of this._discussionPostList) {
-				if (discussionPostEntity.links && discussionPostEntity.links[0].href) {
-					const discussionPost = await this._getDiscussionPostEntity(discussionPostEntity.links[0].href);
+			for (const discussionPostEvaluationEntity of this._discussionPostList) {
+				if (discussionPostEvaluationEntity.links && discussionPostEvaluationEntity.links[0].href) {
```
jeroach, line 165: Do we need to check if `links` has at least one element in it?

components/left-panel/discussions/consistent-evaluation-discussion-post-score.js
```diff
@@ -0,0 +1,78 @@
...
+	flush() {
+		if (this._debounceJobs.grade && this._debounceJobs.grade.isActive()) {
+			this._debounceJobs.grade.flush();
+		}
+	}
+	onGradeChanged(e) {
+		const score = e.detail.value;
+		this._debounceJobs.grade = Debouncer.debounce(
+			this._debounceJobs.grade,
+			timeOut.after(800),
```
jeroach, line 55: How did we choose this length of timeout?

mbosiljevac: I took this length from `consistent-evaluation-grade-result.js` where we do a similar wait for an onGradeChange

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
mbosiljevac: I can add a check for safety, but I think the api works in such a way that we should be guaranteed a link since the discussionPostEvaluationEntity should always have a link to a discussionPostEntity

jeroach: If that's the case then we should be fine as is

jeroach (approved):

PR #408, opened 2021-04-06
Fix coa mastery view

JamieTran:
When refactoring CE to conditionally render assignments vs discussions we missed Course Overall Achievement (COA) so it wasn't rendering. This is a PR with a quick fix as we are blocking team mango's testing.

In the future, I think we would want to extract COA out and render it like how we do with discussions (right now it's coupled with assignments). It would be a good dragonslayer task.

Before:
image
After:
image


jeroach (approved):
Looks good. I agree that we should refactor it out in future.

Perhaps make a story for that so we don't lose it.

mbosiljevac (approved):

PR #417, opened 2021-04-13
US125730/sort discussion posts.

mbosiljevac:
image


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -221,6 +233,28 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
 				<div class="d2l-skeletize d2l-consistent-evaluation-discussion-evidence-body-rating-skeleton"></div>
 			</div>`;
 	}
+	_sortDiscussionPosts() {
```
amauryperez, line 236: this looks good. it feels like adding this type of logic is when we should have unit tests?

mbosiljevac: good point! I'll look into it!

JamieTran:
Quick question, is the sorting-method saved when iterating?

mbosiljevac:
> Quick question, is the sorting-method maintained when iterating?

good question! the value of sorting-method looks to stay the same but the discussionPosts aren't properly sorted, I'll fix that!

amauryperez:
> Quick question, is the sorting-method maintained when iterating?

that is a good question. we should check with Fiona to see if this is something she would like or if she prefers to reset to the default

mbosiljevac:
> > Quick question, is the sorting-method maintained when iterating?
> 
> that is a good question. we should check with Fiona to see if this is something she would like or if she prefers to reset to the default

I'll double check with her! I remember discussing this behaviour in one of the grooming sessions but I don't fully remember what we decided on/if we decided on anything

mbosiljevac:
> > > Quick question, is the sorting-method maintained when iterating?
> > 
> > 
> > that is a good question. we should check with Fiona to see if this is something she would like or if she prefers to reset to the default
> 
> I'll double check with her! I remember discussing this behaviour in one of the grooming sessions but I don't fully remember what we decided on/if we decided on anything

Update: Confirmed with Fiona, we do want to keep filtering and sorting settings when iterating between students!

JamieTran (approved):

amauryperez (approved):

PR #421, opened 2021-04-14
Add status indicator for unscored posts

JamieTran:
Use d2l-status-indicator to display the number of unscored posts when grading posts individually
This is not the original placement as we are still waiting for the d2l-table component

- the indicator is updated onChange of post-score and when saving/publishing

image
image


components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
```diff
@@ -91,14 +102,30 @@ export class ConsistentEvaluationEvidenceDiscussion extends SkeletonMixin(RtlMix
 			}
 		}));
 	}
+	_getUnscoredPostsCount() {
+		let unscoredPosts = 0;
+		if (this.discussionPostList) {
+			this.discussionPostList.forEach(postEntity => {
+				if (postEntity.properties && postEntity.properties.score === null) {
+					unscoredPosts++;
+				} else if (!('properties' in postEntity)) {
```
JamieTran, line 111: I'm not sure if there is a better way of determining if posts are individually graded or not

mbosiljevac: I think just checking if `postEntity.properties.outOf` exists is fine (since I think you need an outOf to individually assess posts), but also I think what you're doing is fine!

mbosiljevac (approved):
LGTM

mbosiljevac:
One another note, we might want the text to not be all upper case, I think if we add the css attribute of `text-transform: none;` to the d2l-status-indicator the text won't be in all caps

jmurray-d2l (approved):

PR #432, opened 2021-04-19
US126243/add filters.

mbosiljevac:
Chatted with Fiona about how filters work. We have two categories: Post (options threads and replies) and Score (options: scored and unscored) (see image). 
Within a category its a multiselect with _or_ logic (if both thread and reply are selected any post that is a thread OR a reply gets returned)
Between categories it uses _and_ logic (if threads and scored are selected a post needs to be a thread AND scored to be returned)

Theres logic to hide scored/unscored filters if posts aren't individually scored. Filters persist when iterating between students and can be selected before the page is done loading

Note: also removed the load more button and removed assumption that the backend returns posts sorted by oldest first (I ran into something where that wasn't the case) 

image
image


lang/en.js
```diff
@@ -48,7 +48,6 @@ export default {
 	"kB": " KB",
 	"late": "late",
 	"leaveBtn": "Leave",
-	"loadMore": "Load More",
```
mbosiljevac, line 51: I'm not sure if I need to remove this langterm in all the other languages too

components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
```diff
@@ -183,6 +243,29 @@ export class ConsistentEvaluationEvidenceDiscussion extends SkeletonMixin(RtlMix
 
 	}
 
+	_setPostsFilter(e) {
+		const newFilters = [...this._selectedPostFilters];
```
mbosiljevac, line 247: We need to do this (create a new array and assign it) since lit-element doesn't treat adding an element to an array attribute a change to the property and won't re-render the component unless we assign the attribute to a new array :(

lang/en.js
amauryperez: yeah, if we have it in other places we should probably clean it up

components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
amauryperez: interesting

mbosiljevac:
> I'm concerned about the fact that we want to remember the filters when navigating because at some point we will be navigating between different tools, so if we go from discussion topic, to assignments, and then to another discussion topic are we expected to remember the filters? should probably run this by Fiona to have an idea what are the expectations

I chatted with Fiona about this! The current expectation is that we keep filters when iterating between discussion, but if we iterate to another activity type, we wouldn't need to keep the filters (even if we iterate from a discussion, to an assignment, to a discussion) because of the context switch

amauryperez (approved):

mbosiljevac:
It looks like Fiona also wants us to make sure that between different discussion topics we remove filters. Personally I think we should worry about that logic when we allow users to iterate between different topics and we better understand how we implement that

amauryperez:
it would be great if we already had that logic in place. it could be a separate PR

components/helpers/discussionPostsHelper.js
```diff
@@ -23,3 +23,27 @@ export function sortDiscussionPosts(discussionPostObjects, sortingMethod) {
...
+		let satisfiesPostFilters = true;
+		let satisfiesScoreFilters = true;
+		if (selectedPostFilters.length > 0) {
+			satisfiesPostFilters = discussionPost.isReply ? selectedPostFilters.includes(filterByReplies) : selectedPostFilters.includes(filterByThreads);
+		}
+		if (selectedScoreFilters.length > 0) {
+			const score = discussionPost.discussionPostEvaluationEntity.properties.score;
+			satisfiesScoreFilters = score === null ? selectedScoreFilters.includes(filterByUnscored) : selectedScoreFilters.includes(filterByScored);
+		}
+		return satisfiesPostFilters && satisfiesScoreFilters;
```
JamieTran, line 45: Curious as to what happens here if one of these is false, does it not apply the filter?

components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
```diff
@@ -132,19 +142,27 @@ export class ConsistentEvaluationEvidenceDiscussion extends SkeletonMixin(RtlMix
...
 	}
-	_renderLoadMoreButton() {
+	_renderFilterDropDownList() {
 		return html`
-			<d2l-button
-				aria-hidden="${this.skeleton}"
-				class="d2l-consistent-evaluation-evidence-discussion-load-more"
-				secondary
-			>${this.localize('loadMore')}</d2l-button>`;
+			<d2l-labs-filter-dropdown
```
JamieTran, line 155: is this keyboard accessible?

components/helpers/discussionPostsHelper.js
mbosiljevac: It would remove the element! The js filtering function iterates through all the elements in the array and applies the given function to the element. If the function returns false, then that element is removed from the resulting array, if it returns true the element is kept in the resulting array. Since if either `satisfiesPostFilters` or `satisfiesScoreFilters` are false the function returns false, the element would get removed

components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
JamieTran: I've run into this issue a couple of times :(

components/left-panel/discussions/consistent-evaluation-evidence-discussion.js
mbosiljevac: good question! It is, the only weird behaviour is that the tab order goes to the sort button (on the right) before the filter button (on the left). This looks like it might be a bit difficult to fix since the focus is on an element within the `d2l-labs-filter-dropdown` but its still accessible

JamieTran (approved):

PR #438, opened 2021-04-21
US127295/remove console error when grading rubrics.

mbosiljevac:
Note still need to do some testing on discussions popout rubrics but still wanted to get some eyes on this

### DIFF
--- a/components/controllers/ConsistentEvaluationHrefController.js
+++ b/components/controllers/ConsistentEvaluationHrefController.js
@@ -215,6 +215,7 @@ export class ConsistentEvaluationHrefController {
 			const evaluationEntity = await this._getEntityFromHref(evaluationHref, false);
 			return evaluationEntity;
 		}
+		return undefined;
 	}
 	async getGradeItemInfo() {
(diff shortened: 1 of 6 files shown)


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -207,6 +207,15 @@ export class ConsistentEvaluationHrefController {
 			return undefined;
 		}
 	}
+	async getEvaluationEntity() {
```
JamieTran, line 210: should `getEvaluationEntity()` return `undefined` if it doesnt find `root.entity`?

JamieTran (approved):
other than that 1 comment this LGTM!

components/controllers/ConsistentEvaluationHrefController.js
mbosiljevac: since `getEvaluationEntity()` doesn't return anything if the `root.entity` doesn't exist, it defaults the return value to undefined, but I can return `undefined` explicitly to make it more clear!

PR #439, opened 2021-04-21
US127329/update attachment icons

mbosiljevac:
image


JamieTran (approved):
This PR looks great but I just noticed there are file-attachment functions shared between assignments and discussions, it might be worth it to extract them into their own helper file to reduce code duplications. What are your thoughts on this? 
eg. `_formatDateTime`, `_getReadableFileSizeString`, and now `_getLinkIconTypeFromUrl`
(Doesnt have to happen in this PR)

mbosiljevac:
> This PR looks great but I just noticed there are file-attachment functions shared between assignments and discussions, it might be worth it to extract them into their own helper file to reduce code duplications. What are your thoughts on this?
> eg. `_formatDateTime`, `_getReadableFileSizeString`, and now `_getLinkIconTypeFromUrl`
> (Doesnt have to happen in this PR)

I think that's a great idea and makes a lot of sense! I can do this in this PR since its pretty small! We can also add tests for these functions if we pull it out into a helper function

mbosiljevac:
> > This PR looks great but I just noticed there are file-attachment functions shared between assignments and discussions, it might be worth it to extract them into their own helper file to reduce code duplications. What are your thoughts on this?
> > eg. `_formatDateTime`, `_getReadableFileSizeString`, and now `_getLinkIconTypeFromUrl`
> > (Doesnt have to happen in this PR)
> 
> I think that's a great idea and makes a lot of sense! I can do this in this PR since its pretty small! We can also add tests for these functions if we pull it out into a helper function

FYI moved `_formatDateTime`, and now `_getLinkIconTypeFromUrl` to `submissionsAndFilesHelper.js`. I couldn't move  `_getReadableFileSizeString` to that file since it uses the localize function and that makes moving it to a helper file a lot more complicated

amauryperez (approved):

JamieTran (approved):

PR #446, opened 2021-04-26
Add discussion ratings 

JamieTran:
- Add rating component
- - works with all 3 types of ratings: star rating, upvote & downvote, and upvote only
- Replace hard-coded ratings with ratings component from evidence body

Will have a separate PR up for vdiff and demo pages to keep this more readable


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -237,6 +242,34 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
 			discussionPostEvaluationEntity
 		};
 	}
+	_formatDiscussionRatings(discussionPostEntity) {
+		let ratingsInfo = {};
+		if (discussionPostEntity.class.includes(fivestarRatingClass)) {
```
amauryperez, line 247: nit: this here should fit the use of the `switch` statement which is easier to read

components/left-panel/discussions/consistent-evaluation-discussion-post-rating.js
```diff
@@ -0,0 +1,80 @@
...
+				return html``;
+		}
+	}
+
+	_renderFiveStarRating() {
+		return html`
+			<div class="d2l-body-compact d2l-consistent-evaluation-discussion-evidence-body-rating-text center">
+				${(this.ratingInformation.numRatings <= 0) ? '-' : this.ratingInformation.ratingAverage} / 5
+				<d2l-icon icon="tier1:subscribe-filled"></d2l-icon> <br>
+				(${this.localize('numRatings', 'num', (this.ratingInformation.numRatings > 99) ? '99+' : this.ratingInformation.numRatings)})
```
amauryperez, line 56: nit: maybe this logic can go in a function so its not duplicated 3 times?

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
JamieTran: I agree a switch statement would be best but the `array.includes()` doesn't play well with a switch statement unless a `switch(true)` is used. (which kind of defeats the purpose of it). I've separated it into two functions to help with this.

amauryperez (approved):

jeroach (approved):

PR #450, opened 2021-04-27
Display discussion word count config turned on

JamieTran:
Added word count to discussion posts when it is supplied from the API.

- 9999 word limit was suggested by Fiona

image


components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
```diff
@@ -226,12 +238,24 @@ export class ConsistentEvaluationDiscussionEvidenceBody extends RtlMixin(Localiz
...
+		>${this.postTitle}</div>`;
+	}
+	_renderWordCount() {
+		if (this.wordCount === undefined) {
+			return html``;
+		}
+		return html`
+			<d2l-icon class="d2l-separator-icon" aria-hidden="true" icon="tier1:dot"></d2l-icon>
+			<span class="d2l-body-small">
+				<a class="d2l-word-count-bold">${(this.wordCount > 9999) ? '9999+' : this.wordCount} </a>
```
jeroach, line 255: I seem to remember that some languages put the plus sign elsewhere. Might be worth making this a lang term just to be safe.

lang/en.js
```diff
@@ -154,5 +154,6 @@ export default {
 	"userIteratorText": "User {num} of {total}",
 	"userSubmissions": "User Submissions",
 	"viewFullSubmission": "View Full Submission",
-	"warning": "Warning"
+	"warning": "Warning",
+	"wordCount": "words"
```
jeroach, line 158: I think this term should incorporate the number - e.g. `{num} words`

jeroach (changes requested):
A few minor things to support other languages

lang/en.js
JamieTran: I thought about doing this as well but I couldn't a simple way to style half of a langterm (the design has the number being bolded and the rest normal) any thoughts?

jeroach: I'd talk to Fiona about if we can change the style. I think correctness should trump style.

JamieTran: We decided to go with unbolding the number

karma.sauce.conf.js
```diff
@@ -6,12 +6,12 @@ const customLaunchers = {
 	chrome: {
 		base: 'SauceLabs',
 		browserName: 'chrome',
-		platform: 'OS X 10.15',
+		platform: 'macOS 11.00',
```
JamieTran, line 9: This update is in reference to this [thread](https://d2l.slack.com/archives/C0PHG3QB0/p1619120383206900) where we ran into the problem with safari, but I've updated it so we use the most updated platform for all browser tests

jeroach, line 9: Are these changes intended?

jeroach (approved):
LGTM - just one question around Sauce versions of OSX

karma.sauce.conf.js
jeroach: And you answered my question already ;)

PR #452, opened 2021-04-28
- Using the correct href for the outcomes header

jmurray-d2l:


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -308,7 +308,7 @@ export class ConsistentEvaluationHrefController {
 			previousHref = this._getHref(root, previousRel);
 			actorHref = this._getHref(root, Rels.Activities.actorActivityUsage);
 			userHref = this._getHref(root, Rels.user);
-			alignmentsHref = this._getHref(root, Rels.assessment);
+			alignmentsHref = this._getHref(root, alignmentsRel);
```
JamieTran, line 311: Oops, I must've broken this when cleaning up our constants. This rel actually [exists](https://github.com/Brightspace/d2l-hypermedia-constants/blob/master/index.js#L189) in `d2l-hypermedia-constants` so we can use `Rels.Alignments.alignments` instead of adding back the constant.

JamieTran (changes requested):

components/controllers/ConsistentEvaluationHrefController.js
jmurray-d2l: Thanks! Couldn't find it! Obviously didn't look hard enough lol

JamieTran (approved):
10/10 commit message

PR #461, opened 2021-05-06
Check for comments before opening submission file

JamieTran:


components/consistent-evaluation.js
```diff
@@ -225,7 +225,8 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 
 			if (numberOfSubmittedFiles === 1) {
 				const isTiiEnabled = attachmentList.entities[0].hasSubEntityByRel(Rels.Assignments.turnItIn);
-				if (!isTiiEnabled) {
+				const hasComments = attachmentList.entities[0].properties.comment || false;
```
jeroach, line 228: Do we need null checks on `.properties.comment` ?

JamieTran: Good call, this could throw if properties is undefined but I can't think of a case when that would be, better safe than sorry.

jeroach (approved):

PR #471, opened 2021-05-12
Remove File Size Rendering For Links

Jevan5:
### Previously
The file size was always rendered for an attachment on the discussion post assessment page, regardless of the attachment type.

### Now
The file size is ignored if the attachment is a link.

### Other
Small amount of refactoring to centralize the attachment logic.


MarquisLP (approved):

components/left-panel/assignments/consistent-evaluation-submission-item.js
```diff
@@ -383,8 +373,8 @@ export class ConsistentEvaluationSubmissionItem extends RtlMixin(LocalizeConsist
 						<div slot="supporting-info">
 							${this._renderFlaggedStatus(flagged)}
 							${extension.toUpperCase()}
-							${this._isLinkAttachment(attachment) ? html`` : html`<d2l-icon class="d2l-separator-icon" aria-hidden="true" icon="tier1:dot"></d2l-icon>`}
-							${this._isLinkAttachment(attachment) ? html`` : this._getReadableFileSizeString(size)}
+							${isLinkAttachment ? html`` : html`<d2l-icon class="d2l-separator-icon" aria-hidden="true" icon="tier1:dot"></d2l-icon>`}
+							${isLinkAttachment ? html`` : getReadableFileSizeString(size, this.localize.bind(this))}
```
JamieTran, line 377: This is a neat way of getting localization into the helper method

JamieTran (approved):

components/left-panel/assignments/consistent-evaluation-submission-item.js
Jevan5: This is the part of the PR that I felt would come under the most fire, but I am glad it's okay :) Feels kind of hacky, but I don't think there are any simpler ways around it

PR #489, opened 2021-05-27
Jroach/fix rubric demo

jeroach:
Please note, this merges into my branch for US127885

Doing the rubric fix bit in a separate PR for clarity.

FYI - all of the new static data files (the bulk of this PR) were copied from the `d2l-rubric` repo


jeroach:
Sauce had a weird CI issue - unit tests pass locally so hopefully it's a one off failure

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -210,19 +210,17 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 			if (!rubric) {
 				return html``;
 			}
+
 			return html`
 				<div class="d2l-consistent-evaluation-rubric ${this.isPopout ? 'd2l-consistent-evaluation-popout' : ''}">
 					${this._renderRubricTitle(rubric.rubricTitle)}
 					${this._renderGradedBy(rubric.assessorDisplayName)}
 					<d2l-rubric
-						data-rubric-id=${rubric.rubricId}
```
jmurray-d2l, line 218: We needed this to determine which rubric was modified if there are multiple rubrics on the page, does that still work?

jeroach: the property no longer exists in `d2l-rubric`: https://github.com/Brightspace/d2l-rubric/blob/master/d2l-rubric.js

Same goes for the others I removed

jmurray-d2l: I don't think its a property of d2l-rubric, I think its just a property that we add to the element.

JamieTran: I think @jmurray-d2l is right, since its a data-attribute, we are adding the rubric-id ourselves to the component so that we can track which rubric is opened at any given time

jeroach: My bad, I've added them back

jmurray-d2l (approved):

JamieTran (approved):

PR #495, opened 2021-06-01
Add tabindex and keyboard accessibility to mobile drawer

JamieTran:
Allows users to tab into the mobile drawer of CE and press `Enter` or `Space` to open the dialog

### DIFF
--- a/components/right-panel/consistent-evaluation-right-panel-block.js
+++ b/components/right-panel/consistent-evaluation-right-panel-block.js
@@ -111,10 +111,18 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 		}
 	}
 
+	_onKeydownHandler(e) {
+		if (e.key === 'Enter' || e.key === ' ') {
+			this._toggleOpenDialog();
+		}
+	}
+
 	_renderListItems() {
 		return html`
 			<d2l-list-item class="d2l-list-item"
-				@click=${this._toggleOpenDialog}>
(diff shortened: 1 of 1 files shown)


components/right-panel/consistent-evaluation-right-panel-block.js
```diff
@@ -111,10 +111,18 @@ class ConsistentEvaluationRightPanelBlock extends LitElement {
 		}
 	}
 
+	_onKeydownHandler(e) {
+		if (e.key === 'Enter' || e.key === ' ') {
```
jeroach, line 115: Do we need an `isMobile` check here, similar to https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/f5cdfbef55e1c8071ce5c2a4a3a386db549d55f3/components/right-panel/consistent-evaluation-right-panel-block.js#L85

As currently this will trigger even on full screen view for all right panel blocks.

JamieTran: No, this `function` is only called when there is a keydown on the `d2l-list-item` https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/f5cdfbef55e1c8071ce5c2a4a3a386db549d55f3/components/right-panel/consistent-evaluation-right-panel-block.js#L122
Which is only rendered when it is in mobile-view (link you posted above)

jeroach (approved):

PR #499, opened 2021-06-03
fix grade validation event issue

cloudislife8899:


components/consistent-evaluation-page.js
```diff
@@ -641,6 +644,15 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		}
 		return this.evaluationEntity.properties.state === publishedState;
 	}
+
+	_isInvalidGradeValue() {
```
jmurray-d2l, line 648: nit: Consider using the reverse of this `IsValidGrade`.  Below you have `if(!this.IsInvalidGrade())` Which is a double negative and adds a bit more confusion

components/consistent-evaluation-page.js
```diff
@@ -487,7 +490,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		this.currentFileId = newFileId;
 	}
 	async _checkAndUpdateEvaluationEntityAndDisplayToast(newEvaluationEntity, errorTerm, successTerm) {
-		if (!newEvaluationEntity) {
+		if (!newEvaluationEntity || newEvaluationEntity.score === null) {
```
jmurray-d2l, line 493: Have you tested this with other actions?  What happens if you only update the feedback?

components/consistent-evaluation-page.js
```diff
@@ -728,14 +740,16 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 
 		await this._mutex.dispatch(
 			async() => {
-				const entity = await this._controller.fetchEvaluationEntity(false);
-				const newEvaluationEntity = await this._controller.publish(entity);
-				this._currentlySaving = false;
+				if (!(this._isInvalidGradeValue())) {
```
jmurray-d2l, line 743: I think this is okay for now, but if we need additional validation, we'll end up having to grow the conditions in this if statement.   Consider making a general validation helper that we could add to instead of having to add to this if statement.  I can help you here if you'd like

components/consistent-evaluation-page.js
JamieTran: Do we need this `null` check? Didn't you determine that the response returns the previous `score` and not `null`

package.json
```diff
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@brightspace-ui-labs/facet-filter-sort": "^4.1.0",
-    "@brightspace-ui-labs/grade-result": "^1.2.3",
+    "@brightspace-ui-labs/grade-result": "^1.2.4",
```
JamieTran, line 42: you'll have to update `package-lock.json` to keep CI from yelling at you

components/right-panel/consistent-evaluation-grade-result.js
```diff
@@ -144,23 +144,22 @@ export class ConsistentEvaluationGradeResult extends LocalizeConsistentEvaluatio
 	onGradeChanged(e) {
 		const score = e.detail.value;
 
-		if (this.grade.isNumberGrade && (score < 0 || score > 9999999999)) {
-			return;
-		}
+		const isInvalidGrade = this.grade.isNumberGrade && (score < 0 || score > 9999999999);
```
amauryperez, line 147: I think we should stick with `isValid` here, I just remembered that leaving the textbox empty is also valid. So `const isValidGrade = !this.grade.isNumberGrade || (this.grade.isNumberGrade && (score >= 0 || score <= 9999999999));`

cloudislife8899: Leaving it empty works. I changed it to isInvalid to prevent some double negatives

jmurray-d2l (approved):

PR #506, opened 2021-06-07
US128858 / US128857 - Fixing scored discussion post filtering issue

maddiegabriel:
Added a conditional around the filtering function call so it only executes when the user changes the filtering method (instead of on each render).

This fixes [US128857](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F601065927259), which makes [US128858 ](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fuserstory%2F601066988365) obsolete.


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -188,7 +190,13 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
 			sortDiscussionPosts(this._discussionPostObjects, this._currentSortingMethod);
 		}
 
-		this._displayedDiscussionPostObjects = filterDiscussionPosts(this._discussionPostObjects, this.selectedPostFilters, this.selectedScoreFilters);
+		this._displayedDiscussionPostObjects = this._discussionPostObjects;
```
JamieTran, line 193: I don't think this will work, if I'm understanding this right `this._discussionPostObjects` is always the full list of discussion posts so when a re-render happens, it will display all discussions again. 
E.g. 
1. filter by thread
2. change score (to force a re-render)
3. all posts will show again (since it doesn't make it into the conditional)

maddiegabriel: Good catch!! Added a condition to avoid that happening, unless no filters are selected

JamieTran (approved):
LGTM

jmurray-d2l (approved):
Nice work!

PR #518, opened 2021-06-09
try fix safari tests

jeroach:


vix633 (approved):
Safari tests goooo. Thanks @svanherk

PR #527, opened 2021-06-11
Use the new profile card

vix633:
This is sort of an MVP. We need to add the user progress link and maybe some other stuff.


components/header/d2l-consistent-evaluation-lcb-user-context.js
```diff
@@ -266,65 +245,18 @@ export class ConsistentEvaluationLcbUserContext extends EntityMixinLit(RtlMixin(
...
-		let displayName = undefined;
-		if (this.enrolledUser) {
-			emailHref = this.enrolledUser.emailPath;
-			instantMessageHref = this.enrolledUser.pagerPath;
-			userProgressHref = this.enrolledUser.userProgressPath;
-			userProfileHref = this.enrolledUser.userProfilePath;
-			displayName = this.enrolledUser.displayName;
-		}
-
-		if (this.anonymousInfo && this.anonymousInfo.isAnonymous && !this.anonymousInfo.assignmentHasPublishedSubmission) {
```
JamieTran, line 283: This is the only usage of `this.anonymousInfo`, is this now handled by the profile-card? If so, we could probably remove it from the properties.

Scy-D2L: Good catch - the profile card doesn't touch this at all. We'll want this check to be down in `_renderProfileImage`

vix633:
@svanherk We'd love love input on the vdiff tests we added

package.json
```diff
@@ -49,7 +49,7 @@
     "d2l-hypermedia-constants": "^6.40.0",
     "d2l-outcomes-level-of-achievements": "github:Brightspace/outcomes-level-of-achievement-ui#semver:^3",
     "d2l-outcomes-overall-achievement": "github:Brightspace/d2l-outcomes-overall-achievement#semver:^1",
-    "d2l-rubric": "github:Brightspace/d2l-rubric#semver:^3",
+    "d2l-rubric": "github:Brightspace/d2l-rubric#25d2fa24d0c13703126f9f0dafc9ce7f09f4e80a",
```
svanherk, line 52: I expect this will fail once it gets to BSI, since everything in BSI needs to be using the same version

test/perceptual/consistent-evaluation-lcb.visual-diff.js
```diff
@@ -29,6 +28,31 @@ describe('d2l-consistent-evaluation', () => {
 		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 	});
 
+	it('renders learner profile card', async function() {
+		await page.$eval('#open-card', async(elem) => {
+			const lcb = elem.querySelector('d2l-consistent-evaluation-learner-context-bar');
+			const userContext = lcb.shadowRoot.querySelector('d2l-consistent-evaluation-lcb-user-context');
+			const card = userContext.shadowRoot.querySelector('d2l-labs-user-profile-card');
+			card._userProfileCardSettings = {
```
svanherk, line 36: Scope of the visual-diff test looks good!  Isn't just testing the card (which should in theory be covered by the visual-diff tests in the profile card repo) but you're testing a smaller piece than "the whole page" which is going to update a lot.

Setting the card settings like this is definitely fragile - ideally you'd configure this through the HM data you're passing into the test, so that changes to the card's underlying data structure and code wouldn't break this test.  But I don't know how easy/tricky that is

Scy-D2L: The component magically requests the instance-based settings internally. We never actually pass that information in to it, it just knows based on where it lives, where it needs to go looking. So running it in a non-lms environment, it will always fail-closed and assume we want to show the absolute minimum (basically just the username and contact links). It's possible that testing for that minimum is fine, since any issues relating to them would get snapped up in the profile card tests? 🤔

test/perceptual/consistent-evaluation-lcb.visual-diff.js
```diff
@@ -29,6 +28,31 @@ describe('d2l-consistent-evaluation', () => {
...
+				showOnlineStatus: true,
+				showRole: true,
+				showBadgeTrophy: true,
+				showOrgDefinedId: true
+			};
+			const listener = new Promise((resolve) => {
+				card.addEventListener('d2l-labs-user-profile-card-opened', resolve, { once: true });
+			});
+			card.open();
+			return listener;
```
svanherk, line 50: We have a [nice little `oneEvent` helper](https://github.com/BrightspaceUI/visual-diff#asynchronous-behaviors) for this if you want to use it, but what you have may work better if your event is this nested and doesn't bubble.

package.json
vix633: Yeah, just wanted to make sure we're barking up the right tree with the rubrics failures here.

test/perceptual/consistent-evaluation-lcb.visual-diff.js
svanherk: Hmm, can we mock out the secret automagic route it's calling at least?  You still know too much about the component's internals, but at least you'd only need to update things if the data/api side changed, not the component code?
Mostly just calling this out as a use-at-your-own-risk strategy, the ROI might not be worth implementing my suggestion

Scy-D2L: After some poking it's simpler to test it without passing anything. The fallback hides a bunch of information but does explode, and the effects of that only matter on the actual profile card - so we don't need to bother with it here. :) Simplified things.

Scy-D2L (approved):

PR #543, opened 2021-06-16
fix table showing up in skeleton view

cloudislife8899:


jeroach (approved):
Good catch. Sorry for missing this, and thanks for fixing it!

JamieTran (approved):

PR #554, opened 2021-06-21
♿️ Add aria label for discussions link

Scy-D2L:
Discussion post titles will now read "`{postTitle} (opens in a new tab)`" for screen-reader users

### DIFF
--- a/components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
+++ b/components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
@@ -231,9 +231,11 @@ export class ConsistentEvaluationDiscussionEvidenceBody extends RtlMixin(Localiz
 		}
 	}
 	_renderTitle() {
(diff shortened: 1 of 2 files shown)


JamieTran (approved):

components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
```diff
@@ -231,9 +231,11 @@ export class ConsistentEvaluationDiscussionEvidenceBody extends RtlMixin(Localiz
 		}
 	}
 	_renderTitle() {
+		const labelText = `${this.postTitle} (${this.localize('opensInANewTab')})`;
```
Scy-D2L, line 234: Not sure what our best practice is for accessibility text appended onto titles. This could also be done as a langterm that takes the post title to format itself or without the `( )` - but the term is generic enough not to need to be tied to this link and the parenthesis separate the tip from the title. Open to alternatives if we've done something like this before.

PR #575, opened 2021-06-25
US129462 - The score status in the evidence table is not correct for discussions set up to be graded at the overall Grade level

amauryperez:


maddiegabriel (approved):
Good catch, LGTM!

PR #588, opened 2021-07-07
Fix de-sync between scored post and overall grade

JamieTran:
[US129155](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F601750512757)

- Fixes the issue where the post scores would not be synced up with the overall grade score (this is because we stopped re-fetching `discussionPostList` is updated)

PR #590, opened 2021-07-20
Fix: saving evaluation would populate discussion post scores with old grades

JamieTran:
#588 re-introduced a bug where saving evaluation would populate discussion post scores with old grades

`_refreshDiscussionPosts()` was periodically fetching the old `post-scores` due to a race condition with our transient-save. Fixed this by updating the discussion list `on-transient-save`  instead of re-fetching


jeroach (approved):
LGTM

maddiegabriel (approved):
Good stuff!

PR #594, opened 2021-07-21
Mwang/us129679

cloudislife8899:
Changing a key value in an object doesn't trigger lit element to re-render the component. Fixed by essentially passing a new object by de-restructuring the old one and updating the evaluation state.


jeroach (approved):
LGTM

PR #595, opened 2021-07-22
Adjust Discussions Table to Fit on Mobile (Again)

Scy-D2L:
The div containing the table was stretching off screen.

Not sure what happened but the changes from #552 are missing in the latest version (blame says they never happened despite the merged pr 🤷🏼) so reapplying the fix


JamieTran (approved):
lgtm!

jeroach (approved):
lgtm - just confirming since this is a slightly different diff than the PR before

Scy-D2L:
> lgtm - just confirming since this is a slightly different diff than the PR before

A few other changes were made to the same CSS class and a skeleton variant was added, so the diff looks different but the changes are applied to the same class. 😃

PR #596, opened 2021-07-23
DE44214 - Rubrics - Fix read-only scenario

amauryperez:


components/consistent-evaluation-page.js
```diff
@@ -422,7 +418,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 						.gradeItemInfo=${this.gradeItemInfo}
 						.token=${this.token}
 						?can-save-overall-grade=${this._gradeEntity && this._gradeEntity.hasActionByName(saveGradeActionName)}
-						?rubric-read-only=${this.rubricReadOnly}
+						?rubric-read-only=${!this._allowEvaluationWrite()}
```
amauryperez, line 421: using the same check we use to determine when the evaluation components can be edited

maddiegabriel (approved):
LGTM!

JamieTran (approved):

components/consistent-evaluation-page.js
JamieTran: it doesn't look like `_rubricReadOnly` ever changed from being `false` lol

PR #597, opened 2021-07-23
🐛 Fix post date not displaying correctly

Scy-D2L:
Fix for [US130207](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F605006069985&fdp=true?fdp=true): [Consistent Eval][Discussion] Post date and time is not being displayed


maddiegabriel (approved):
nice catch 👍

jeroach (approved):
lgtm

PR #600, opened 2021-07-29
🐛 Use instance timezone instead of OS timezone

Scy-D2L:
Fix for [DE44018](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F601903897593&fdp=true?fdp=true): Assignments >20.21.5 > Time zone gets changed on turning new content experience


jeroach (approved):
LGTM

https://github.com/BrightspaceUI/core/blob/main/helpers/dateTime.js#L148

PR #621, opened 2021-08-10
Jtran/de44472/fix discussion filter

JamieTran:
Kinda messy but there's 3 things I want to highlight here:

1.  Removed the `applyFilter` parameter on `_getDiscussionPostEntities()` as we decided we always want to apply the filter when selected
2. Fixed a weird bug where re-rendering the table would try to re-use elements (in our case the `post-score`). Changing from `itemTemplate.push()` to `repeat()` fixes this as each row now contains its own unique ID
table-input-states
3. Split `_renderTableBody()` into 3 smaller functions for easier readability `_renderTableRow`, `_renderTableBody`, `_renderTablePostScore`,


components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -327,6 +284,34 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
...
+		`;
+	}
+	/* eslint-disable indent */
+	_renderTableRow() {
+		if (typeof this.displayedDiscussionPostObjects === 'undefined' || (this.displayedDiscussionPostObjects && this.displayedDiscussionPostObjects.length === 0)) {
+			return html`<tr><td>${this._renderNoPostsInFilteredRange()}</td></tr>
+						<d2l-offscreen role="alert">${this.filteringStatus}</d2l-offscreen>`;
+		}
+		return html`
+			${repeat(this.displayedDiscussionPostObjects, (discussionPost) => discussionPost.postHref, (discussionPost) => {
```
JamieTran, line 304: Using postHref as the uniqueID, docs: https://lit.dev/docs/templates/lists/#the-repeat-directive

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
```diff
@@ -327,6 +284,34 @@ export class ConsistentEvaluationDiscussionPostPage extends SkeletonMixin(RtlMix
...
+			<td class="d2l-consistent-evaluation-discussion-post-score">
+				<d2l-consistent-evaluation-discussion-post-score
+					.discussionPostEntity=${discussionPostEvaluationEntity}
+					.postScore=${discussionPostEvaluationEntity.properties.score}
+				></d2l-consistent-evaluation-discussion-post-score>
+			</td>
+		`;
+	}
+	/* eslint-disable indent */
+	_renderTableRow() {
```
jeroach, line 298: nit: probably should name this `_renderTable`

since it renders the entire table instead of just a single row

jeroach (approved):
lgtm - thanks for cleaning this up!

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
JamieTran: haha yeah I swear naming is the hardest part 😆 ~ there is a  `_renderTable()`  [function](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/jtran/DE44472/fix-discussion-filter/components/left-panel/discussions/consistent-evaluation-discussion-post-page.js#L236), maybe something like `_renderTablePosts()` is more descriptive

jeroach: Sure, or `_renderTableRows`

Your choice :)

components/left-panel/discussions/consistent-evaluation-discussion-post-page.js
maddiegabriel: very cool, this `repeat` is way cleaner than what we had before 👍

maddiegabriel (approved):
Good stuff! LGTM

PR #628, opened 2021-08-17
Jroach/us130950

jeroach:
Organizing the demo page folder structure to match that of the components themselves

This looks like a total mess - so I've added an image of the folder structure below to help.

What changed:
- moved files to their respective folders (including data)
- updated file references to match new file locations
- updated the `index.html` page to also be organized by page area (e.g. header, footer, left-panel, right-panel)

image


jeroach:
@Scy-D2L I figured it would be best to do this first :)

I've confirmed that all demo pages are working as they did before - e.g. nothing is worse, nor better.

Now with it organized, we can tackle files more easily.

JamieTran (approved):
This is way tidier! Thanks for fixing this 💯

PR #649, opened 2021-08-31
basic demo page for nav bar

jeroach:
This really isn't much, but my efforts to get more than one demo scenario for the nav-bar didn't pan out.

So just timeboxing this work, and getting something in.


JamieTran (approved):

Scy-D2L:
Given the troubles with the way this component works, I think having multiple demo pages for different cases would be fine (if we need those different cases at some later time) but it's probably fine to just have the one.

PR #659, opened 2021-09-14
DE43824 - Retract with unsaved changes dialogue

tdevos-d2l:
[DE43824](https://rally1.rallydev.com/#/?detail=/defect/600942267527&fdp=true)

User is now prompted with a dialogue when retracting with unsaved changes as shown in the image below.

image

General flow for detecting unsaved changes and displaying the dialog was followed closely to when a user tries to navigate with unsaved changes.
Wording in the dialog was approved by Megan Shuker.


tdevos-d2l:
Not sure if a new unit test will be needed for these changes. Let me know if there is a new use case that would benefit from having one and I'll add it.

lang/en.js
```diff
@@ -156,9 +157,12 @@ export default {
 	"unflaggedFile": "{fileName} unflagged",
 	"unsavedAnnotationsBody": "Are you sure you want to leave without saving your annotation changes?",
 	"unsavedAnnotationsDiscardButton": "Discard annotation changes",
+	"unsavedChangesDiscardAndRetractButton": "Discard and Retract",
+	"unsavedChangesDiscardButton": "Discard changes",
 	"unsavedAnnotationsTitle": "Unsaved annotation changes",
```
maddiegabriel, line 162: Should we update this dialog to use title case too while we're at it? Along with `unsavedAnnotationsDiscardButton`

maddiegabriel (approved):
LGTM! 👍

JamieTran (approved):
Great work!!

PR #661, opened 2021-09-16
Add support nullable score rubric grade box

JamieTran:
When unselecting all criteria in a rubric (thus returning null as a total-score) the letter grade will auto sync to `None` instead of staying at the previous grade


jeroach (approved):
lgtm

PR #668, opened 2021-09-21
Hooked up the bypassRubricState variable for score updated events

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F609385638273

Rubrics PR: https://github.com/Brightspace/d2l-rubric/pull/1076

The problem was [here](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/helpers/rubricGradeSyncHelpers.js#L64). The event handling code was checking if a rubric was open or not before letting the score update event continue. If you closed the accordion fast enough the rubric would be considered closed before the event could propagate


components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -353,14 +353,13 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 	_syncActiveScoringRubricGradeHandler(e) {
 
 		const score = e.detail.score;
-
+		const bypassRubricState = e.detail.bypassRubricState;
```
JamieTran, line 356: This wouldn't work for our other use cases. It looks like you are always setting this to `true` when `d2l-rubric-total-score-changed` is fired. We don't want the syncing to happen if there are no rubrics open, this is mainly due to the fact that the `d2l-rubric-total-score-changed` is also fired on page load and navigating to a different student (which we don't want on sync).

PR #669, opened 2021-09-21
DE45202 - Rubric criteria sticking when navigating to next/previous student

tdevos-d2l:
[DE45202](https://rally1.rallydev.com/#/?detail=/defect/609267791893&fdp=true)
First, refreshed rubric as one of the last steps of loading the next student.
Then, made skeleton disappear only when the rubric is done refreshing. This makes sure there is no flickering in rubric score.


maddiegabriel (approved):
LGTM! 👍

components/consistent-evaluation.js
```diff
@@ -253,7 +253,13 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 				return;
 			}
 		}
-		this._loading = false;
+
+		await this._mutex.dispatch(
+			async() => {
+				this._refreshRubrics().then(() => this._loading = false);
```
vix633, line 259: So this definitely now moves setting `_loading` to `false` into this then. What if `_refreshRubrics` fails?

vix633 (approved):

PR #677, opened 2021-09-28
DE45362 - Rubric popout read-only when user doesn't have correct permissions

tdevos-d2l:
[DE45362](https://rally1.rallydev.com/#/?detail=/defect/609981079925&fdp=true)
LMS PR: [https://github.com/Brightspace/lms/pull/14532](https://github.com/Brightspace/lms/pull/14532)
With new changes in the backend, users can now see the popout rubric without the 'Give Feedback' permission.
This change now makes sure that if they don't have that permission, they cannot edit the popout rubric. Done exactly the same as the right-panel rubric, with the `read-only` attribute.


JamieTran (approved):
LGTM

components/right-panel/consistent-evaluation-popup-rubric.js
```diff
@@ -79,6 +81,14 @@ export class ConsistentEvaluationPopupRubric extends LocalizeConsistentEvaluatio
 			this.setTitle();
 
 			const evaluationEntity = await controller.getEvaluationEntity();
+
+			let hasWritePermission = false;
```
maddiegabriel, line 85: should this be called `hasGiveFeedbackPermission`?

tdevos-d2l: Sure, I'll change that.

maddiegabriel (approved):
👍

maddiegabriel (approved):
👍

PR #591, opened 2021-07-20
DE44007 - Assignments - Submission File downloaded does not include student name

amauryperez:

PR #684, opened 2021-09-30
DE45442 - Discussions - Fixing posts attached files broken link

amauryperez:
image


components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
```diff
@@ -165,11 +165,11 @@ export class ConsistentEvaluationDiscussionEvidenceBody extends RtlMixin(Localiz
 	_renderAttachments() {
 		if (this.attachmentsList) {
 			return html`${this.attachmentsList.map((attachment) => {
-				const { name, size, extension, href, downloadHref } = attachment.properties;
```
amauryperez, line 168: downloadHref does not exist for Discussions posts, href is the correct property to use here

JamieTran (approved):

components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
JamieTran: Is this something that changed or was it always `href` 🤔

tdevos-d2l (approved):

components/left-panel/discussions/consistent-evaluation-discussion-evidence-body.js
amauryperez: I think the discussions API has not been touched in a long time

amauryperez: Looks like it was me here https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/591. That is why I should not touch the front-end :-) . That changed was for assignment and somehow the discussion file was changed as well

PR #692, opened 2021-10-01
DE45341 - fix issue where rubric zero grade synced to None

JamieTran:
The issue was that when zero was graded on the rubric it would select the `None` option in the grade select box.
The problem was that `!newScore` returned true when `newScore` was 0
rubric-zero-select-box
.


tdevos-d2l (approved):
👍

PR #705, opened 2021-10-06
DE45506

vix633:
Multiple rubrics with no default now link to overall grade on load

Non scoring rubrics now will not reset overall grade


tdevos-d2l (approved):
LGTM 🚀

PR #716, opened 2021-10-13
DE45476 - rubric navigation fix

maddiegabriel:
[DE45476](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F610669761497
)

Previously, when navigating between students while a rubric is still saving, the last rubric criterion scored would be applied to the wrong student's `Overall Grade`. We fixed this by waiting on the siren action queue (`window.D2L.Siren.ActionQueue`) to empty before navigating away from the page.

These changes also include a fix to make the 'Unsaved Changes' dialog show up if there is any rubric criterion saving in the ActionQueue. This preempts the dialog not showing up when the 'unsaved' class doesn't get added to the evaluation entity in time. This causes an issue where the dialog can be shown even if no changes are made to the Overall Grade, when a user changes a non-active scoring rubric with slow internet and navigates to the next student.

~~However even with this fix, there is still an issue happening on Slow 3G internet only. Occasionally, the last rubric criterion scored is not being saved to the `Overall Grade` after navigating away from the page. This seems to be happening because there is a slight delay in the siren action being enqueued after a criterion is clicked by the user. This means that the queue can become artificially empty before the rubric is done saving. We've brought this up with Mango and they are taking a look. This PR adds the behaviour of waiting on the queue, so if the delay in enqueuing actions can be improved after Mango's investigation, this should be perfectly robust. As it is currently, this solution still greatly reduces the chance of saving and navigating before the rubric is done updating the Overall Grade.~~ **Mango has fixed this:** https://github.com/Brightspace/d2l-rubric/pull/1195

(for reference, this also seems to be the cause of the popout rubric issue [DE45564](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F611264039037))

fixed_navigation


components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -362,6 +362,7 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 				composed: true,
 				bubbles: true
 			}));
+			this._dispatchUpdateRubricChange(1);
```
tdevos-d2l, line 365: Why was this added?

PR #769, opened 2021-11-24
Fix bug where discussion page-load would overwrite grade with first rubric

JamieTran:


components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -171,7 +171,7 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 			this.rubricWindowPopout.close();
 		}
 
-		if (changedProperties.has('rubricInfos')) {
+		if (changedProperties.has('rubricInfos') && this.activityType !== 'discussionActivity') {
```
maddiegabriel, line 174: should we use the `discussionActivity` constant like in consistent-evaluation-page?
https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/bb5752023a28a29fa205d40261cf8c108cbf1875/components/consistent-evaluation-page.js#L544

maddiegabriel (approved):
LGTM

components/right-panel/consistent-evaluation-rubric.js
Scy-D2L: Typically I'd agree but I think the right-panel checking on the activity type is meant to be a temporary thing until we get the rubrics list enabled in discussions, so adding additional wiring for that might not be necessary 🤷🏼

Scy-D2L (approved):
lgtm

PR #778, opened 2021-11-30
DE43890 - making the grade component more a11y friendly

maddiegabriel:
[DE43890 - [Grade Component] The grade component is not A11y friendly when the user enters values out of bounds](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fdefect%2F601367018963)

See gif below for the new behaviour. Testing on Chrome and Firefox with NVDA screen reader (my JAWS is broken).

Screen reader announces the full error message:
1. When the user enters an invalid grade then clicks enter
2. When the user clicks the Save / Publish / Update buttons
3. When the user returns to the input box to re-enter the grade

If the user enters an invalid grade then navigates away from the input box, the screen reader says "invalid input". When the user tabs back to re-enter the grade, it reads the full error message.

NVDA_GRADE_THING


components/consistent-evaluation-page.js
```diff
@@ -480,7 +480,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 						@d2l-consistent-eval-active-scoring-rubric-change=${this._transientSaveActiveScoringRubric}
 					></consistent-evaluation-right-panel>
 				</div>
-				<div slot="footer">
+				<div slot="footer" aria-live="assertive">
```
maddiegabriel, line 483: @Scy-D2L using `aria-live` on the parent component of `d2l-alert-toast` seems to have worked 👍

JamieTran (approved):

Scy-D2L (approved):

PR #781, opened 2021-12-01
DE46107 rubric mismatch warnings

mpharoah-d2l:
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F615600529585

 * New Unsaved Dialog https://xd.adobe.com/view/1358a32a-699b-4e10-8d91-bd5927e82c73-460b/
 * Changes immediately visible - inline alert https://xd.adobe.com/view/1c49f605-e43c-4811-953a-248980d661e8-491a/
 * *(Moved out of scope)* Mismatching grades alert https://xd.adobe.com/view/1843172a-2193-4b1a-94a6-c227f6fdcb16-4d10/

This PR addresses the first two points by update the unsaved changes dialog when the rubric has changed and by passing a flag to d2l-rubric when the evaluation is published to show the warning.

The third point is not handled in this PR because it turns out that consistent eval has a bug where manually setting the topic score for a discussion topic does not set the flag in the database that the score has been overridden, so we are unable to tell the difference between an intentionally overridden score and an unintentional desync. This will be handled in a future change.


components/consistent-evaluation-page.js
```diff
@@ -452,6 +456,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 						.grade=${this._grade}
 						.gradeItemInfo=${this.gradeItemInfo}
 						.token=${this.token}
+						.evaluationPublished=${!!this._isEvaluationPublished()}
```
jeroach, line 459: Typo? do we need the double `!!` here?

I would have just expected a call to `this._isEvaluationPublished()`

components/consistent-evaluation-page.js
```diff
@@ -520,6 +526,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 				.publishClicked=${this._isPublishClicked}
 				.updateClicked=${this._isUpdateClicked}
 				.retractClicked=${this._isRetractClicked}
+				.rubricAssessmentModified=${!!this._isRubricAssessmentModified}
```
jeroach, line 529: Same comment as above re: the `!!`

jeroach (commented):
Looks solid overall - just a couple of small questions before approving :)

jeroach:
Oh wait, we're going to want to add some visual diff tests to cover the new dialogs :)

components/consistent-evaluation-page.js
mpharoah-d2l: The linter considers the type to be `boolean | undefined` without the `!!`, and complains.

components/consistent-evaluation-page.js
mpharoah-d2l: The linter considers the type to be `boolean | undefined` without the `!!`, and complains.

mpharoah-d2l: Ah. it's because it wants me to set it in the constructor. Just giving at default value isn't detected by the linter

mpharoah-d2l:
How do you run visual diff tests locally? Running it with mocha doesn't work for me

mpharoah-d2l:
Is a demo page required for this change? None of the dialogs currently have demo pages

jeroach:
> Is a demo page required for this change? None of the dialogs currently have demo pages

I'm fine with just vdiff tests - there isn't much logic to demo with the dialog (since clicking the buttons just cause it to hide)

jeroach (approved):

PR #794, opened 2021-12-03
US134164 - Adding margin to right of question UI for RTL

tdevos-d2l:
[US134164](https://rally1.rallydev.com/#/?detail=/userstory/618478120003&fdp=true)

## Before: non-RTL
image
## Before: RTL
image

## After: non-RTL
image
## After: RTL
image


vix633 (approved):
Marginally good

maddiegabriel (approved):

PR #795, opened 2021-12-03
upgrading profile-card to 1.0

dlockhart:
I've bumped profile-card to 1.0 so that consistent-eval and BSI (which both depend on it) can better stay in sync with a simple `^1` semver dependency. Otherwise every time it releases we need to coordinate the upgrade.

I'll update BSI (for the last time!) once this releases.


annabelsegalld2l (approved):
Thanks Dave! This will definitely save us time in the future 🎉

jeroach (approved):
lgtm

tdevos-d2l (approved):

PR #796, opened 2021-12-03
Scy/switch to esm and upgrade actions

Scy-D2L:
These are some updates to our dependendies inspired by difficulties getting local vdiff to run and just keeping up with the newer tech. Following the same process Dave L set up in [this PR](https://github.com/BrightspaceUILabs/user-profile-card/pull/100) for the `user-profile-card` to be consistent with the rest of the platform, I'll steal his update notes here:

### 1) New Visual Diff dependencies
The main motivation for this change. By switching to the new `main` branch of the actions repo, you pick up the latest visual-diff, mocha@9 and puppeteer@12 -- this will likely case minor visual differences.

### 2) Make this project use ES Modules by default
Mocha 9 (required for the new visual-diff action) requires it, and web projects shouldn't be CommonJS anyway. I've converted the things using CommonJS things (like `require()` and `__dirname`) to their ESM equivalents.

**Other minor updates**
While I was in here, I updated a few things like the latest dev-server and test-runner.

Out linting-configs remain out of date but require some more substantial refactoring so skipping that for now.


test/consistent-evaluation.test.js
```diff
@@ -1,7 +1,7 @@
 import '../consistent-evaluation.js';
 import { expect, fixture, html } from '@open-wc/testing';
 
-describe('d2l-consistent-evaluation', () => {
+describe.skip('d2l-consistent-evaluation', () => {
```
Scy-D2L, line 4: This test errors because CE with no data doesn't render correctly anyway. It's basically invalid and should be removed, but I kept it here for the moment as an example of how to structure a proper axe test until others are added.

web-test-runner.sauce.config.js
```diff
@@ -0,0 +1,56 @@
...
+
+const sauceLabsOptions = {
+	// eslint-disable-next-line no-undef
+	user: process.env.SAUCE_USERNAME,
+	// eslint-disable-next-line no-undef
+	key: process.env.SAUCE_ACCESS_KEY,
+};
+
+const sauceLabsCapabilities = {
+	name: '@brightspace-ui-labs/user-profile-card unit tests',
```
dlockhart, line 12: You'll want to change these two lines to be `@brightspace-hypermedia-components/consistent-eval` -- this is useful if you want to find your test runs in the Sauce web UI.

dlockhart (approved):
Amazing, thanks for doing this -- saves me one less place to switch over today! 🥇

amauryperez (approved):
I think it looks good

JamieTran (approved):

Scy-D2L:
Just noting here that we're getting flake from the profile card 🤦🏼 it consistently failed the other way a few times but now that it's merged it's back to this? I'm going to see if it flips again in a few runs.

PR #802, opened 2021-12-08
US133016 & US130329 - Quizzing attempt selection hookup

tdevos-d2l:
[US133016](https://rally1.rallydev.com/#/?detail=/userstory/612467325791&fdp=true) & [US130329](https://rally1.rallydev.com/#/?detail=/userstory/605344103921&fdp=true)

-Attempt selection now opens the respective Attempt/Completion Summary chosen.
-Navigating between students when the Completion Summary is chosen will show the completion summary of the next/previous student
-Navigating between students when any Attempt is chosen will show the first Attempt (Attempt 1) of the next/previous student
-If a user has no attempts and is navigated to, the Completion Summary will always be shown
-Also hooked up the entry point of Completion Summary vs Attempt (currently we don't know which attempt was clicked on entry, we just know that they clicked an attempt - waiting for [US134360](https://rally1.rallydev.com/#/?detail=/userstory/618920350837&fdp=true))

#### Gifs are a little hard to follow because the loading skeleton for LHP is not complete
z2VBZ9DKAT
Case where student has no submissions:
5ofhyftbjf


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -384,48 +384,50 @@ export class ConsistentEvaluationHrefController {
...
 
 		if (root && root.entity) {
 			root = root.entity;
 			if (root.getSubEntityByClass(Classes.quizzes.quizAttemptList)) {
 				const quizAttemptLinks = root.getSubEntityByClass(Classes.quizzes.quizAttemptList).links;
 
-				await Promise.all(quizAttemptLinks.map(async(attemptLink) => {
-					const quizAttemptEntity = await this._getEntityFromHref(attemptLink.href, false);
+				if (quizAttemptLinks) {
+					quizAttemptInfoList = {};
```
tdevos-d2l, line 395: This fixes the case where a student has no attempts. Previously it would error here

components/header/consistent-evaluation-attempt-context.js
```diff
@@ -52,16 +57,48 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
...
 		`;
 	}
 
-	_onAttemptChange() {
-		//TODO: We should dipatch and event to trigger a re-render of the left-hand panel quizzing
+	refreshQuizSelect() {
+		this.shadowRoot.querySelector('select').value = this.currentQuizAttemptNumber;
+	}
+
+	_renderAttemptOption(attempt) {
```
tdevos-d2l, line 70: Don't even ask about this.... silly lit related thing we've dealt with in the past.
Basically `?selected='...'` doesn't work reliably

components/header/consistent-evaluation-attempt-context.js
```diff
@@ -52,16 +57,48 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
...
+				composed: true,
+				bubbles: true
+			}));
+		} else {
+			const attemptNumber = e.target.value;
+			this._dispatchQuizzingAttemptSelectedEvent(attemptNumber);
+		}
+	}
+
+	_renderAttemptOption(attempt) {
```
tdevos-d2l, line 92: Don't even ask about this.... silly lit related thing we've dealt with in the past.
Basically `?selected='...'` doesn't work reliably

components/header/consistent-evaluation-attempt-context.js
```diff
@@ -52,16 +57,48 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
 
 		return html`
 			<select class="d2l-input-select d2l-truncate" aria-label=${this.localize('userAttempts')} @change=${this._onAttemptChange} >
-				${Object.values(this.quizAttemptInfoList).map(attempt => html`
-					<option value=${attempt.attemptNumber} label=${this.localize('quizAttempt', { num: attempt.attemptNumber })} ></option>
-				`)}
-				<option value="-1" label=${this.localize('completionSummary')} ></option>
+				${Object.values(this.quizAttemptInfoList).map(attempt => this._renderAttemptOption(attempt))}
+				<option value=${completionSummary} label=${this.localize('completionSummary')} ?selected=${this.currentQuizAttemptNumber === completionSummary}></option>
```
maddiegabriel, line 61: does this `?selected` work alright?

tdevos-d2l: It does, which I'm wondering if the broken `?selected` only happens when it's in a `.map`

maddiegabriel (approved):
looks good to me, good stuff!

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -384,48 +384,50 @@ export class ConsistentEvaluationHrefController {
...
-							extendedDeadlineExceededBy: quizAttemptEntity.entity.properties.extendedDeadlineExceededBy,
-							startDate: quizAttemptStartDateEntity.properties.date,
-							endDate: typeof quizAttemptEndDateEntity === 'undefined' ? undefined : quizAttemptEndDateEntity.properties.date,
-							isTimeEnforced: isTimeEnforced
-						};
-					}
-				}));
+							// if the attempt has not been completed, endDate is set to undefined
+							quizAttemptInfoList[attemptNumber] = {
+								attemptNumber: attemptNumber,
```
JamieTran, line 418: we're are using `parseInt` in 3 different places, does it make sense to just convert it at the beginning so we can use it as a number?

tdevos-d2l: The problem is that `currentQuizAttemptNumber` represents `completionSummary` as a string. I could represent that as a `0` or `-1` if you think that's a good idea

components/header/consistent-evaluation-attempt-context.js
```diff
@@ -52,16 +57,48 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
 
 		return html`
```
JamieTran, line 58: During your demo I saw that the dropdown disappears when there are no attempts. I don't think this is what we want. This is definitely my bad lol, this work doesn't have to exist in this PR.

components/controllers/ConsistentEvaluationHrefController.js
JamieTran: I like the idea of having it represented as `0` and having that be the `default` when there's no attempts

tdevos-d2l: Done!

JamieTran (approved):

JamieTran (approved):

components/consistent-evaluation.js
```diff
@@ -410,6 +419,26 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 		}
 		this._loading = true;
 	}
+	async _setQuizAttemptNumber() {
+		// TODO: handle entry points properly (use this.submissionId to find correct attempt number)
+		if (this.currentQuizAttemptNumber === undefined) {
+			if (this.submissionId === completionSummaryId || !this._quizAttemptInfoList) {
```
vix633, line 425: consider moving this check `!this._quizAttemptInfoList` to 424.

components/consistent-evaluation.js
```diff
@@ -410,6 +419,26 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
...
+				this.currentQuizAttemptNumber = completionSummary;
+			} else if (this._quizAttemptInfoList && this._quizAttemptInfoList[defaultQuizAttemptNumber]) {
+				this.currentQuizAttemptNumber = defaultQuizAttemptNumber;
+			}
+		} else {
+			if (this._quizAttemptInfoList) {
```
vix633, line 431: If you move that check, you can get rid of this if.

tdevos-d2l: I found a way of combining this whole if block into 
```
if ((this.currentQuizAttemptNumber === undefined && this.submissionId === completionSummaryId) || !this._quizAttemptInfoList) {
	this.currentQuizAttemptNumber = completionSummary;
} else if (this._quizAttemptInfoList && this._quizAttemptInfoList[defaultQuizAttemptNumber] && this.currentQuizAttemptNumber !== completionSummary) {
	this.currentQuizAttemptNumber = defaultQuizAttemptNumber;
}
```
But then I lose the separation of an entry point or navigating between students (`this.currentQuizAttemptNumber === undefined`)
Also, this will likely be completely changed very soon when attempt entry points are added so it might be silly to change this now. What do you think?

vix633: I'm ez, if you wanna leave it.

Scy-D2L (approved):

vix633 (approved):

PR #806, opened 2021-12-08
Force focus on displayName to prevent visual diff flakes in learner context bar

annabelsegalld2l:
Will re-run the visual diff tests a couple of times to ensure the flake is removed. I also ran around 50 times locally with no flaking so I am pretty confident the solution works.


test/perceptual/components/header/consistent-evaluation-lcb.visual-diff.js
```diff
@@ -36,6 +36,8 @@ describe('learner-context-bar', () => {
 			const listener = new Promise((resolve) => {
 				card.addEventListener('d2l-labs-user-profile-card-opened', resolve, { once: true });
 			});
+			// force focus on the name to prevent diff flakes
+			elem._focusName = true;
```
annabelsegalld2l, line 40: I did the same for the user-profile-card tests and they stopped flaking. The underlying issue is a race condition where sometimes the focus does not enter the focus-trap. Setting this private property (only used for tests) ensures the focus will be put on displayName every time.

test/perceptual/components/header/consistent-evaluation-lcb.visual-diff.js
```diff
@@ -36,6 +36,8 @@ describe('learner-context-bar', () => {
 			const listener = new Promise((resolve) => {
 				card.addEventListener('d2l-labs-user-profile-card-opened', resolve, { once: true });
 			});
+			// force focus on the name to prevent diff flakes
+			card._focusName = true;
```
annabelsegalld2l, line 40: I did the same for the user-profile-card tests and they stopped flaking. The underlying issue is a race condition where sometimes the focus does not enter the focus-trap. Setting this private property (only used for tests) ensures the focus will be put on displayName every time.

vix633 (approved):
This is great, thank you! :shipit:

jeroach (approved):
Thanks so much!

PR #810, opened 2021-12-13
Scy/lms version action

Scy-D2L:
Switching to the LMS-version action
This will make a dry-run to ensure the new versioning system is working correctly


vix633 (approved):
Tested

PR #812, opened 2021-12-14
Switch to lms-version-action

Scy-D2L:
Once again: Switching to the LMS-version action
This will make a dry-run to ensure the new versioning system is working correctly


JamieTran (approved):
lgtm💯

PR #815, opened 2021-12-14
DE45776 - Fixing assignment submission out of sync with LHP

tdevos-d2l:
[DE45776](https://rally1.rallydev.com/#/?detail=/defect/612991311769&fdp=true)

I have fixed this a few times in various differing places we use the `select` component. For an `option` tag we generally use `?selected="..."` but this often causes issues in Lit where the incorrect `option` is shown as selected even though the correct one has the `selected` attribute shown in the dev console.
Removing the `?selected` and hardcoding `selected` into the correct `option` tag fixes this, as done in grading rubric dropdown and Quizzing attempt dropdown.


maddiegabriel (approved):
Nice 👍

JamieTran (approved):

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -161,6 +159,17 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeConsist
 			/* forceTriggerOnCancel: */ false
 		);
 	}
+	_renderFileOption(sf) {
+		if (sf.properties.id === this.currentFileId) {
+			return html`
+				<option value=${sf.properties.id} label=${this._truncateFileName(sf.properties.name)} selected class="select-option"></option>
```
Scy-D2L, line 165: I think this could be simplified to  `${<condition>? 'selected' : ''} ` inside the HTML string with less duplication

tdevos-d2l: This is probably a good idea. And I can change it everywhere we do this to that. Unfortunately my local build is not working anymore so I'll get to this tomorrow.

Scy-D2L: This was causing some weird issues so we can come back to it later.

Scy-D2L (approved):

PR #821, opened 2022-01-05
Update html-block API usage

grant-cleary:
See this PR for context: [BrightspaceUI/core#1960](https://github.com/BrightspaceUI/core/pull/1960)

Slight change to the html-block API, in that it now takes its contents directly in a liteDOM slot vs. within a template. Otherwise behaviour should remain the same.


grant-cleary:
> * [x]  I have made the necessary changes to the Demo pages
No changes needed - demo page doesn't reference the html-block directly
> * [x]  I have updated the tests to reflect the changes in this PR (Unit & Visual-diff)
Same as above. This change should carry no visual changes so any existing tests should pass as is.
> * [x]  I have included langterms where applicable
N/A.

grant-cleary:
Oh, but I'm dumb and forgot to bump the core version.

grant-cleary:
Mm, looks like none of core's subdependencies changed, so I'm going to roll back the package-lock update to only include the core change. Something about the other changes is making the tests angry.

JamieTran (approved):
Looks good, thanks!

PR #825, opened 2022-01-06
DE45765 - avoid transient save for rubric on discussions scored by post

maddiegabriel:
[DE45765 - CE Discussions - When scoring rubric linked to Discussion set to grade individual posts an error is thrown in the Browser console](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F612902763509)

When a discussion has a rubric and is set up to score posts individually, the rubric cannot be an active scoring rubric ([we don't render the dropdown](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/b5b9bd7d47583ad72f71a21f27268f5b3a21e7d5/components/right-panel/consistent-evaluation-rubric.js#L271)). But we were still trying to update the overall grade when the rubric was changed, causing an [error to be thrown](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/b5b9bd7d47583ad72f71a21f27268f5b3a21e7d5/components/controllers/ConsistentEvaluationController.js#L293) from `transientSaveGrade` because the evaluation entity doesn't have a `SaveGrade` action. This change avoids that, not sure if it's the best solution though!

image (4)


JamieTran (approved):
lgtm!

amauryperez (approved):

vix633 (approved):

PR #829, opened 2022-01-12
US135051 - Quizzing UI - Adding mobile view for attempts dropdown

maddiegabriel:
[US135051 - [Consistent Eval][Quizzing] Make quizzing attempts dropdown responsive](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fuserstory%2F621940993305)

In CE Assignments, we hide the file context dropdown for screens smaller than 930px to reduce clutter. This works because the user has the "Back to User Submissions" button to navigate between submissions.
image

However in CE Quizzing, the dropdown is currently the only way to navigate between attempts without leaving CE and coming back. Spoke with Gilberto about this, he asked for the dropdown should persist for smaller screen sizes in Quizzing 👍 

mobile_quiz_dropdown


components/header/consistent-evaluation-attempt-context.js
```diff
@@ -42,8 +42,8 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
 					white-space: nowrap;
 				}
 				@media (max-width: 930px) {
-					:host {
-						display: none;
+					.d2l-input-select {
+						max-width: 9rem;
```
jeroach, line 46: How did we decide on this size for `max-width`?

maddiegabriel: Good point, that size was leftover from hiding the dropdown on assignments. This code just shrinks the dropdown width so it fits better on smaller screens. I'll change it to `max-width: 767px` which is the mobile breakpoint we use elsewhere in CE

JamieTran (approved):
lgtm!

components/header/consistent-evaluation-attempt-context.js
jeroach: fyi I'm assuming that this update will be a part of this PR, and that's why I'm holding off on approving for now :)

let me know if that's not the case.

maddiegabriel: oops I thought I pushed it up already, thanks for the reminder 😆

jeroach (approved):

PR #830, opened 2022-01-17
Update to `v4` of `d2l-rubric`

devpow112:
Updating to lit version of rubrics (`4.x`).


devpow112:
> Any notable changes in this major version?

It's now lit instead of polymer. Overall the external uses are supposed to be the same. We've been testing it with `consistent-evaluation` and others to make sure that's the case.

MitchelPaulin (approved):

mpharoah-d2l (approved):

devpow112:
Activities changes merged and needed to update it to fix a slight weirdness in the package-lock that was gonna trip up BSI.

PR #836, opened 2022-01-25
Jroach/us135593

jeroach:
This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
- [x] audit `ifDefined` usages for null parameters

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.


jeroach:
Regarding `audit ifDefined usages for null parameters`

There are over 80 usages of `ifDefined` in CE - I don't think it's practical to check them all programmatically. I think in most cases we handle if the value is `undefined` - I'd hope this will cover the `null` case as well.

If anyone has a good suggestion here I'm all ears :)

annabelsegalld2l:
> Regarding `audit ifDefined usages for null parameters`
> 
> There are over 80 usages of `ifDefined` in CE - I don't think it's practical to check them all programmatically. I think in most cases we handle if the value is `undefined` - I'd hope this will cover the `null` case as well.
> 
> If anyone has a good suggestion here I'm all ears :)

What I did in the activities repo was check them all manually. It was actually pretty quick - in 90% of the cases, null should be treated as undefined. As long as there is no mention of passing null into ifDefined you are usually good. 

Example of an okay case: ifDefined(this.name) where this.name is a string property. Null is not one of the valid cases, and can be treated as undefined. 

Example of a case where you need to make changes: If you set this.name to null to mean there is no name for an explicit reason, and want to pass along the null value. 

Hope those examples make some sense 😆 . Generally it will take a couple seconds to verify each case if it's a Lit property that gets checked.

components/header/d2l-consistent-evaluation-lcb-file-context.js
```diff
@@ -146,8 +148,13 @@ export class ConsistentEvaluationLcbFileContext extends RtlMixin(LocalizeConsist
...
 		];
 
+		let opener;
+		if (this.shadowRoot) {
+			opener = this.shadowRoot.querySelector('d2l-button-subtle');
+		}
+
 		D2L.LP.Web.UI.Legacy.MasterPages.Dialog.Open(
-			/*               opener: */ this.shadowRoot.querySelector('d2l-button-subtle'),
+			/*               opener: */ opener,
```
Scy-D2L, line 157: How does this respond if `opener` isn't defined?

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -301,8 +301,11 @@ export class ConsistentEvaluationRightPanel extends SkeletonMixin(LocalizeConsis
 			this.grade,
 			newScore
 		);
-		const gradeResult = this.shadowRoot.querySelector('d2l-consistent-evaluation-right-panel-grade-result');
-		gradeResult.initializeScore(newGrade.score);
+
+		if (this.shadowRoot) {
+			const gradeResult = this.shadowRoot.querySelector('d2l-consistent-evaluation-right-panel-grade-result');
+			gradeResult.initializeScore(newGrade.score);
```
Scy-D2L, line 307: This looks good but with all of the grief the rubric/grade setup has given us I think we'll want to focus some regression testing on the grade result box.

components/header/d2l-consistent-evaluation-lcb-file-context.js
annabelsegalld2l: We can actually remove this shadowRoot check and not worry that it would be null. This function is just an event listener, so it can only be triggered once the shadowRoot is in place. For any event listeners, I haven't added a null check. Adding an extra check never hurts, but in this case it's not explicitly required 😄

jeroach: Good question...I'll test locally to confirm.

Though it's worth mentioning that if it ever was null, we'd have bigger problems

jeroach: Ah I see, thanks Annabel! I'll revert these types of changes

jeroach:
> > Regarding `audit ifDefined usages for null parameters`
> > There are over 80 usages of `ifDefined` in CE - I don't think it's practical to check them all programmatically. I think in most cases we handle if the value is `undefined` - I'd hope this will cover the `null` case as well.
> > If anyone has a good suggestion here I'm all ears :)
> 
> What I did in the activities repo was check them all manually. It was actually pretty quick - in 90% of the cases, null should be treated as undefined. As long as there is no mention of passing null into ifDefined you are usually good.
> 
> Example of an okay case: `ifDefined(this.name)` where `this.name` is a string property. Null is not one of the valid cases, and can be treated as undefined.
> 
> Example of a case where you need to make changes: If you set `this.name=null` to mean there is no name for an explicit reason, and want to pass along the null value.
> 
> Hope those examples make some sense 😆 . Generally it will take a couple seconds to verify each case if it's a Lit property that gets checked.

Thanks for the clarity - this does help!

The vast majority of `ifDefined` usages in CE are indeed on `String` variables, so those should be fine.

I found a few instances where the type of the variable passed to `ifDefined` is a `Number` - for example:

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/left-panel/quizzing/consistent-evaluation-quizzing-attempt.js#L204

Is `Number` handled the same way as `String` or do we need to do something to handle these cases? Thanks!

annabelsegalld2l:
Numbers should be fine as well! As long as they are not being explicitly set to null (I can't think of a case where we would do that).

components/left-panel/assignments/consistent-evaluation-assignments-evidence-text.js
```diff
@@ -40,6 +40,11 @@ export class ConsistentEvaluationAssignmentsEvidenceText extends LocalizeConsist
 	}
 
 	render() {
+		let content = '';
```
JamieTran, line 43: Does this make content empty every time we render text assignments?

jeroach: This is just a placeholder value - assuming `this.content` has a value, then `content` will be assigned just below

JamieTran: derp you're right haha, I got confused since they both have the same name

JamieTran (approved):
lgtm

Scy-D2L (approved):

annabelsegalld2l (approved):
Looks good, thanks Josh!

components/left-panel/assignments/consistent-evaluation-assignments-evidence-text.js
jeroach: it's okay :D

PR #839, opened 2022-02-01
Us134340/jtran poc assignments refactor

JamieTran:
This is the first passthrough of the POC refactor, this PR focuses on decoupling the `downloadAllSubmission` button from CE and having it only live in the left-panel. This sets up the pattern of not having to instantiate all the data in our top-level component `consistent-evaluation.js` but will redistribute dependencies to the left and right panels.

- [x] Factor out `downloadAllSubmission`
- [x] Create left-panel controller with tests

The plan moving forward is to be proactive with quiz development with quiz attempts and questions


components/left-panel/consistent-evaluation-left-panel.js
```diff
@@ -80,14 +85,23 @@ export class ConsistentEvaluationLeftPanel extends SkeletonMixin(LocalizeConsist
 			return html`${this._renderQuizzes()}`;
 		}
 	}
+
+	async updated(changedProperties) {
+		super.updated();
+		if (changedProperties.has('href')) {
+			const leftPanelController = new ConsistentEvaluationLeftPanelController(this.href, this.token);
+			this._downloadAllSubmissionLink = await leftPanelController.getDownloadAllSubmissionLink();
```
JamieTran, line 93: Passing down the `this.href` so that we can make the API call closer to component rather than in `consistent-evaluation.js`

maddiegabriel (approved):
Nice! this is looking good so far to me 👍

vix633:
I've looked this over as well last week. Should merge this soon, before it gets too stale.

vix633 (approved):

PR #840, opened 2022-02-04
updateComplete is not a function

jeroach:
As per Slack message from Annabel :)


annabelsegalld2l (approved):
Thanks for jumping on this Josh!

JamieTran (approved):

PR #841, opened 2022-02-04
Pass the outcomes title to rubrics

MitchelPaulin:
Rubrics needs to know what the current outcome term is, passing that information along


components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -322,6 +326,7 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 						@d2l-rubric-finished-saving=${this._syncActiveRubricChangeHandler}
 						?ce-popout=${this.isPopout}
 						?show-published-warning=${this.evaluationPublished}
+						outcomes-title-text=${this.outcomeTerm}
```
vix633, line 329: Do we want to handle this if it has no value?

MitchelPaulin: Can it have no value? To be safe I could pass it through a localize call which has a default https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/lang/en.js#L102

MitchelPaulin: Actually that might break if we try to re-localize it on our end, I'll just default it to "Standards'

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -177,18 +181,19 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
...
+			if (rubricInfos && rubricInfos.length === 1) {
 				this._dispatchInitScoringRubric();
 			}
 			if (this.activeScoringRubric !== undefined
 				&& this.activeScoringRubric !== 'undefined'
 				&& (changedProperties.get('activeScoringRubric') !== undefined
 				|| changedProperties.get('activeScoringRubric') !== 'undefined')) {
 				if (this.activeScoringRubric === 'null' || this.activeScoringRubric === this.unassignedActiveScoringRubricId) {
-					this.activeScoringRubric = changedProperties.get('rubricInfos').filter(rubric => rubric.rubricScoringMethod !== 0)[0].rubricId;
+					this.activeScoringRubric = rubricInfos.filter(rubric => rubric.rubricScoringMethod !== 0)[0].rubricId;
```
MitchelPaulin, line 196: Kept getting console errors dumped here, its possible for the key `rubricInfos` to be in the map but its actual value to be undefined, adding another check for that here

JamieTran (approved):

PR #864, opened 2022-03-02
US136699 - Quizzing - Use questionNumber from API

maddiegabriel:
https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F628155253149

Previously the question number was being identified by a loop index, which assumed the question responses were coming back in the desired order. [We added](https://github.com/Brightspace/lms/pull/19101) a `questionNumber` prop to the attempt response entity - this PR updates the front end to use that instead.


vix633 (approved):
LGTM

JamieTran (approved):
lgtm!

PR #869, opened 2022-03-03
Scy/refactor rubric close

Scy-D2L:
It's generally a bad practice to dive into another component's `shadowDom` to do things, or to use `_private` functions of other components, even if they live inside CE. This is a small poc refactor to fix a spot where we're doing this - more will come in relation to the refactor work.


components/consistent-evaluation-page.js
```diff
@@ -367,6 +367,9 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 	connectedCallback() {
 		super.connectedCallback();
 		window.addEventListener('beforeunload', this.unsavedChangesHandler);
+		requestAnimationFrame(() => {
```
Scy-D2L, line 370: This is a way of ensuring the full dom exists before we try to grab the `_rightPanel`.

maddiegabriel (approved):
much better 👌 lgtm!

JamieTran (approved):

vix633 (approved):

PR #872, opened 2022-03-04
Call `closeRubric` as `_closeRubric` no longer exists

devpow112:
Looks like this was missed as part of https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/869.


JamieTran (approved):
Thank you for finding this, we must have missed it when refactoring the rubric closing

PR #873, opened 2022-03-04
Sync right-panel feedback & grade-result with quizzing

JamieTran:
The right-panel can now use the evaluation entity from selected attempts and completion summary

TODO:
Loading state on attempt switch
Sync footer with completion summary (will currently error due to the fact that the completion summary does not support publish/retract/update/save draft
rightpanelquizzing
)


components/consistent-evaluation.js
```diff
@@ -120,6 +120,7 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 		this._mutex = new Awaiter();
 		this._loading = true;
 		this._pageTitle = '';
+		this._evaluationHref = undefined;
```
Scy-D2L, line 123: Not specific to this PR, but are we sure that we need to do this. 🤔 I see we're doing it for a lot of properties, but if we don't explicitly tell the constructor to initialize them as `undefined`, shouldn't they be `undefined` anyway?

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -356,9 +356,11 @@ export class ConsistentEvaluationHrefController {
 						const quizAttemptEntity = await this._getEntityFromHref(attemptLink.href, false);
 						if (quizAttemptEntity && quizAttemptEntity.entity) {
 							const attemptNumber = quizAttemptEntity.entity.properties.attemptNumber;
+							const attemptEvaluationHref = this._getHref(quizAttemptEntity.entity, Rels.Activities.evaluation);
```
Scy-D2L, line 359: There is a [related unit test](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/jtran/refactor-evaluatuionhref-quizzing/test/ConsistentEvaluationHrefController.test.js#L460) that should be updated.

components/consistent-evaluation.js
JamieTran: Not entirely sure, I always thought it was best practice to define the variables before use but I've seen this pattern since joining CE. I did a quick sanity check with removing them and didn't see any changes but didn't extensively test.

Scy-D2L: To my understanding it's a best practice to declare/define them before use, but defining a variable as undefined literally doesn't do anything so it's a weird grey area.

While we're doing our refactors this seems worth looking into 👀 not req for this PR though.

Scy-D2L (approved):

jeroach (approved):
lgtm

PR #880, opened 2022-03-08
Fixes for `d2l-outcomes-overall-achievement` and vdiff

devpow112:
Grabs new version of https://github.com/Brightspace/d2l-outcomes-overall-achievement/pull/297 with a re-worked render to get around what ever issue is coming up in vdiff. Could only get it to happen in vdiff. All demos and use in the LMS were fine. Also explicitly adds `lit-html` as it was used but not referenced.


JamieTran (approved):
lgtm

PR #884, opened 2022-03-08
Remove rubrics `legacy` attribute and related variables - US133770

imcdonald-d2l:
This attribute is being controlled by the `rubrics-inline-grading` [flag](https://github.com/Brightspace/lms/blob/5f7a3c48f8b2e87d0fbe25a5a0b8a5708adc42fc/le/activities/D2L.LE.Activities/FeatureToggles/RubricsInlineGradingRevampFeature.cs) which we are removing in the back-end. This attribute is only in `3.x` rubrics (polymer) and set to `false` by default [here](https://github.com/Brightspace/d2l-rubric/blob/3.x/d2l-rubric.js#L475) so we no longer need to pass this through now that the flag will be removed.  

back-end will be updated when this is merged


devpow112 (approved):

imcdonald-d2l:
back-end PR: https://github.com/Brightspace/lms/pull/20272

imcdonald-d2l:
@Scy-D2L @maddiegabriel would I be able to get a second approval?

maddiegabriel (approved):
LGTM!

PR #889, opened 2022-03-10
DE47648 - quizzing - sort questions by questionNumber

maddiegabriel:
[DE47648 - [Consistent Evaluation][Quizzes] Questions loading incorrectly when switching between attempts](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fdefect%2F628615671665
)

@Scy-D2L and I looked at this issue.


components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js
```diff
@@ -108,7 +108,8 @@ export class ConsistentEvaluationLeftPanelController {
 			}
 		}
 
-		return questions;
+		const sortedQuestions = questions.sort((a, b) => a.questionNumber - b.questionNumber);
```
maddiegabriel, line 111: The question responses are coming back in the correct order from the API. The map function (line 93 above) deals with the question responses in the order the promises come back and pushes their info onto the `questions` array in that order, sometimes resulting in the question order being incorrect on the attempt page. So we thought sorting the array afterwards was the best call

Scy-D2L (approved):
Nit: we could write a test for this, but I'm not sure how valuable that would be (unless we're planning on having a sort option - which isn't in our current designs).

LGTM

JamieTran (approved):
I think this is okay for now (M1). I know for M2 we want to add question filtering and that would live closer to the component like we do with [discussions](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/left-panel/discussions/consistent-evaluation-evidence-discussion.js#L263) but out of scope here.

PR #891, opened 2022-03-14
Hide average time taken in completion summary for quiz attempts in progress

Girujan1998:
Rally: https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F628624126081
Constants PR: https://github.com/Brightspace/d2l-hypermedia-constants/pull/171

Matching functionality in legacy for Completion Summary, in regards to pending evaluation, published evaluation, and in progress attempts to display Average Time Taken field.

**Cases we are checking for to display or not display "Average Time Taken" field:**
Published Evaluation Attempt and In Progress Attempt -> display average time taken
Pending Evaluation Attempt and In Progress Attempt -> hide average time taken
Pending Evaluation Attempt and Published Evaluation Attempt -> display average time taken
Pending Evaluation Attempt and Pending Evaluation Attempt -> hide average time taken
Pending Evaluation Attempt -> hide average time taken
Published Evaluation Attempt -> display average time taken
In Progress Attempt -> hide average time taken

image

**Additional Fix**
A fix is also implemented in this PR to display zero percentage evaluations in the Completion Summary. Previously, it would just display nothing for each grade statistic if the published evaluation was zero. 

Fix located here: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/812cab998b5ea7081f5bb4f5dc1e434b0bec81ba/components/left-panel/quizzing/consistent-evaluation-quizzing-overall-summary.js#L160

**Before**

**After**


components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js
```diff
@@ -16,6 +16,69 @@ export class ConsistentEvaluationLeftPanelController {
 		this.token = token;
 	}
 
+	async getAverageTimeTakenVisibility() {
```
Scy-D2L, line 19: This is big enough that it needs a test.

components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js
```diff
@@ -16,6 +16,69 @@ export class ConsistentEvaluationLeftPanelController {
...
+						&& attemptEntity.entity.hasLinkByRel(Rels.Activities.evaluation)) {
+						const evaluationEntityLink = attemptEntity.entity.getLinkByRel(Rels.Activities.evaluation).href;
+						const evaluationEntity = await this._getEntityFromHref(evaluationEntityLink, false);
+
+						// evaluation pending - do not display field
+						if (evaluationEntity
+							&& evaluationEntity.entity
+							&& evaluationEntity.entity.properties
+							&& evaluationEntity.entity.properties.state === Classes.activities.unevaluated) {
+							return false;
```
Scy-D2L, line 48: This logic (37-49) is almost duplicated below (57-69). It may be worth splitting out a separate function `getQuizAttemptEvaluationState()`. This should also simplify writing a test for this function. :)

components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js
Girujan1998: Do we want a test/demo of displaying and not displaying "Average Time Taken" field? I kind of created a test case for displaying the field with `Quiz Overall Summary - Published Evaluation`, but grouped it with published evaluation instead of specifically looking at displaying and not displaying the field.

Scy-D2L: You're right that the `vdiff test` cases will catch changes, and I think those are fine the way they're set up now, but those tests are meant to handle the visual results of specific display components. Because this logic lives in a controller and involves a fair amount of interaction with the API I think it would be good to have a `unit test` to protect against unexpected changes in the future.

PR #898, opened 2022-03-23
US131413 - Quizzing - Save question feedback

maddiegabriel:
[US131413 - [ConsistentEvaluation][Quizzes][UI] Connect Quiz Question Feedback to API](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F608063986497)


components/consistent-evaluation-page.js
```diff
@@ -1184,6 +1185,40 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+	async _transientSaveQuestionFeedback(e) {
+		await this._mutex.dispatch(
+			async() => {
+				let questionResponseEntity;
+				const newFeedbackText = e.detail.feedback;
+
+				const evaluationEntity = await this._controller.fetchEvaluationEntity(false);
+				if (evaluationEntity) {
+					const evidenceEntities = evaluationEntity.getSubEntityByRel(evidenceRel).getSubEntitiesByRel(evidenceRel);
+					for (const i in evidenceEntities) {
```
maddiegabriel, line 1197: mimicking what we do to save discussion post scores (`_transientSaveDiscussionPostScore`)

components/consistent-evaluation-page.js
```diff
@@ -1184,6 +1185,40 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
+							break;
+						}
+					}
+				}
+
+				this.evaluationEntity = await this._controller.transientSaveQuestionFeedback(questionResponseEntity, newFeedbackText);
+
+				if (this.evaluationEntity && this.evaluationEntity.hasSubEntityByRel(evidenceRel)) {
+					const leftPanelController = new ConsistentEvaluationLeftPanelController(this.href, this.token);
+					const quizAttemptInfo = await leftPanelController.getQuizAttemptInfo(this.currentQuizAttemptHref);
```
maddiegabriel, line 1209: this works but we might want to consider splitting up the functionality in `getQuizAttemptInfo()` in the LHP controller?

that way we could refresh the question list being displayed without also having to also get all the other quiz attempt info (time limits, etc.)

components/controllers/constants.js
```diff
@@ -74,4 +74,4 @@ export const lowestGrade = 'lowest';
 export const averageGrade = 'average';
 export const firstGrade = 'first';
 export const lastGrade = 'last';
-export const questionResponseRel = 'question-response';
+export const questionResponseClass = 'question-response';
```
maddiegabriel, line 77: just renaming constant for accuracy

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
```diff
@@ -169,6 +198,19 @@ export class ConsistentEvaluationQuizzingAttemptResult extends SkeletonMixin(Loc
...
 
+	_saveOnFeedbackChange() {
+		if (this.shadowRoot) {
+			const feedback = this.shadowRoot.querySelector('d2l-htmleditor').html;
+			if (this.feedbackText === feedback) {
+				return;
+			}
+			this._debounceJobs.feedback = Debouncer.debounce(
+				this._debounceJobs.feedback,
+				timeOut.after(800),
```
jeroach, line 209: How did we choose the timeout length? Sorry if it's been used elsewhere and I just missed it.

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
```diff
@@ -105,6 +113,28 @@ export class ConsistentEvaluationQuizzingAttemptResult extends SkeletonMixin(Loc
...
+		}
+	}
+
+	_emitFeedbackEditEvent(feedback) {
+		this.dispatchEvent(new CustomEvent('d2l-consistent-eval-question-feedback-changed', {
+			composed: true,
+			bubbles: true,
+			detail: {
+				feedback: feedback,
+				questionResponseHref: this.questionResponseHref
```
jeroach, line 133: Is `this.questionResponseHref` always guaranteed to be set if we fire this event? 

Should it be passed in to `_emitFeedbackEditEvent` along with `feedback`?

Is it even a problem if the event is fired with a null value for `this.questionResponseHref`?

components/consistent-evaluation-page.js
jeroach: What you're suggesting makes sense in general to me, but I don't have enough context on the underlying logic to give good advice here :)

I shall defer to others more knowledgeable on the topic ;)

JamieTran: I think that's a good call. I am not a fan of instantiating a left-panel controller on the main page just to get this data. Maybe this can exist in `ConsistentEvaluationController.js` or better yet a helper file if it's used in multiple places.

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
maddiegabriel: Good question, this uses the same timeout as the right panel htmleditor here:
https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/725e81b9779cc18d1bbf1081f1db82ba843bb254/components/right-panel/consistent-evaluation-right-panel-feedback.js#L202

components/consistent-evaluation-page.js
maddiegabriel: that makes sense JT, working on it now 👍

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
jeroach: Sounds good - thanks for addressing my questions :)

components/consistent-evaluation-page.js
vix633: ew who uses `i` for this, clearly it should be `evidenceEntity`

jeroach: it's the ghost of VIKTOR

maddiegabriel: 🚪 👈 😎

components/consistent-evaluation-page.js
maddiegabriel: @JamieTran I pushed some updates, I addressed the issue we were talking about in DM's by passing the evaluation entity when the `d2l-consistent-eval-on-question-feedback-transient-save` event fires from `c-e-page.js`. This way we instantiate the LHP controller and fetch only the new questions list from `_updateQuestionList()` in `c-e-quizzing-attempt.js`

maddiegabriel:
adding @Scy-D2L since Josh is off till tues 😎 👍

JamieTran (approved):
lgtm

jeroach (approved):
No concerns with the new logic, looks to just be a few vdiff test issues

PR #906, opened 2022-03-25
Add Quiz Event Log menu item to overflow menu

Girujan1998:
Rally: https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F608939929551

Adding Quiz Event Log to overflow menu and when selected open dialog (rubric style). 

**Overflow Menu**
image
**Quiz Event Log Dialog**
image


components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -542,6 +604,30 @@ export class ConsistentEvaluationRightPanel extends SkeletonMixin(LocalizeConsis
 			></consistent-evaluation-right-panel-skeleton>`;
 	}
 
+	async _setQuizEventLogPath() {
```
amauryperez, line 607: I probably dont know what I'm talking about but, have we tried splitting out things that are tool specific from the right panel class?

Girujan1998:
Not really sure how to approach VDiff/Demo for this PR. If anybody has any suggestions towards if we need them for overflow menu.

components/right-panel/consistent-evaluation-right-panel.js
Girujan1998: What do you mean by splitting it out? Like a helper function in a separate file?

amauryperez: yeah maybe, we could also try to not set properties such as `_quizEventLogPath` that are only applicable to quizzing. not sure how feasible any of this is, just throwing some ideas around :-)

amauryperez: Anyways, whatever @JamieTran or @Scy-D2L say will be OK

Girujan1998: I do agree to kind of separate the logic function, so the file doesn't get cluttered with right panel functionality for discussion/assignment/quiz. I think we may still need the property due to anytime `this.evaluationHref` changing to recall the function and have a local property path updated to be used in the non-synchronous render.

JamieTran: As of right now, there hasn't been any refactoring of the right-panel but I think this is a great place to start so that we're not adding on to the problem. 
What we can do is create a right-panel controller that contains this method, we would still call `_setQuizEventLogPath` when `evaluationHref` has changed but through the controller instead of in this component.

components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -316,6 +335,10 @@ export class ConsistentEvaluationRightPanel extends SkeletonMixin(LocalizeConsis
 		}
 	}
 
+	async _getEntityFromHref(targetHref) {
```
JamieTran, line 338: This would live in the right-panel controller instead

components/controllers/right-panel/ConsistentEvaluationRightPanelController.js
```diff
@@ -0,0 +1,45 @@
...
+
+		if (typeof evaluationHref !== 'string') {
+			throw new Error(ConsistentEvaluationHrefControllerErrors.INVALID_TYPE_BASE_HREF);
+		}
+
+		this.evaluationHref = evaluationHref;
+		this.token = token;
+	}
+
+	async setQuizEventLogPath() {
```
JamieTran, line 18: nit: rename `set` to `get`

JamieTran (commented):
Looks good just need to add tests for this method & controller

JamieTran (approved):
lgtm

components/controllers/right-panel/ConsistentEvaluationRightPanelController.js
```diff
@@ -0,0 +1,45 @@
...
+						}
+					}
+				}
+			}
+		}
+
+		return null;
+	}
+
+	async _getEntityFromHref(targetHref) {
```
maddiegabriel, line 42: this should probably have a `bypassCache` param like the one in the LHP controller

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/329fe5964b71633f91189edbdf174919bd376edf/components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js#L302

maddiegabriel (approved):
great stuff! 👏

amauryperez (approved):

PR #913, opened 2022-03-28
Hide average time taken in completion summary for quiz attempts in progress

Girujan1998:
Rally: https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F628624126081

Fixing display issues due to for loop checking only one attempt.

Other PRs:
PR in CE: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/891
PR in d2l-hypermedia-constants: [Brightspace/d2l-hypermedia-constants#171](https://github.com/Brightspace/d2l-hypermedia-constants/pull/171)


maddiegabriel (approved):
good catch - lgtm!

Scy-D2L:
I support Maddie's suggestion of moving the information into `quizSummaryInfo`. Looking at this it's actually a bad pattern for a controller to be dictating whether a display property is shown in such a direct way. On the same note - passing in "average time taken" and "should we display the average time taken" is a bit weird, we should try either not setting the average time or returning undefined when it isn't meant to be displayed, sort of the way the hypermedia only returns relevant info.

Girujan1998:
> that's a good point Girujan! but I think Scy's idea sort of addresses that concern by only sending back `averageTimeTaken` with the summary info. + the benefit of making less API calls outweighs
> 
> I also had a question about the expected behaviour here, I might just be confused - do we only want to show the average time taken when there is at least one published attempt? and hide it in all other cases?

I agree with what you and Scy are saying by using `averageTimeTaken` instead of having an` average time taken value` and a `should we display the average time taken value` fields. I am going remove the visibility property I added and instead set `averageTimeTaken` to null/undefined if we do not want to display the value. 

The expected behavior cases are the following when I checked the legacy behavior :
Published Evaluation Attempt and In Progress Attempt -> display average time taken
Pending Evaluation Attempt and In Progress Attempt -> hide average time taken
Pending Evaluation Attempt and Published Evaluation Attempt -> display average time taken
Pending Evaluation Attempt and Pending Evaluation Attempt -> hide average time taken
Pending Evaluation Attempt -> hide average time taken
Published Evaluation Attempt -> display average time taken
In Progress Attempt -> hide average time taken

maddiegabriel:
> 
> The expected behavior cases are the following when I checked the legacy behavior : Published Evaluation Attempt and In Progress Attempt -> display average time taken Pending Evaluation Attempt and In Progress Attempt -> hide average time taken Pending Evaluation Attempt and Published Evaluation Attempt -> display average time taken Pending Evaluation Attempt and Pending Evaluation Attempt -> hide average time taken Pending Evaluation Attempt -> hide average time taken Published Evaluation Attempt -> display average time taken In Progress Attempt -> hide average time taken
>

ok cool thanks!! so looking at those scenarios - it seems like if there is a published attempt, we want to show the average time taken. otherwise, we want to hide it.

is there a reason we need to check for attempts in progress in the if block below and in `checkEvaluationStateExists`? or do you think we could simplify this by just iterating over the attempts (like in the else block) and returning true if there's a published attempt, false otherwise?

https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/56c43f0588a543810f04ba757a8b5a4ca761e580/components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js#L43

Girujan1998:
> > The expected behavior cases are the following when I checked the legacy behavior : Published Evaluation Attempt and In Progress Attempt -> display average time taken Pending Evaluation Attempt and In Progress Attempt -> hide average time taken Pending Evaluation Attempt and Published Evaluation Attempt -> display average time taken Pending Evaluation Attempt and Pending Evaluation Attempt -> hide average time taken Pending Evaluation Attempt -> hide average time taken Published Evaluation Attempt -> display average time taken In Progress Attempt -> hide average time taken
> 
> ok cool thanks!! so looking at those scenarios - it seems like if there is a published attempt, we want to show the average time taken. otherwise, we want to hide it.
> 
> is there a reason we need to check for attempts in progress in the if block below and in `checkEvaluationStateExists`? or do you think we could simplify this by just iterating over the attempts (like in the else block) and returning true if there's a published attempt, false otherwise?
> 
> https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/56c43f0588a543810f04ba757a8b5a4ca761e580/components/controllers/left-panel/ConsistentEvaluationLeftPanelController.js#L43

I think we can simplify it even more by doing what Scy mentioned of the `averageTimeTaken` property. It looks like if there is no published attempt, the properties do not exist and that would mean `averageTimeTaken` would be undefined.

Scy-D2L:
Generally looks good but with this adjustment it looks like we don't have a vdiff case where `averageTimeTaken` doesn't appear. You should probably add one.

Girujan1998:
> Generally looks good but with this adjustment it looks like we don't have a vdiff case where `averageTimeTaken` doesn't appear. You should probably add one.

I assumed that `quizzing-overall-summary-renders-quiz-overall-summary-with-no-Grade-Statistics.png` for `noGradeStatistics` would cover the case of no `averageTimeTaken`. Since no statistics would mean there is no published evaluation which would then also mean no `averageTimeTaken`.

maddiegabriel (approved):
Nice - looks good to me! 💯

PR #935, opened 2022-04-05
US137603 -- Pass the workingCopyId to the rubrics popout dialog

mpharoah-d2l:
Associated LMS Side PR: https://github.com/Brightspace/lms/pull/21510

Pass the rubrics workingCopyId to the popout dialog so it stays in sync with the inline view


components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -21,14 +21,26 @@ export class ConsistentEvaluationHrefController {
 	}
 
 	addWorkingCopyId(href, workingCopyId) {
-		if (href && workingCopyId) {
-			const parsedUrl = new URL(href);
-			if (!parsedUrl.searchParams.get('workingCopyId')) {
-				parsedUrl.searchParams.set('workingCopyId', workingCopyId);
-				return parsedUrl.href;
+		if (!href || !workingCopyId) return href;
```
mpharoah-d2l, line 24: Had to change this because the `URL` class doesn't work for relative URLs. Unit test still passes

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -21,14 +21,26 @@ export class ConsistentEvaluationHrefController {
 	}
 
 	addWorkingCopyId(href, workingCopyId) {
```
JamieTran, line 23: nit: `_addWorkingCopyId()` if private method

components/controllers/ConsistentEvaluationHrefController.js
```diff
@@ -302,7 +314,10 @@ export class ConsistentEvaluationHrefController {
 			}
 
 			if (root.hasSubEntityByRel(Rels.Assessments.assessRubricApplication)) {
```
JamieTran, line 316: nit: This _could_ be extracted into its own method like `getRubricPopoutLocation()` so that `getHrefs()` doesn't have to know about `workingCopyId` but I am indifferent.

JamieTran (approved):
a couple `nit` but lgtm

components/controllers/ConsistentEvaluationHrefController.js
mpharoah-d2l: It is "private", but called directly by unit tests. Should I add an `_`?

JamieTran: I would think so, to indicate that we don't want this method being used outside of this class (excluding tests)

devpow112 (approved):

PR #936, opened 2022-04-07
Hide warning if workingCopyId is present

MitchelPaulin:
We don't want the warning to show if we are using a working copy. Hid the warning in the inline component here https://github.com/Brightspace/d2l-rubric/pull/1831. This will also hide the warning in the rubric popout if we get a `workingCopyId`


Scy-D2L (approved):

maddiegabriel (approved):
looks good to me!

PR #938, opened 2022-04-07
Add manually override grade

JamieTran:
US130345

Grade object will contain `calculatedScore` and `isManuallyOverridden` and will be used when determining when to display the override button.

Reliant on this [PR](https://github.com/BrightspaceUILabs/grade-result/pull/85)

quiz-override


components/right-panel/consistent-evaluation-right-panel-grade-result.js
```diff
@@ -192,6 +191,19 @@ export class ConsistentEvaluationRightPanelGradeResult extends LocalizeConsisten
 		}
 	}
 
+	_handleManualOverrideClearClick() {
```
jeroach, line 194: Sorry, where is this new method called from?

JamieTran: Oh lol that is confusing, the event-listener was already implemented so git didn't pick up a change. https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/right-panel/consistent-evaluation-right-panel-grade-result.js#L132

Girujan1998 (approved):

components/right-panel/consistent-evaluation-right-panel-grade-result.js
jeroach: Thanks!

jeroach (approved):

PR #939, opened 2022-04-08
US137592: Unsaved Changes Dialog for Quiz Attempts

Scy-D2L:
Shows unsaved changes dialog when there are unsaved changes for individual quiz attempts.
If the user selects `save` or `discard` the changes are handled appropriately. If the user `closes` the dialog, the `attempt dropdown` reverts to the current attempt rather than whatever was selected.

[US137592](https://rally1.rallydev.com/#/?detail=/userstory/630730923585&fdp=true): [Consistent Evaluation][Quizzing] Show unsaved changes dialog when switching between attempts


components/consistent-evaluation-dialogs.js
```diff
@@ -122,7 +122,7 @@ export class ConsistentEvaluationDialogs extends LocalizeConsistentEvaluation(Li
 	}
 
 	async _fireDialogClosedEvent() {
-		this.dispatchEvent(new CustomEvent('d2l-dialog-closed', {
```
Scy-D2L, line 125: This event is fired directly by the `d2l-dialog` component whenever *any* of our dialogs are closed, so this was redundant. Replaced with a unique event to identify selecting the `close` button.

components/consistent-evaluation-page.js
```diff
@@ -373,7 +375,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 				@d2l-consistent-evaluation-evidence-back-to-user-submissions=${this._setSubmissionsView}
 				@d2l-consistent-evaluation-file-selected=${this._selectFile}
 				@d2l-consistent-evaluation-evidence-back-to-completion-summary=${this._setCompletionSummaryView}
-				@d2l-consistent-evaluation-quizzing-attempt-selected=${this._selectQuizAttempt}
+				@d2l-consistent-evaluation-quizzing-attempt-selected=${this._handleSwitchAttempt}
```
Scy-D2L, line 378: Our mix of `handle-` `on-` and unqualified event-handler method names is confusing and something I want to fix in a refactor sweep when there's time.

components/consistent-evaluation-page.js
```diff
@@ -897,7 +946,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 					.anonymousInfo=${this.anonymousInfo}
 					.token=${this.token}
 					.currentFileId=${this.currentFileId}
-					.currentQuizAttemptHref=${this.currentQuizAttemptHref}
```
Scy-D2L, line 900: This is a string so I'm not sure why it was property-bound

amauryperez (approved):
LGTM

components/consistent-evaluation-page.js
```diff
@@ -549,14 +552,18 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
...
 	}
 
-	_changeFile(newFileId) {
-		this.currentFileId = newFileId;
+	_cancelDialog() {
+		this._navigationTarget = null;
+		if (this._targetAttemptHref) {
+			//Reset LCB if we had selected a different attempt
+			this._learnerContextBar = this.shadowRoot && this.shadowRoot.querySelector('d2l-consistent-evaluation-learner-context-bar');
+			this._learnerContextBar.currentQuizAttemptHref = this.currentQuizAttemptHref;
```
jeroach, line 560: I think this is fine, but just double checking to be safe :)

Can `this._learnerContextBar` ever be null here?

Scy-D2L: I don't thiiiiink so, but I'll put a check in anyway. It also occurs to me that we should be grabbing that at the same place where we get the other subcomponents so we aren't grabbing it every time.

components/consistent-evaluation-page.js
```diff
@@ -830,6 +872,13 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		this._isRubricAssessmentModified = false;
 	}
 
+	_onSwitchAttempt() {
+		this.currentQuizAttemptHref = this._targetAttemptHref;
+		const foundAttempt = Object.values(this.quizAttemptInfoList).find(attempt => attempt.attemptHref === this._targetAttemptHref);
+		this.evaluationHref = foundAttempt.attemptEvaluationHref;
```
jeroach, line 878: Similar question to above - I assume we always will find an attempt and that the error state can't happen,

but just want to check that assumption that `foundAttempt` can never be null

Scy-D2L: `foundAttempt` can only be null if the `quizAttemptInfoList` has changed since the attempt selection event was triggered, because that's the source of the `_targetAttemptHref`. That would be a complete failure case we should never be able to hit without code-changes.

jeroach (approved):
lgtm!

PR #942, opened 2022-04-11
US137885 -- Commit/Discard working copy

mpharoah-d2l:
When using rubrics working copy, commit the working copy when publishing, saving, or updating, and discard the working copy when retracting.

This PR does *NOT* update any dialogs or text to reflect the working copy experience. This work will be done in future pull requests / user stories. This PR is simply causing the working copy to be committed/discarded when the appropriate button is clicked.

The rubrics working copy flag (`rubrics-working-copy`) is still disabled by default in dev.


mpharoah-d2l:
Hmmm, the tests don't fail on my local checkout

components/consistent-evaluation-page.js
```diff
@@ -868,7 +868,7 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 			async() => {
 				if (this._isValidEvaluation()) {
 					const entity = await this._controller.fetchEvaluationEntity(false);
-					const newEvaluationEntity = await this._controller.publish(entity);
+					const newEvaluationEntity = await this._controller.publish(entity, this.href, this.rubricWorkingCopyId);
```
JamieTran, line 871: What are your thoughts on passing in the `rubricInfos` on publish, update, retract, and save instead of instantiating the `hrefController` in the `CEController`? In terms of our transient-save we try to decouple the two controllers but working copy might be different.

mpharoah-d2l: I can do that. I did it this way so that it doesn't refetch the rubric infos if it doesn't need to (there's no working copy)

mpharoah-d2l: Oh wait, I didn't realize that `rubricInfos` is already passed into the controller. That works better then

JamieTran: yeah `rubricInfos` already exists on the [CE-Page](https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/blob/master/components/consistent-evaluation-page.js#L73) level where we perform all our actions on the evaluation

JamieTran (approved):
lgtm

devpow112 (approved):
Looks good to me :tada:

PR #943, opened 2022-04-11
🐛 Possible fix for spaces not rendering in dialog

Scy-D2L:
The story issue doesn't seem to happen on local, so I'm not 100% sure this will fix it - but it's trivial and can't break anything so giving it a shot.

[US137944](https://rally1.rallydev.com/#/?detail=/userstory/631341156903&fdp=true): CE Quizzing - Reset Evaluation dialog is displaying tags in the text


maddiegabriel (approved):
worth a shot 👌

amauryperez (approved):

PR #946, opened 2022-04-12
fix: US137580 - Save Score/Feedback Without Having To Always Click Outside Grade Box

Girujan1998:
Rally: https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F630713522257

Fixing saving question grade/feedback issue where user would always be required to click outside the grade/feedback box before clicking Save Draft / Publish / Update to persist the score when performing one of the following:
- Switching attempts
- Switching to Completion Summary and back
- Refreshing the page


components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
```diff
@@ -232,11 +222,8 @@ export class ConsistentEvaluationQuizzingAttemptResult extends SkeletonMixin(Loc
 		if (this.score === newScore) {
 			return;
 		}
-		this._debounceJobs.grade = Debouncer.debounce(
```
Girujan1998, line 235: Removing the debounce, since the function is called once when we click outside the gradebox.

amauryperez (approved):

Scy-D2L (approved):

PR #954, opened 2022-04-13
jtran hook up attempt entry points

JamieTran:
Opens consistent-evaluation with correct `attempt` when selected, this information is provided in the URL as `selectedItem` and is encoded. 

We decode the selectedItem and compare that with the list of `attemptHrefs` that we have. There is some extra logic for navigation to determine whether or not to display `1st attempt` or `completion summary`.


Scy-D2L:
Should we be reflecting this to the browser when we change attempts? (ie: `_updateCurrentUrl()` when `_currentQuizAttempt` changes?)

components/consistent-evaluation.js
```diff
@@ -427,6 +419,30 @@ export class ConsistentEvaluation extends LocalizeConsistentEvaluation(LitElemen
 		this._loading = true;
 	}
 
+	_setQuizzingHrefs() {
+		if (this._stripSelectedItemFromUrl()) {
+			let attemptFound = false;
+			const numAttempts = Object.keys(this._quizAttemptInfoList).length;
+			for (const attemptNumber in this._quizAttemptInfoList) {
```
Scy-D2L, line 426: Because we're only expecting one match, this could be streamlined with [array.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

JamieTran:
> Should we be reflecting this to the browser when we change attempts? (ie: `_updateCurrentUrl()` when `_currentQuizAttempt` changes?)

We could, but I am wondering what we would use it for other than a state of truth. We only use the URL once when we enter into CE-quizzing and don't use it after that.

Scy-D2L:
> > Should we be reflecting this to the browser when we change attempts? (ie: `_updateCurrentUrl()` when `_currentQuizAttempt` changes?)
> 
> We could, but I am wondering what we would use it for other than a state of truth. We only use the URL once when we enter into CE-quizzing and don't use it after that.

If you switch attempt/student after entering CE and refresh the page at the moment on your branch, will it remain on the same student/attempt?

JamieTran:
> > > Should we be reflecting this to the browser when we change attempts? (ie: `_updateCurrentUrl()` when `_currentQuizAttempt` changes?)
> > 
> > 
> > We could, but I am wondering what we would use it for other than a state of truth. We only use the URL once when we enter into CE-quizzing and don't use it after that.
> 
> If you switch attempt/student after entering CE and refresh the page at the moment on your branch, will it remain on the same student/attempt?

ahh ok, I see where you're coming from. On student switch, yes it will stay on the correct student. But on refreshing on attempt switch will not save the current attempt you are on and will go through the `if attempts === 1 display first attempt else display completion summary`

Scy-D2L:
> > > > Should we be reflecting this to the browser when we change attempts? (ie: `_updateCurrentUrl()` when `_currentQuizAttempt` changes?)
> > > 
> > > 
> > > We could, but I am wondering what we would use it for other than a state of truth. We only use the URL once when we enter into CE-quizzing and don't use it after that.
> > 
> > 
> > If you switch attempt/student after entering CE and refresh the page at the moment on your branch, will it remain on the same student/attempt?
> 
> ahh ok, I see where you're coming from. On student switch, yes it will stay on the correct student. But on refreshing on attempt switch will not save the current attempt you are on and will go through the `if attempts === 1 display first attempt else display completion summary`

I suspect we'd prefer to keep the same attempt loaded in this case.😃

components/consistent-evaluation.js
JamieTran: `array.find()` doesn't work on `this._quizAttemptInfoList` as its an object and not an array. I think changing to an array would be beneficial. I don't think we should make this change now being so close to release and not knowing how that would impact the rest of our code.

JamieTran: jk I think i figured it out

Scy-D2L: Yep, it's possible to treat object properties as array entries. :) And I agree about changing it to directly operate as an array in the longer term, it's kind of weird to have an object with numbered entries as properties 🤔

JamieTran:
> > > > > Should we be reflecting this to the browser when we change attempts? (ie: `_updateCurrentUrl()` when `_currentQuizAttempt` changes?)
> > > > 
> > > > 
> > > > We could, but I am wondering what we would use it for other than a state of truth. We only use the URL once when we enter into CE-quizzing and don't use it after that.
> > > 
> > > 
> > > If you switch attempt/student after entering CE and refresh the page at the moment on your branch, will it remain on the same student/attempt?
> > 
> > 
> > ahh ok, I see where you're coming from. On student switch, yes it will stay on the correct student. But on refreshing on attempt switch will not save the current attempt you are on and will go through the `if attempts === 1 display first attempt else display completion summary`
> 
> I suspect we'd prefer to keep the same attempt loaded in this case.😃

This is not required for [M1](https://d2l.slack.com/archives/CSL64JP9U/p1649892300309049?thread_ts=1649875141.812259&cid=CSL64JP9U). I'll create a new story for this

amauryperez (approved):

maddiegabriel (approved):
LGTM

Scy-D2L (approved):

PR #956, opened 2022-04-13
US137887 - Quizzing - Question level score validation

maddiegabriel:
[US137887  - [Consistent Evaluation][Quizzing][UI] Update question and attempt score to not support null](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F631329991735)

This PR adds validation to the question-level score box. Previously we weren't getting a score box error for null/empty scores, and also weren't seeing the error toast for any invalid scores (i.e. negative).

The ticket asks the functionality to match legacy, where if the user clears the scorebox then tries to save the evaluation they will get an error and the scorebox will remain blank. Like this:
image

I don't think this is possible with the `d2l-labs-d2l-grade-result-presentational` component we use in CE, since [the `scoreNumerator` property defaults to 0](https://github.com/BrightspaceUILabs/grade-result#d2l-labs-d2l-grade-result-presentational) when you pass in an undefined value (i.e. we can't render with a blank box). You can see in the second gif below that is automatically populates the score box to 0 once it's cleared. This isn't perfect because it makes it look like the score is zero and invalid. Open to suggestions!

New functionality:

QUESTION_SCORE_INVALID_TOAST_WORKING

QUESTION_SCORE_NULL_TOAST_WORKING


components/consistent-evaluation-page.js
```diff
@@ -1288,6 +1288,11 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		await this._mutex.dispatch(
 			async() => {
 				const newScore = e.detail.score;
+				if (typeof e.detail.isValidScore !== 'undefined') {
+					this._isValidGrade = e.detail.isValidScore;
```
maddiegabriel, line 1292: this is to make the error toast appear; same functionality we use in `_transientSaveGrade()`

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
```diff
@@ -169,6 +180,31 @@ export class ConsistentEvaluationQuizzingAttemptResult extends SkeletonMixin(Loc
...
 
+	_renderQuestionGradeResult() {
+		if (typeof this._gradeValidationError !== 'undefined') {
+			return html`
+				<d2l-labs-d2l-grade-result-presentational
+					.gradeType=${GradeType.Number}
+					scoreNumerator=${this.score}
+					scoreDenominator=${this.outOf}
+					subtitleText=${this.localize('score')}
+					validationError=${this._gradeValidationError}
```
maddiegabriel, line 191: I tried just passing the `validationError` into `d2l-labs-d2l-grade-result-presentational` as `undefined` when the score is valid, but I was having issues with the error message persisting if the user entered an invalid score followed by a valid one. This is why I moved this into a function and am rendering with or without `validationError` based on the definition of `this._gradeValidationError`

JamieTran: I think the best way to solve this is by fixing this in `d2l-labs-d2l-grade-result-presentational` and finding out why it is persisting with a valid score (probably not receiving an update when undefined is passed in). This is a good quick fix if we need this in for M1.

maddiegabriel: cool i'll take a look!

JamieTran:
> I don't think this is possible with the d2l-labs-d2l-grade-result-presentational component we use in CE, since [the scoreNumerator property defaults to 0](https://github.com/BrightspaceUILabs/grade-result#d2l-labs-d2l-grade-result-presentational) when you pass in an undefined value (i.e. we can't render with a blank box). You can see in the second gif below that is automatically populates the score box to 0 once it's cleared. This isn't perfect because it makes it look like the score is zero and invalid. Open to suggestions!

`not auto-graded: a default grade of 0 will be assigned` IMO I think that if the user purposely puts nothing in the field and it defaults the score of 0 is fine, as long as the validationError doesn't pop up. This would be more of a design question though.

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
maddiegabriel: fix for this is in https://github.com/BrightspaceUILabs/grade-result/pull/88

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
```diff
@@ -128,20 +137,22 @@ export class ConsistentEvaluationQuizzingAttemptResult extends SkeletonMixin(Loc
...
 			detail: {
 				score: newScore,
+				isValidScore: isValidScore,
 				questionResponseHref: this.questionResponseHref
 			}
 		}));
 	}
 
+	_isValidScore(score) {
+		return typeof score === 'number' && score >= 0 && score <= 9999999999;
```
Scy-D2L, line 153: Adding an entire function for this may be overdoing it when when the logic is only called one time.

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-result.js
```diff
@@ -90,6 +94,11 @@ export class ConsistentEvaluationQuizzingAttemptResult extends SkeletonMixin(Loc
 		`];
 	}
 
+	constructor() {
+		super();
+		this._scoreValidationError = undefined;
```
Scy-D2L, line 99: Properties default to undefined, so we shouldn't need a constructor for this.

JamieTran (approved):
lgtm

Scy-D2L (approved):
LGTM

PR #957, opened 2022-04-14
US137944: 🐛 Fix reset dialog text displaying tags

Scy-D2L:
Fix for [US137944](https://rally1.rallydev.com/#/?detail=/userstory/631341156903&fdp=true): CE Quizzing - Reset Evaluation dialog is displaying tags in the text


components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -588,10 +588,12 @@ export class ConsistentEvaluationRightPanel extends SkeletonMixin(LocalizeConsis
 	}
 
 	_renderResetEvaluationConfirmDialog() {
+		//Text is split into 3 pieces because `localize` doesn't like newlines.
```
JamieTran, line 591: damn, that's annoying. Nice fix

amauryperez (approved):

PR #963, opened 2022-04-20
Switched to rubric built in title and graded by

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fuserstory%2F630669079213

Were making some changes in rubrics and switching to handling the graded by info on our end, this change is being controlled by the `rubrics-archive` feature flag. Since we are now handling the graded by info we can also handle displaying the title ourselves using the `show-title` tag.


imcdonald-d2l (approved):

devpow112 (approved):

components/right-panel/consistent-evaluation-rubric.js
```diff
@@ -285,15 +285,18 @@ class ConsistentEvaluationRubric extends LocalizeConsistentEvaluation(RtlMixin(L
 		return `(${this.localize('notGradedSummary')})`;
 	}
 	_getRubrics() {
+
+		const rubricsArchivedFeature = window.D2L?.LP?.Web?.UI?.Flags?.Flag('rubrics-archive', false) ?? false;
```
amauryperez, line 289: Nice, I didn't know there was a way to use feature flags within web components. I think we used send FF values as attribute of the component.

amauryperez (approved):

components/right-panel/consistent-evaluation-rubric.js
devpow112: It's a bit clunky but it works. This also works within FRA's (I added some sync functionality a while back)

amauryperez: Does it work in all supported browsers?

devpow112: Should be, it's used in a few spots already and hasn't thrown up any red flags.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility

PR #964, opened 2022-04-20
fix: DE48211 - Dropdown displays Attempt 1 while page displays Completion Summary

Girujan1998:
Rally: https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F632015377193

When iterating between users, we displaying Completion Summary page but the dropdown displays Attempt 1.


components/header/consistent-evaluation-attempt-context.js
```diff
@@ -63,8 +63,15 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
...
+	async updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('currentQuizAttemptHref')) {
+			this.refreshQuizSelect();
+		}
+	}
+
 	refreshQuizSelect() {
-		if (this.shadowRoot) {
+		if (this.shadowRoot && this.currentQuizAttemptHref) {
```
Girujan1998, line 74: Adding this as we can get errors trying to set the value to null. We only want to update the select with a defined value.

amauryperez (approved):
LGTM

JamieTran (approved):
lgtm!

PR #966, opened 2022-04-21
US137548 - Show standard unsaved changes dialog when rubrics working copy ID present

devpow112:
Made sure this is backwards compatible since we still want to show the existing rubric specific dialog if the working copy ID is not present. Added new vdiff for making sure when working copy is present we use the "standard" dialog.

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F630652950611


jeroach (approved):
I'm not very familiar with this code, but what's here looks good to me :)

PR #972, opened 2022-04-22
Update rubrics

devpow112:
Seeing if recent changes to rubrics broken vdiff.


devpow112:
Nope no changes so abandoning.

PR #973, opened 2022-04-25
Us138055 - add overridden confirm dialog

JamieTran:
[Us138055 ](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F631594745891)

Add warning dialog for an overridden grade when saving, updating, and publishing an evaluation.


components/consistent-evaluation-page.js
```diff
@@ -637,6 +644,21 @@ export default class ConsistentEvaluationPage extends SkeletonMixin(LocalizeCons
 		);
 	}
 
+	async _displayOverridenGradeWarning() {
```
JamieTran, line 647: This function displays the dialog and waits for the dialog confirmation response.

amauryperez:
Looks good. I was also thinking that since this is only for Quizzing if we can put into a separate file or something, it will help limit the size of `consistent-evaluation-page.js` and keep things that don't apply to all activities separated?

maddiegabriel (approved):
looks good to me!

PR #976, opened 2022-04-27
Scy/ us138136 / Display Attempt Lateness

Scy-D2L:
Display the lateness for quiz attempts

image

Lateness is displayed using langterms/logic based on [what's used by legacy quiz evaluation](https://github.com/Brightspace/lms/blob/master/le/D2L.LE/D2L.LE.Core/Domain/QuizLateText/FuzzyLateTimeProvider.cs).

Related to [US138136](https://rally1.rallydev.com/#/?detail=/userstory/631718704117&fdp=true): [Consistent Eval][Quizzes][UI] Attempt summary needs to display if late when due date set


JamieTran (approved):
lgtm

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-summary.js
```diff
@@ -44,6 +44,10 @@ export class ConsistentEvaluationQuizzingAttemptSummary extends SkeletonMixin(Rt
 			isTimeEnforced: {
 				attribute: 'time-limit-enforced',
 				type: Boolean
+			},
+			lateTime: {
+				attribute: 'time-late',
```
jeroach, line 49: nit: this naming is a touch confusing because of the inconsistency - "latetime" vs "timelate"

Is there a reason it can't always be one or the other?

components/left-panel/quizzing/consistent-evaluation-quizzing-attempt-summary.js
```diff
@@ -130,6 +137,25 @@ export class ConsistentEvaluationQuizzingAttemptSummary extends SkeletonMixin(Rt
 		return Math.floor(this.graceLimit / 60);
 	}
 
+	_renderLate() {
+		if (this.lateTime) {
```
jeroach, line 141: Nit: with these conversions, could we explicitly name the `lateTime` in the units - e.g. `timeLateInSeconds` or `secondsLate` ?

Scy-D2L: Going to update with `timeLate` for consistency, but not adding units to the name at the moment because we aren't doing that anywhere else. It would be a good thing to do, but possibly confusing to only have for this one timespan and not the others.

PR #978, opened 2022-04-27
DE48178 - Discarded changes to attempt evaluations persist when switching back and forth between students

Girujan1998:
Rally: https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F631783812905

Fixing issues where if we try to iterate between students with unsaved changes, when we return to the student attempt with the unsaved changes the unsaved changes still persist on screen.

This fix currently fixes the issue where unsaved changes score persists when coming back to the attempt and displays the correct score. However, there exists a small flicker where we display the unsaved change and then immediately change the score to the correct one.


JamieTran (approved):
I think this is a good fix, do we have a secondary story created encompassing the flickering work?

Scy-D2L:
> I think this is a good fix, do we have a secondary story created encompassing the flickering work?

[US138683](https://rally1.rallydev.com/#/?detail=/userstory/633593219505&fdp=true): [Consistent Evaluation][Quizzing][UI] Discarded score flickers when returning back to student

PR #982, opened 2022-04-28
Backport to 2022.5.x of "US137887 - Quizzing - Question level score validation #956" 

maddiegabriel:
Rally: https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fuserstory%2F632101169639
PR: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/956


jeroach (approved):
lgtm - looks exactly like the original PR :)

JamieTran (commented):
Sorry if I missed it but have we had a discussion of whether the bug (changing score from 0 -> null -> not showing tooltip) outweighs the pros of adding the feature? I assume it does but I ask because the main [story](https://rally1.rallydev.com/#/42960320374d/detail/userstory/631329991735) was sent back to dev.

maddiegabriel:
good question! talked about it with Dale and Peggy at the end of grooming - we decided that this is better than what was there before, especially since the toast wasn't appearing at all. We're going to move the original story to 'accepted' (I just did that) and backport these changes. Then Peggy created a new story for the remaining validation issues (the zero thing we're trying to fix) - here it is https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fuserstory%2F633568672801

JamieTran (approved):
thanks for the info :)

PR #987, opened 2022-05-02
Added "Evaluator" section

alexvlsem:
[US134174](https://rally1.rallydev.com/#/?detail=/userstory/618510463515&fdp=true): Consistent Eval >  Add "Evaluator" section


components/right-panel/consistent-evaluation-right-panel.js
```diff
@@ -492,6 +498,16 @@ export class ConsistentEvaluationRightPanel extends SkeletonMixin(LocalizeConsis
 			`;
 		}
 	}
+
+	_renderEvaluators() {
+		return html`
+		<d2l-consistent-evaluation-right-panel-evaluators
+			aria-hidden="true"
```
Scy-D2L, line 505: Should this be `aria-hidden="${this.skeleton}"` or is there a reason we don't want this to be accessible?

components/right-panel/consistent-evaluation-right-panel-evaluators.js
```diff
@@ -0,0 +1,69 @@
...
+			},
+			token: {
+				type: String
+			}
+		};
+	}
+
+	constructor() {
+		super();
+		this.store = store;
```
Scy-D2L, line 24: Somewhat confused about having this here and at `60`, but using the global imported `store` at `65`. Also should it be `_store` given that it's only used internally?

Scy-D2L (commented):
I have a mild concern that there are no tests included for the new component - we generally prefer to have at least a vdiff test/demo to ensure nothing changes within the app (and help identify if an API change differs from our test/demo data and might be responsible.) The block itself isn't heavily reliant on other CE components so if the `store` used makes local testing too complicated it's probably ok. 🤔

components/right-panel/consistent-evaluation-right-panel.js
alexvlsem: Thank you, forgot to switch back.  Fixed.

components/right-panel/consistent-evaluation-right-panel-evaluators.js
alexvlsem: Definitely agree with you.

Scy-D2L (approved):
LGTM

PR #998, opened 2022-05-09
DE47028 - Make archived-rubrics read-only 

JamieTran:
[DE47028](https://rally1.rallydev.com/#/42960320374d/custom/236803223728?detail=%2Fdefect%2F623703963645)

- Makes `archived-rubrics` uneditable in CE

Notes:
In both activity creation and the activity evaluation, if there is a rubric aligned that has an archived status, then we should display the archived tag with that rubric. (I.E. always display the rubric)


amauryperez (approved):
nice

maddiegabriel (approved):
lgtm!

PR #1004, opened 2022-05-11
US138865: html block margins

Scy-D2L:
Added margins/padding to `d2l-html-block`s. Previously we allowed the contents of these blocks to define the spacing, but recent updates removed the top/bottom content spacing within the blocks - so now we need to define out own.

Units were chosen to most closely resemble the padding inherited from paragraph styles. Vdiffs are expected to be slightly different because of the mix of paragraphs and headers in test data (headers having different spacing that would have been used before) and possibly off by a few pixels in image size/aliasing because of recent vdiff updates.

Expected to resolve issues noted in https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/990 and https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/970

Related story: [US138865](https://rally1.rallydev.com/#/?detail=/userstory/634970449605&fdp=true): [Consistent Evaluation] Update html-block margins to fix vdiff errors


amauryperez (approved):

maddiegabriel (approved):
L G T M! 🥇

PR #1007, opened 2022-05-12
DE48407 - Check for attempts when setting quizHref

JamieTran:
When iterating to a student with no attempts would throw an error. This is mainly because we do not yet support changing the URL when iterating/attempt change. This work is a part of [US138182](https://rally1.rallydev.com/#/42960320374d/detail/userstory/631777070717) but this is a a safety to not run into the error


components/header/consistent-evaluation-attempt-context.js
```diff
@@ -71,7 +71,7 @@ export class ConsistentEvaluationAttemptContext extends RtlMixin(LocalizeConsist
 	}
 
 	refreshQuizSelect() {
-		if (this.shadowRoot && this.currentQuizAttemptHref) {
+		if (this.shadowRoot && this.shadowRoot.querySelector('select') && this.currentQuizAttemptHref) {
```
JamieTran, line 74: this is a fix to a different error where we were trying to set the attempt drop-down when the dropdown didn't exist

Scy-D2L: Might be worth pulling that query into a variable and reusing it so we don't query our shadowdom twice.

jeroach (commented):
lgtm once Scy's comment is addressed :)

Scy-D2L (approved):

jeroach (approved):

